PR #8, opened 2017-11-15
Add new Config Switch dvi_mode

umatthe:
This adds a new config line to MiSTer.ini:

dvi_mode=1 ; set to 1 for DVI-Monitor / 0 (default) for HDMI-Monitor

The new config needs to be added the fpga core hdmi_config (see example in C64 core).
Without changes in the fpga core this configuration does nothing.


sorgelig:
will be implemented along with other features.

PR #13, opened 2018-04-30
replace hardcoded IP address with mDNS hostname

sonic74:


sonic74:
For e.g. https://www.waveshare.com/product/modules/oleds-lcds/raspberry-pi-lcd/7inch-hdmi-lcd-h-with-case.htm
I hoped _Desktop Linux_ would pick it up, but it seems to need some closed source change.

sonic74:
3111fcb was caused by both Sublime and gedit automatically converting these two lines with Linux line endings in an otherwise Windows-line-endings-file. Quite a pain to edit without causing these spurious changes .-)

PR #19, opened 2018-10-01
Additions for Sharp MZ

pdsmart:
The Sharp MZ emulator has made extensive use of the MiSTer menus for configuration options. Due to the number of getter/setter methods and the number of menu options I didnt want to overload the menu.cpp file with this clutter, hence following along the lines of the Archimedes and creating a seperate module and making a call to it should the emulation be running a Sharp MZ core.


sorgelig:
Please re-make the changes to avoid such large changes where only CR/LF are changed.
i remember Git has option to ignore line endings differences.

pdsmart:
Not sure what happened, on first push I looked at the diffs and saw the issue, so I manually editted the files and re-pushed. Diff before Pull Request showed only my changes so I made the pull request but looking at the diffs now, it is back to the original state!

Will either set auto crlf and see what happens.

pdsmart:
Auto cr-lf makes it worse, deleting pull request and will try to re-do before making a new pull

sorgelig:
probably you need to re-fork again, turn the option on and then apply your patch.

PR #20, opened 2018-10-06
SharpMZ Additions

pdsmart:
I finally found the issue with the large diff between the master and the fork after editing. It wasn't the git autocrlf, although this is good practice to set, it was the editor I was using, gvim and gvimdiff. If gvim option fileformat=dos then it keeps the dos format and converts any lf to crlf upon write, it gvim option fileformat=unix then it will convert all crlf to lf upon write. Trouble was, the files user_io.h, user_io.cpp and menu.cpp have a mix of formats, some sections have crlf, some don't so using gvim normalised to one format or the other, thus creating large differences with the main repository.

Solution was to us vi as it doesn't make any form of conversion upon write.


sorgelig:
I've merged the code myself as your PR still shows huge diff. It seems Linux tools cannot support different formats transparently.

Actually special support for computer which is even lower range than ZX Spectum isn't rational. If every core would require such large code from ARM side, then MiSTer binary would be several megabytes. And i don't even speak about maintaining such code. Probably it would be more rational to use soft-cpu like ZPU or NIOSII for custom support if it's hard to implement in FPGA.
You can check ZX Spectum has disk support and tape support. And most code is written on FPGA side rather than moving it to ARM. So this is preferred way to make the cores.
Later MiSTer may get some re-design and it will be a headache to deal with custom code.

PR #25, opened 2018-11-01
Organizing directory structure

dentnz:
Hi,

This pull request is about creating a directory structure for Main_MiSTer. I began to look at the codebase and saw I could categorize some of the core-specific files into a 'support/<core>' structure. The hope being, that this will make the code easier to manage and maintain as new cores are added.

Here's a summary of the changes made:

- Created a directory structure for core support code and moved core-specific files into said structure.
- Updated the Makefile to include new core directories in the build.
- Updated all core support files includes to use a relative path for the main Main_MiSTer header files.
- Created a support.h file that can be used by the main Main_MiSTer files to include core support files.
This allows you to maintain the set of core includes externally to the files that use them.

Looking for your feedback and consideration on this *only* at this stage. If this is a path you want me to continue down, I'd like to see what else can be moved into the core support files.

Cheers,
Karl


sorgelig:
Moving core specific files to separate folders - ok. But i don't want further tossing the files.
I will check your changes and some of them probably will be reverted as i see you changed the path of toolchain i don't agree.

sorgelig:
Ok. it seems current version compiled without problems.

dentnz:
Thanks Sorgelig! Yes, I removed the change to the makefile for the docker container. Best to keep that out of this repository. Instead, I created another repo for this and other development environment setup.

Cheers.

PR #28, opened 2018-11-14
Automatically create blank .sav files for cores that save.

Kitrinx:
If the core indicates FS in the menu, and a .sav file does not exist for the rom, it will be created. The size is 128kb by default, or the core can specify the desired .sav size with a power-of-two decimal number following FS in the menu option, eg FS13. If the .sav file was not used by a game, it is automatically removed from the disk the next time a rom is loaded for any core, to avoid clutter.

### DIFF
--- a/file_io.cpp
+++ b/file_io.cpp
@@ -330,6 +330,27 @@ int FileCanWrite(const char *name)
 	return ((st.st_mode & S_IWUSR) != 0);
 }
 
+int FileRemove(const char *name, int mode, char mute)
+{
+	const char *root = getRootDir();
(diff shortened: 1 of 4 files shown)


sorgelig:
You need to rework this feature cause i don't like constant save/delete of file on every rom loading. Saves to SD card should be minimized to prevent corruptions. User may switch off or press hard reset right after loading the rom which may leave file system in intermediate state.
This situation is worse because it involves not only write (write to the same file more or less secure as it only mess inside the file, not in FAT/catalog) it involves file creation and deletion - so it's most dangerous operation!
I suggest to create the file only when actual save of backup happens. So it will eliminate creation/deletion for every rom loading.

PR #30, opened 2018-12-27
Clang build support

dholm:


sorgelig:
I don't understand purpose of these strange pull requests (including that one i've just rejected).
It doesn't add any feature except the tossing and obfuscating the code and scripts.

sorgelig:
Btw, GCC is also available almost for every platform.
I compile MiSTer under windows.
I want to tell even more: I use Microsoft Visual Studio 2015 to develop and compile the MiSTer (that's why you see MSVS project files). GCC is more wide spread and supported than newcomer clang. 
So, no need to make simple things more complicated.

dholm:
@sorgelig Clang is very convenient because you don't need to build cross-compilers for every target as you can compile for any supported target with a single installation. This change simply made it possible to manually switch to using Clang, the default was still to use GCC.

sorgelig:
I would like to keep a single unified compiler GCC. 
Probably clang can be convenient for someone but as a whole different compiler with its own requirement it may produce binary with other behavior. So it's not good idea to do this.

dholm:
@sorgelig Understood.

PR #31, opened 2018-12-27
makefile: Enable additional warnings

dholm:
Enables additional compiler warnings and treats warnings as errors.
Also fixes all warnings detected by Clang 7.0.0.


sorgelig:
i hate std::*

sorgelig:
If you've found some error in assignments of handling some variables, then you are welcome to provide the patch.
But i don't accept PR in the way "I like this style of programming lets convert the whole code to my style."

dholm:
@sorgelig Do you hate all things from std::*? I'm working on a feature and if you dislike all of std::* I will avoid using it.

My intention was not to change the style of your code with this change, in fact I tried to stay as close to the current style as I could.  I simply tried to fix issues detected by Clang, many of them are probably minor (different signedness in comparisons) but there are also a number of important fixes.  If you want I can try to extract the important fixes and submit another pull request?

DiskImage.cpp
```diff
@@ -1394,7 +1394,7 @@ void TDiskImage::readFDI(int hfile, bool readonly)
 		{
 			if (rsize < fdiOFF)
 			{
-				delete tracksinfo;
```
dholm: This is a memory leak since the pointer was allocated as an array but deleted as a pointer (only the first element will be deleted, the rest will be leaked).

Makefile
```diff
@@ -27,7 +27,7 @@ VPATH	= ./:./support/minimig:./support/sharpmz:./support/archie:./support/st:./s
 OBJ	= $(SRC:.c=.o) $(SRC2:.cpp=.o) $(MINIMIG_SRC:.cpp=.o) $(SHARPMZ_SRC:.cpp=.o) $(ARCHIE_SRC:.cpp=.o) $(ST_SRC:.cpp=.o) $(X86_SRC:.cpp=.o) $(SNES_SRC:.cpp=.o)
 DEP	= $(SRC:.c=.d) $(SRC2:.cpp=.d) $(MINIMIG_SRC:.cpp=.d) $(SHARPMZ_SRC:.cpp=.d) $(ARCHIE_SRC:.cpp=.d) $(ST_SRC:.cpp=.d) $(X86_SRC:.cpp=.d)	$(SNES_SRC:.cpp=.d)
 
-CFLAGS	= $(DFLAGS) -c -O3 -D_FILE_OFFSET_BITS=64 -D_LARGEFILE64_SOURCE -DVDATE=\"`date +"%y%m%d"`\"
+CFLAGS	= $(DFLAGS) -Wall -Wextra -Werror -c -O3 -D_FILE_OFFSET_BITS=64 -D_LARGEFILE64_SOURCE -DVDATE=\"`date +"%y%m%d"`\"
```
dholm: Are you opposed to enabling the additional diagnostics here or can I resubmit with the enabled but without treating them as errors?

file_io.cpp
```diff
@@ -55,7 +57,7 @@ int FileOpenEx(fileTYPE *file, const char *name, int mode, char mute)
 	}
 	else
 	{
-		sprintf(full_path, name);
```
dholm: It's generally considered a security issue to not use a hardcoded format string since if you can manipulate the input string you can get sprintf to dump data from the current context.  For this application I guess it's not a big issue though.

sorgelig:
i'm not against std:: in general. I just don't want to use std:: with every variable declaration. There are simple and clean looking types like int or uint32_t. So i want to use them instead of my_long_space::my_long_type which is cluttering the code making harder to read it.

input.cpp
```diff
@@ -475,7 +478,7 @@ static const int ev2ps2[] =
 	NONE, //191 KEY_F21			
 	NONE, //192 KEY_F22			
 	NONE, //193 KEY_F23			
-	0x5D, //194 U-mlaut on DE mapped to \
```
dholm: The backslash here will result in the next line being commented out completely which I don't think was the intention?

sorgelig:
As i've told i'm open for fixes. So please open new PR(s) with these fixes.

input.cpp
```diff
@@ -986,7 +989,7 @@ typedef struct
 	int      accx, accy;
 }  devInput;
 
-static devInput input[NUMDEV] = { 0 };
```
dholm: Since C++11 you zero-initialize a struct/class with empty braces, Clang complains here as it only partially initializes the struct.

 - https://en.cppreference.com/w/cpp/language/zero_initialization

menu.cpp
```diff
@@ -355,7 +357,9 @@ void siprintbinary(char* buffer, size_t const size, void const * const ptr)
 	unsigned char *b = (unsigned char*)ptr;
 	unsigned char byte;
 	int i, j;
-	memset(buffer, '\0', sizeof(buffer));
```
dholm: The buffer is being memset with the size of a pointer instead of the size of the buffer.

menu.cpp
```diff
@@ -678,14 +679,12 @@ void HandleUI(void)
 		return;
 	}
 	
-	struct RigidDiskBlock *rdb;
```
dholm: Clang complains that this variable can be accessed before being initialized.

input.cpp
sorgelig: In C standard, initialization with { 0 } means initialize all members with 0. So it's pretty much standard way.

menu.cpp
sorgelig: It's dead code left from original MiST version. 
The whole function must be removed.

menu.cpp
sorgelig: By code logic it won't be accessed before assign. Probably dummy declaration with 0 assign may fix the warning.

dholm: Do you prefer `nullptr`, `NULL` or `0` for pointers?

input.cpp
sorgelig: I'm not sure the change to {} will initialize it with 0 if compiled in GCC. This must be verified before change.

menu.cpp
```diff
@@ -2176,7 +2170,7 @@ void HandleUI(void)
 		// use left/right to adjust video position
 		if (left || right) {
 			if ((menusub == 2) || (menusub == 3)) {
-				if (left && (tos_get_video_adjust(menusub - 2) > -100))
```
dholm: `char` is unsigned on ARM so this comparison is always true unless the result of `tos_get_video_adjust` is cast to `signed char`.

input.cpp
sorgelig: you are right. Fortunately it was the last useful key in the array.

input.cpp
dholm: @sorgelig I checked it with `ARM gcc 5.4` on godbolt and it produces the same instructions. https://godbolt.org/z/Fh1vKr

menu.cpp
sorgelig: right, but all MENU_MIST_* cases are used currently as it's related to MiST's Atari ST core which is not ported(yet). MiSTer's Atari ST core doesn't use these menus. So basically it's dead code which will be removed when ST core will become open source. The code is still here because i'm not sure if ST core will be open source or not.

dholm: Oh, I see.

file_io.cpp
sorgelig: MiSTer isn't compiled with security thoughts :)

Makefile
sorgelig: I think it's not bad idea to enable it. I will check it.

input.cpp
sorgelig: good.
I didn't know about this service.

PR #33, opened 2018-12-29
Enable compiler diagnostics during build

dholm:


dholm:
You might want to explicitly disable some warnings as well but I'm not sure which one are generated when you build with GCC since I haven't got it installed at the moment.

sorgelig:
well.. you gave me bubble gum for some time now :)

dholm:
lol

PR #35, opened 2018-12-31
Zip file support

dholm:
Introduces support for loading files stored in zip archives.

Treats zipped files as folders and allows loading files from inside the
zip file.  Files will be decompressed to shm and removed once they are
unloaded.

Current limitations:

 - Does not support nested zip files.
 - All files inside a zip file will be listed regardless of if they are
   inside a sub-folder or not.

### DIFF
--- a/file_io.cpp
+++ b/file_io.cpp
@@ -12,6 +12,8 @@
 #include <sys/vfs.h>
 #include <sys/mman.h>
 #include <linux/magic.h>
+#include <algorithm>
+#include <string>
 #include "osd.h"
 #include "fpga_io.h"
 #include "menu.h"
@@ -20,13 +22,37 @@
 #include "user_io.h"
 #include "cfg.h"
 #include "input.h"
+#include "miniz.h"
 
 int nDirEntries = 0;
 struct dirent DirItem[10000];
+const int maxDirEntries = (int)(sizeof(DirItem)/sizeof(DirItem[0]));
 int iSelectedEntry = 0;       // selected entry index
 int iFirstEntry = 0;
 
 static char full_path[1200];
+
+static bool FileIsZipped(const std::string path, std::string* zip_path, std::string* file_path)
+{
+	const std::string zipext(".zip");
+	auto it = std::search(path.begin(), path.end(),
+						  zipext.begin(), zipext.end(),
+						  [](char ch1, char ch2) { return std::toupper(ch1) == std::toupper(ch2); });
+
+	if (it != path.end())
+	{
+		if (zip_path)
+		{
+			*zip_path = std::string(path.begin(), it + zipext.length());
+		}
+		if (file_path)
+		{
+			*file_path = std::string(it + zipext.length() + 1, path.end());
+		}
+		return true;
+	}
+	return false;
+}
 
 void FileClose(fileTYPE *file)
 {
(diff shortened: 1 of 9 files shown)


dholm:
I'm posting this primarily to get some feedback from @sorgelig.  The code should work but I still need to do some more testing and a little cleaning.

file_io.cpp
```diff
@@ -20,20 +22,44 @@
...
 
 int nDirEntries = 0;
 struct dirent DirItem[10000];
+const int maxDirEntries = (int)(sizeof(DirItem)/sizeof(DirItem[0]));
 int iSelectedEntry = 0;       // selected entry index
 int iFirstEntry = 0;
 
 static char full_path[1200];
 
+static bool FileIsZipped(const std::string path, std::string* zip_path, std::string* file_path)
```
dholm: I used `std::string` to save a few lines of code and having to pass two extra length arguments here.  If you prefer I can rewrite this to use C strings and string functions.

file_io.cpp
```diff
@@ -43,7 +69,7 @@ void FileClose(fileTYPE *file)
 			file->type = 0;
 		}
 	}
-	file->fd = -1;
+	file->filp = nullptr;
 }
 
 int FileOpenEx(fileTYPE *file, const char *name, int mode, char mute)
```
dholm: I can create two helper functions for `FileOpenEx`, one for zipped files and the other for non-zipped and then simply call them from here if you think that is cleaner?

file_io.h
```diff
@@ -1,14 +1,15 @@
...
+#include <stdio.h>
 #include <unistd.h>
 #include <dirent.h>
 #include <fcntl.h>
 #include "spi.h"
 
 typedef struct
 {
-	int       fd;
+	FILE     *filp;
```
dholm: Are you ok with replacing the file descriptor with a file stream in `fileTYPE`?  Without this change I would have to do some ugly hacks in `FileOpenEx` and you can always get the descriptor from the stream by calling `fileno(f->filp)`.

user_io.cpp
```diff
@@ -1304,10 +1304,7 @@ int user_io_file_tx(const char* name, unsigned char index, char opensave, char m
 
 	if (opensave)
 	{
-		strcpy((char*)buf, name);
-		char *p = strrchr((char*)buf, '.');
-		if (!p) p = (char*)buf + strlen(name);
-		strcpy(p, ".sav");
+		FileGenerateSavePath(name, "sav", (char*)buf, sizeof(buf));
```
dholm: Are there any more cases in the code where the path to the image is used to generate a file that is written to?

dholm:
Oh, and happy new year! :)

sorgelig:
What's the usage scenario?
I don't see how it can be useful. Normally you just use FTP to add/remove ROMs. With ZIP it will require too much hassle.

sorgelig:
Also, with zip support users immediately will start to abuse it by using 2GB,5GB,10GB and bigger archives. Then will complain "why it's so slow?" 
Also there is a limit of files per directory can be displayed while ZIPs you download contain thousands files per folder. 
So, i see many problems without meaningful advantages (besides the space).

nihilisticeevee:
I think this is a must have feature, maybe just add in the wiki about the limit per dir and just direct people to the wiki if they ask why its slow. I generally keep each rom zipped separately so its not a problem for me.

Also i want to thank you all for the great work, it makes me happy to see development of this great system!

I hope to one day be able to contribute, i am a computer science major and i love following the development and hope to learn.

dholm:
@sorgelig So what would you like me to fix to consider it useful?

As for performance, I did not notice any real difference between loading compressed vs decompressed files (of about 1MiB).

One thing that I can fix is to be able to read files directly from the archive instead of decompressing them to a temporary file first. Then the performance impact should not be noticeable even for large compressed files.

sorgelig:
Well, i know about dynamic arrays and lists. I also know these dynamic structures fragment the memory and also are primary source of memory leak. 
My original idea was to jail the file list in the fixed block of memory.
Also unlimited file list doesn't bring unlimited speed of sorting the list. So expect the complain "I enter the folder and it hangs for long time."

sorgelig:
Another problem is HPS response time for FPGA. If some systems like Minimig or ao486 will access the hard disk and user will browse such long folders, then any kind of problems may happen, even hard disk corruption due to lost requests.

MiSTer is not a traditional emulator where everything can be suspended in case if OS or concurrent thread is busy. Here you have FPGA which is not suspended and expects replies on time. Even if HPS-FPGA communication is suspended, the emulated system doesn't expect it and may fail by timeouts. So you cannot blindly add features as you like.

As for ZIP support the concept needs to be revised later. I don't like the idea to support huge ZIPs.

PR #36, opened 2019-01-01
Fix high cpu usage

paulb-nl:


sorgelig:
I've replied in forum.
HPS has no power manager, so this sleep won't make any difference in power consumption.

PR #37, opened 2019-01-01
file_io: Support unlimited number of directory entries

dholm:
Stores directory entries in a std::vector instead of an array of
hardcoded size.  The vector will dynamically grow to the number of
available entries in a folder without having a hardcoded upper limit.

### DIFF
--- a/file_io.cpp
+++ b/file_io.cpp
@@ -12,6 +12,8 @@
 #include <sys/vfs.h>
 #include <sys/mman.h>
 #include <linux/magic.h>
+#include <algorithm>
+#include <vector>
 #include "osd.h"
 #include "fpga_io.h"
 #include "menu.h"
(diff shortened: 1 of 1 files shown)


dholm:
@sorgelig Regarding your comments from PR #35.

It's possible to pre-allocate space for the vector if you want to avoid fragmentation. Do you want me to submit a pull-request that does that or should we wait and see if it is an issue?

If you have latency concerns in the main loop then it's possible to use a coroutine for `HandleUI`.  Then it would be able to yield during long-running tasks, such as loading direntries or sorting them, while still maintaining a run-to-completion model and without having to use threads.  If you are interested in this I can try to put something together for you to review?

sorgelig:
I remember some implementations (like MFC) of dynamic structures have granularity settings so it will pre-allocate the space for next several items which can be defined. So it would be good to define if it's possible.

I'm not sure what you mean by coroutine. You may make PR - so i will look at it.

PR #38, opened 2019-01-04
main: Always pin to core #1

dholm:
Pin the MiSTer process to core #1 in order to stay clear of the hardware
interrupt handlers running in the Linux kernel on core #0.  This gives a
consistent improvement in the main loop cycle time of about 6-7x.

### DIFF
--- a/main.cpp
+++ b/main.cpp
@@ -35,22 +35,13 @@ const char *version = "$VER:HPS" VDATE;
 
 int main(int argc, char *argv[])
 {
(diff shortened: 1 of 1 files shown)


dholm:
I did some crude measurements with `clock_gettime` in the main loop and when MiSTer is running on core 0 the maximum loop time each second is around 630μs but if I pin it to core 1 it is around 70-80μs.  The results are consistent through numerous test executions.

PR #39, opened 2019-01-04
Main loop latency reduction with coroutines

dholm:
Schedules tasks in the main loop using coroutines so that long running
tasks, particularly relating to the UI, can yield execution in order to
call the poll functions at a tighter interval.

Yields at an interval during the loading of directory entries as well as
while sorting them in order to avoid starving the poll functions while
entering directories containing a large number of files.

The maximum time between calls to `user_io_poll` in the main loop were
measured while loading a folder containing 29026 entries:

 - Original main-loop:       591ms
 - With coroutines + yields:  29ms

The amount of time it takes to load and sort the 29k entries:

 - Original main-loop:       591ms
 - With coroutines + yields: 686ms

By increasing the value of `YieldIterations` in `file_io.cpp` the load
time will decrease while increasing the delay between calls to
`user_io_poll`.

### DIFF
--- a/3pp/libco/amd64.c
+++ b/3pp/libco/amd64.c
@@ -1,0 +1,160 @@
+/*
+  libco.amd64 (2016-09-14)
+  author: byuu
+  license: public domain
+*/
+
+#define LIBCO_C
+#include "libco.h"
+#include "settings.h"
+
+#include <assert.h>
+#include <stdlib.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+static thread_local long long co_active_buffer[64];
+static thread_local cothread_t co_active_handle = 0;
+static void (*co_swap)(cothread_t, cothread_t) = 0;
+
+#ifdef LIBCO_MPROTECT
+  alignas(4096)
+#else
+  section(text)
+#endif
+#ifdef _WIN32
+  /* ABI: Win64 */
+  static const unsigned char co_swap_function[4096] = {
+    0x48, 0x89, 0x22,              /* mov [rdx],rsp          */
+    0x48, 0x8b, 0x21,              /* mov rsp,[rcx]          */
+    0x58,                          /* pop rax                */
+    0x48, 0x89, 0x6a, 0x08,        /* mov [rdx+ 8],rbp       */
+    0x48, 0x89, 0x72, 0x10,        /* mov [rdx+16],rsi       */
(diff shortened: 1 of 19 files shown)


sorgelig:
Looks interesting. Never heard about co-routines before.

dholm:
It's starting to grow in popularity now because it is very effective if you are developing services that need to serve large number of requests with low overhead.  It is one of the fundamental features of Go (the Google programming language) where it is called goroutines.

I'm considering taking a look at the input handling next.  The system calls are the main contributors to the latency and one way to improve this would be to run the `poll` from `input_test` in a separate thread pinned to core 0 and then push events through a lock-free queue to the input code running in the main loop.  I'll try to put something simple together for you to review, unless you think it's unnecessary?

sorgelig:
it seems you didn't pull my changes before you made PR - so all my fixes against warnings were reverted..

dholm:
Are you sure?  I rebased just before I pushed this change to GitHub.

sorgelig:
False alarm. The warnings were from other file i didn't tweak yet. Strange there was no warnings before. Anyway, it's fixed now.

sorgelig:
is this code:
```
	co_delete(co_ui);
	co_delete(co_poll);
	co_delete(co_scheduler);
```
really executed? It looks like dead code.

sorgelig:
btw, it seems you are the first one who didn't start with "let's add sleep(x) in the main thread to reduce CPU usage" ;)

dholm:
> is this code:
> 
> ```
> 	co_delete(co_ui);
> 	co_delete(co_poll);
> 	co_delete(co_scheduler);
> ```
> really executed? It looks like dead code.

It's dead but I always add cleanup code in the proper places.  If someone wants to change the application in the future so that it can do a clean shutdown they don't have to worry about where the cleanup code is supposed to be.

dholm:
> btw, it seems you are the first one who didn't start with "let's add sleep(x) in the main thread to reduce CPU usage" ;)

There are so many better things you can do with CPU cycles other than wasting them on sleeping. :)

sorgelig:
Especially when CPU has no PM and cannot sleep :)

sorgelig:
so coroutines is lightweight threading with one important feature - you can control at which places thread can be switched to another one.
Too bad it also requires splitting the stack so need to be careful not to overflow it.

dholm:
Sure, I will take a stab at it after I'm done with the latency reduction stuff in `input_poll`.

PR #44, opened 2019-01-17
Change MidiLink order /FSYNTH/MUNT/TCP/UDP

bbond007:
"2) After each start of ao486, the Midilink resets to TCP mode. Would be good to keep it as local MIDI emulation now will be preferred mode."

### DIFF
--- a/menu.cpp
+++ b/menu.cpp
@@ -168,7 +168,7 @@ const char *joy_button_map[] = { "RIGHT", "LEFT", "DOWN", "UP", "BUTTON 1", "BUT
 const char *config_stereo_msg[] = { "0%", "25%", "50%", "100%" };
 const char *config_uart_msg[] = { "     None", "      PPP", "  Console", "    MLINK", "MLINK-38K" };
 const char *config_scaler_msg[] = { "Internal","Custom" };
-const char *config_softsynth_msg[] = {"       TCP", "       UDP", "      MUNT", "FluidSynth"};
+const char *config_softsynth_msg[] = {"FluidSynth", "      MUNT", "       TCP", "       UDP"};
 
 char joy_bnames[12][32];
 int  joy_bcount = 0;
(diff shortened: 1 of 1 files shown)


sorgelig:
So, this option is not saved like UART mode?

bbond007:
Because I was using a USB sound board, there was no guarantee it would
still be installed, which is why I did save and made the TCP the default.

Thanks

sorgelig:
Oh, i see. Current implementation of sound board will always work even if current core doesn't support ALSA sound. You even can switch the cores with ALSA audio working and it will be caught by core if it supported.

bbond007:
Right now the menu creates files:

tmp/ML_TCP
tmp/ML_UDP
tmp/ML_FSYNTH
tmp/ML_MUNT

If MidiLink is started with the "MENU" option then it looks for those
files. If none exist, then in now assumes FluidSynth.

If MidiLink finds a real USB MIDI device, it ignores other options.

I would assume if you have a real MIDI, you intend on using it and not
emulation.

I think it would be better if read directly from config file, but then
MidiLink needs to know what core is running so it reads the correct file.

I can refine this better, but its going to take me some time as I'm not
proficient with the menu code. I promise I will  try and improve.

sorgelig:
Do you know tmp is ramdisk? Nothing survive reset/powercycle there.
With working MIDI emulation it's more actual now to have different configs for different cores. For Minimig i think i will mostly use PPP while MIDI is more suitable for ao486.

Actually i need to refresh my memory about UART mode settings :)
It's possible to make a separate OSD page for UART if it will be more convenient. Currently every change to UART mode calls heavy task of killing/creating the processes. May be separate page where settings can be made before apply would be better. What do u think?

bbond007:
yes, I know its a ram disk :)

I think now that it is an official feature, yes, I agree dedicated
configuration menu off the system menu makes a lot of sense.

During development I was really trying to have as small of footprint as
possible in the MiSTer core - which changes frequently :)

As far as Minimig, PPP was my favorite, but now TCP is because I can easily
download files from Linux (using ATRZ command) and play MP3s (ATMP3)

I plan to make a AmiAmp like front-end (in probably sas c) that works with
the TCP option.

Thanks!

PR #50, opened 2019-01-18
Add symbolic links handling

D0023R:
Hi Sorgelig,

I know about your position with respect to the symbolic links (topic I have addressed long ago in the forum). Nevertheless, the following change enables the file selector to show and navigate the linked files and folders. If the change looks fine to you feel free to merge it.

Cheers,

Doozer

### DIFF
--- a/file_io.cpp
+++ b/file_io.cpp
@@ -1122,6 +1122,27 @@ int ScanDirectory(char* path, int mode, const char *extension, int options, cons
 				de = &_de;
 			}
 
+			if (de->d_type == DT_LNK)
+			// Handle symbolic link type in the directory entry 
+			{
+				char *new_full_path = (char *) malloc (strlen(full_path)+strlen(de->d_name)+2);
+				sprintf(new_full_path, "%s/%s", full_path, de->d_name);
+				struct stat entrystat;
+
+				if (!stat( new_full_path, &entrystat ))
+			        {
+					if ( S_ISREG(entrystat.st_mode))
+					{
+						de->d_type = DT_REG;
+					}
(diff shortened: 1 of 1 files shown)


sorgelig:
I've just checked this code. readdir() on exFAT volume doesn't return DT_LNK type of links. 
Although in shell link is properly shown as link.

D0023R:
Hi,

I don't know the situation on exFAT. But readdir returns DT_LNK on any Unix like filesystem.
Personally, I am using my MiSTer with NFS and the symbolic links are processed correctly with the provided change.
Please note that the code does not change the fileselector behaviour on Windows based filesystem. So in it last current form it is harmless and not used on exFat. I assume FAT fs will require a complementary code to honor the Microsoft file based shortcuts. 

Regards,







Le 18 janvier 2019 17:24:05 GMT+01:00, sorgelig <notifications@github.com> a écrit :
>I've just checked this code. readdir() on exFAT volume doesn't return
>DT_LNK type of links. 
>Although in shell link is properly shown as link.
>
>-- 
>You are receiving this because you authored the thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/MiSTer-devel/Main_MiSTer/pull/50#issuecomment-455604038

-- 
Envoyé de mon appareil Android avec Courriel K-9 Mail. Veuillez excuser ma brièveté.

sorgelig:
I've already enhanced the link support code. It works on exFAT now as well.
Linux uses other approach for link files on FAT than windows. It's just a text file with path to target inside. Since OS supports it, i don't see why MiSTer shouldn't.

PR #54, opened 2019-02-04
Fix PAD emulation on axis 0/1

D0023R:
Hi,

My understanding is that axes 0/1 should also be taken into account for emulation of the DPAD. At least it is was the comment says. Without the proposed change the offset variable is incremented twice and I am not sure it is the intended behavior. 
I have tested the fix in OSD menu, some Cores and with mouse emulation. It is working as expected and I do not experience any odd behavior.
Let me know if the removed line has a required purpose.

Regards,

### DIFF
--- a/input.cpp
+++ b/input.cpp
@@ -2185,7 +2185,6 @@ int input_test(int getchar)
 								char offset_mask1, offset_mask2;
 
 								if (ev.code < 16) offset += 4;
-								if (ev.code < 2)  offset += 4;
 								offset_mask1 = (1 << (offset >> 1)); // 01XX
 								offset_mask2 = (2 << (offset >> 1)); // 10XX
 


sorgelig:
Yes, it's on purpose.
Some analog axis can emulate the buttons. So depending on axe number, the emulated button ID is different. Analog 0/1 emulate the DPAD but the buttons ID are different from original DPAD, so it will be up to the user how he define it.

D0023R:
Hi,

If I understand you well, the DPAD emulation is supposed to enable the user to assign button ID to the analogue sticks. I am fine with this.

I just want to mention that the MiSTer default behaviour already use the 2nd analogue stick as a digital directional pad. It means that you can use the cross pad and the 2nd stick simultaneously to control the directions without specific configuration. Nevertheless, I find it not convenient to have the directional keys located on the right hand side of the controller together with the main buttons. 

Is there a method to invert the sticks and have the directions emulated on the left sticks? Otherwise, could this be made available as an option inside the ini?

sorgelig:
In Menu core you define 2 sets of directional buttons. Set first set for dpad, and second set (Alt/Mouse) for left analog stick.
Then left analog stick will work as alternative dpad for all cores.

D0023R:
Thanks a lot for the explanation. 

I confirm the Alt/Mouse configuration is working well and doing the intended job.

PR #60, opened 2019-03-03
lastcoreboot feature

spark2k06:
Allows to autoboot the last used core.


sorgelig:
1) You unconditionally write to config file with every core boot even when feature is not enabled. This is not acceptable. Even if enabled you have to avoid to write the same data to file. This is flash memory, not HDD.
2) With lastcoreboot=1 how to boot into Menu?
3) After solving 1 and 2. You need to re-fork the code, apply clean patch and make new PR. Please avoid making trashcan in commit history.

I close this PR. You are free to make a new PR when will be ready.

spark2k06:
1. You're right. I will solve it and in addition I will add an improvement to the feature.

2. This is not a problem, the last core is only loaded when the main function does not receive arguments, that is, when the Mister is poweron at first time (and also when resetting from the IO board). As you can see in the demonstration video (https://www.youtube.com/watch?v=GWp-JB4359Q), you can return to the main menu through a cold reset from a core menu (F12), this is possible because I force a parameter "menu.rbf" instead of NULL in the restart of the application after loading the core "menu.rbf" -> app_restart (! strcasecmp (name, "menu.rbf")? "menu.rbf": path) ;. In addition, then this core "menu.rbf" is used as lastcore, so that if the mister is turned off, when it turns on, it returns to it.

3. Thanks for the tips. I will do a re-fork and I will make the changes in my github but before doing pull request I will leave it in beta for a while with a new functionality that I will announce in the corresponding atari-forum thread.

PR #64, opened 2019-03-11
input: Reduce 8bitdo retro receiver analogue stick dead zone event flood

Optiroc:
Identical fix to the PS3 controller event log flooding, but for playstation controllers connected via the 8bitdo Retro Receiver USB.

Seeing how this is a common (and great!) controller setup for MiSTer, it seems appropriate to make the Main menu adequatly responsive with it.

### DIFF
--- a/input.cpp
+++ b/input.cpp
@@ -2154,6 +2154,10 @@ int input_test(int getchar)
 									if (input[i].vid == 0x054c && input[i].pid == 0x0268)
 									{ if (ev.code <= 5 && ev.value > 118 && ev.value < 138) break; }
 
+									//reduce flood from PS3/PS4 gamepads via 8bitdo retro receiver
+									if (input[i].vid == 0x2dc8 && input[i].pid == 0x3100)
+									{ if (ev.code <= 5 && ev.value > 118 && ev.value < 138) break; }
+
 									//aliexpress USB encoder floods messages
 									if (input[i].vid == 0x0079 && input[i].pid == 0x0006)
 									{ if (ev.code == 2) break; }


sorgelig:
I need to check PS3+8bitdo myself. I don't like this fix. Basically it makes analog stick useless with such huge dead zone. Some cores use analog input and they will be affected. The VID/PID of 8bitdo receiver is the same regardless the gamepad connected, so you propagate this hack to all gamepads connected through 8bitdo receiver even though they have no event flood problem.
Probably option in ini file can be added as a workaround.

sorgelig:
i found a solution. You need to switch to X-Input mode (keep Select+DPadUP for 3 seconds). In this mode PS3 controller doesn't produce flood.

Optiroc:
Ah, good! 

But perhaps a configurable dead zone should be an ini option, in those cases where the user wants analogue inputs enabled?

James-F2:
Is this issue still valid despite the wiki recommends 8bitdo RR with Dualshock3/4?

Is there a difference between grey/orange RR versus the Console RR for the mister usage?
The console rr can also be connected to the mister via micro usb cable.
When I connect the console rr to a windows system via usb, it is always reported as x-input (xbox360) controller no matter what gamepad I use.

sorgelig:
> 
> 
> Ah, good!
> 
> But perhaps a configurable dead zone should be an ini option, in those cases where the user wants analogue inputs enabled?

the solution i've posted doesn't exclude analog input. It's fully working as analog but doesn't produce flood. After pairing 8bitdo receiver with PS3 gamepad at gets other VID:PID (like in your patch) and has message flood. But once you switch to X-Input mode it stops flooding and start to work normally. 

> Is this issue still valid despite the wiki recommends 8bitdo RR with Dualshock3/4?
after switching to X-Input issue is gone.

> Is there a difference between grey/orange RR versus the Console RR for the mister usage?
i don't know. Probably there is a difference as firmware is different (and HW probably too).

Optiroc:
I haven't used the analogue sticks for anything thus far, but for the playstation analogue input to respond well without being jittery around the dead zone a simplified version of [this technique](https://www.gamasutra.com/blogs/JoshSutphin/20130416/190541/Doing_Thumbstick_Dead_Zones_Right.php) could be implemented.

The difference is that the dead zone would be diamond shaped rather than circular, with the benefit of simpler maths and no knowledge of the axis pairs needed.

I jotted down some python as a test:
```
def deadzone_scale(val, deadzone):
  if abs(val - 128) < deadzone:
    return 128
  else:
    # scaling factor can be precalced at config load to avoid the div
    scaling = 255 / (255 - deadzone - deadzone)
    if val < 128:
      return 128 + ((val - 128 + deadzone) * scaling)
    else:
      return 128 + ((val - 128 - deadzone) * scaling)
```

sorgelig:
Not sure if you read my messages. It seems no.

Optiroc:
Xinput on the Retro Receiver does indeed play a lot nicer with MiSTer than the Dinput mode. It’s a bit counterintuitive since Dinput looks like the most basic mode in the RR docs, so a note about this in the wiki wouldn’t hurt. 

I don’t know if it’s possible to make PRs on the wiki, but I could try. Thanks for an awesome project!

sorgelig:
Wiki allows direct edit

James-F2:
I've edited the Input wiki page.

PR #73, opened 2019-05-02
Added INI options to hide script warning

wesclemens:
@Locutus73 suggested this in the Scripting news thread.

### DIFF
--- a/MiSTer.ini
+++ b/MiSTer.ini
@@ -1,3 +1,4 @@
+; The global setting section is 'MiSTer'. Additinal sections with core names (e.g. Minimig, NES) can be added for pre-core settings/overrides.
 [MiSTer]
 key_menu_as_rgui=0     ; set to 1 to make the MENU key map to RGUI in Minimig (e.g. for Right Amiga)
 forced_scandoubler=0   ; set to 1 to run scandoubler on VGA output always (depends on core).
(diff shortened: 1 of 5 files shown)


sorgelig:
Sorry, this warning should not be hidden.
Scripts are not for regular use like just another menu extension.

Locutus73:
I like disclaimers, they empower users and make script developers life easier. But the idea here was: _script_hide_warn_ option in MiSTer.ini (default value 0, false, so warning page enabled) in order to let users, who understand what they're doing, to willingly disable the warning in order to have a quicker access to scripts. A quality of life improvement for users who understand what they're doing; in the end manually editing the ini is just like signing a permanent agreement to the risks exposed by the warning screen.

sorgelig:
In this case some users will get so-called "MiSTer pack" from somewhere with this option set in the beginning. And will never see this warning.

Locutus73:
I see your point, you're right. Ok, I'll try to think something better which can't becircumvented with a commonly downloaded file, forcing the warning to all users until they willingly accept a "don't show me that again" clause.

wesclemens:
I have some ideas how we can ensure the message was accepted via device using the SD card info (/sys/block/mmcblk0/device/cid).

@sorgelig If you are ok with only showing this message once I would love to give it a try.

sorgelig:
i will check it

Locutus73:
@wesclemens the idea of an individual, per SD, “semaphore file” is clever. If I can add my two cents, maybe a second page could be added after the warning page. So the first page warns the user about the potential risks of using scripts and asks if he/she wants to go ahead (default choice no), and the second page (if he/she answered yes to the first) asks if he/she wants this warning the next time (default choice yes). The semaphore file could be something in /media/fat/linux or /media/fat/config or in the Linux root filesystem too (but you have to check if it’s RO, and eventually remount it RW and then RO again) so that @sorgelig could add an empty (invalid) one in the SD-installer and, at each Linux update the warning would be on again.

wesclemens:
@locutus73 Maybe just a third options on the warning message like "Yes, don't ask again". I don't like the idea of adding a second dialog as if you want it to show each time you have two menus that need pass though to run your script.  I defiantly feel like this would be best under config directory.  If we wish to re-prompt after updates we can check a simple hash of the MiSTer binary and Linux image. I will update my branch with these changes and we can evaluate from there.

Locutus73:
@wesclemens Good point about the two prompts each time you run a script if you don't disable the warning screen. So the options would be No, Yes (show the warning again) and Yes (don't show the warning again). The hash of MiSTer binary is another good idea, I'd avoid hashing the (big and possibly changing) Linux image.

PR #75, opened 2019-05-07
Add support for FC

alanswx:
FC -  C stands for CODES. This code will cause the directory path to start in:
codes/{HomeDir} instead of {HomeDir}
It also assumes the last file open might be a rom (if it didn't just load something out of codes}, so it parses apart the file name and tries to open a zip of the same name in the codes directory.

CODES: before: SelectedPath [SNES/T/Tetris Attack (USA).sfc]
CODES: codesfile[codes/SNES/T/Tetris Attack (USA).sfc]
CODES: codesfile[codes/SNES/T/Tetris Attack (USA).zip]
CODES: codesfileroot[codes/SNES/Tetris Attack (USA).zip]
If the zip doesn't exist, it will go through the default loading which will set SelectedPath to codes/SNES


James-F2:
May make a small suggestion.

> If` the zip doesn't exist, it will go through the default loading which will set SelectedPath to codes/SNES.

If possible, if .zip doesn't exist fallback to codes/snes/gamename(dir)/  first,, and if that doesn't exist, then to codes/snes/

alanswx:
https://gamehacking.org/ added MiSTer support - you can download .zip files with all the .gg files inside. I am not sure why you would need to unzip the file.

James-F2:
I manually select what code I download from gamehacking.org but some codes do not work, or do not do what they supposed to do.
It is much easier to manage individual code files if they are in a folder and not in a zip.

alanswx:
I will take a look at this tomorrow. I will unzip the directory and make sure it does the right thing. Thank you for looking at it.

sorgelig:
current Game Genie support is in intermediate stage. It's more likely will be changed to a single text file per game file name. So there is no reason to change MISTer right now.

Kitrinx:
If you really plan to make a text file, then you'll have to be careful about it. There's a lot of different kinds of codes for each system, and maintaining such a file by hand would be a nightmare, so use some kind of format like JSON or YAML. At least that way you can attempt to match the game name with the rom like Alan's PR does. You have to account for every different kind of code, gameshark, action replay, raw, etc, otherwise it will be a regression. Some codes are multiple lines, so that also is a consideration, and you'll have to figure out which core the code belongs to. Right now with this system we can toggle codes, get all the different kinds of codes for all the different systems, and it can locate the correct game to display to the user. It could definitely be better, but unless you plan to replace this right all right away to a point where it's feature-equivalent to what we have now, you should probably merge this.

When this came up on the forums, you didn't seem interested in working on codes.  If you had mentioned that you'd be willing to write a menu when I asked a few times, it could have saved a lot of trouble.

James-F2:
The current solution is very easy to use and is user friendly.
I have tested this thoroughly with all supported cores and it is already much better and easier than any flashcart.

sorgelig:
@Kitrinx,

> When this came up on the forums, you didn't seem interested in working on codes. If you had mentioned that you'd be willing to write a menu when I asked a few times, it could have saved a lot of trouble.

There are many ideas coming to the forum. In most cases people just throw an idea without willing to implement by themselves. So they just think i will rush to implement it while i'm may be not interested in this. Sometimes i see some features are developed by some one and i'm willing to help to integrate it smoothly, especially if it requires system-wide changes. It's like it happened with lightgun support. You did a core support and i willing to add it on system level, so now Wiimote is supported natively (not release yet). Would i do it originally - nope.
I don't see what troubles i could save as i'm not going to rewrite or replace what you did with Cheat Engine support.
And since it's been added to several cores i need to think how HPS side can make it easier.
Ok, if you think text version of codes aren't worth to support, i will just use zip files with *.gg file inside. I've just though text codes are much easier to add by hand than relay on just one specific site.

Kitrinx:
You are correct in that passing the values through HPS is a lot cleaner than having to use file loading, and ultimately, the right way to do this. I think in the end text files are the right move, and that was everyone's first instinct, and I'm in favor of anything that will make it easier and more intuitive for the user. I just wish you had brought it up before we made the effort to try to make the menu nicer for the users.

One thing to consider is how to communicate from the core to the linux side, for things like how many slots are used, what codes are already enabled, or even just a simple way to trigger a message telling people if the code is being enabled or disabled. The alternative is maintaining state on the linux side and sending all the codes over at once whenever there is a change, replacing the core's saved valued.

We can discuss it more on the forums.

sorgelig:
As you told before, i want to move step by step. 
Probably i won't change file I/O method at this stage but will introduce some additional flags to the core. 
I wont to make a separate OSD option for cheats. Since i'm not familiar with cheat codes formats i will use .gg files for now. Later, when i will have more info it can be moved to text version. 

May be someone (you?) will write a good converter from text version to binary so it will be integrated to HPS part. 
Text format can be like:
CODE CODE CODE *Description
So it will allow to write several codes in one line. Each line is separate cheat.

Kitrinx:
That's what Alan and I decided too. To go one step at a time. I will see if I can find a good reference for all the different kinds of codes. Every system has between 2-4 different kinds of codes for it.

I think no matter what, we want the cheats to live in /cheats/system/ folder so they don't get mixed up with roms, the same as .sav files.

alanswx:
I am happy to rewrite this patch the way you would like it.

Here is how we ended up in this spot:

The obvious thing would be to make a text file with all the codes in it.
there is some source:
http://www.wyrmcorp.com/software/uggconv/index.shtml
and some javascript to convert the codes
https://github.com/alanswx/GameGenieCodeFormatter
Kitrinx and I came up with a binary format that makes sense in verilog - so the HPS can send it over via SPI. For now we used the file code, this is a shortcut.

I then realized that retroarch has a .cht format in text, but their repo has the .cht files, and they are lacking a lot of codes. I contacted the retroarch team, and the would be happy for us to issue pull requests to put more codes into their files. This is a possibility. 

Meanwhile, Kitrinx contacted gamehacking.org and found he was a big MiSTer fan, and agreed to format the codes in his giant database in any format we wanted.

Once we had integration with his database, it didn't seem like we needed to edit the codes by hand, since he has SO MANY, and if they are missing you can create an account and add them to his website. He has a web interface already, so super simple.

The thing that is lacking with this implementation is state. You don't know what codes are turned on or off. What would be nice would be to have a list stored in memory of the codes for the particular game. And we could use the checkbox font in the OSD to toggle the cheats on/off. Then we could use the SPI bus to send all the cheats to the game genie on the verilog side, and it could replace all the codes. This would be the cleanest / best implementation.

I originally thought the codes should be in one file /media/fat/NES/codes.csv - but once I started building the files I realized how many codes there are, the file is huge. 

The super simple thing to do is match the name of the ROM to the name of the cht / gg file, then people can't either use the file dialog to open it or they can make sure the .cht files are named the same as their rom (easy step) and we can automatically open it.  ie: if it isn't named the same, it still works, but if it is, it works more seamlessly.

Looking forward to your feedback / direction.

sorgelig:
Ok, now i see you work together with it.
My intention is to remove file browsing. One zip file with the same name as ROM is enough. So if file is found, then option is available, if not found then it's grayed out.
I'm still debating to myself how it will be sent through SPI. 
For code status it can be simple local status regardless the real status in the core. So first send will mark it as active, second send will unmark it. 
Making two-way communication with cheat engine will be over-complicated i think. Some cheats have several codes inside so it will me hard to match it. At the end it's not life saving feature :) We don't need to be 100% sure. Just click once - marked, second - unmarked.

I want to make at least initial cheat code implementation in HPS because i need to evaluate different options. Then improvements are welcome.

sorgelig:
Sending all codes at once probably also good idea, so cheatengine.v won't need to check if code activated or not. Just need to define what to send to clear all codes. Probably some dummy cheat doing nothing. Is there such?

Kitrinx:
If you just resend all codes, I can just make it so the core's module clears it's buffer when it gets new codes sent, so that's just as good. All state can be on the linux side.

For HPS, all codes have the same four elements, in the end: flags (has_compare), address, compare value, and replace value.  If you make each of those about 32 bits, it will cover all present and future systems mister might have. (GBA or PS1 might use 32 bits). Probably the easiest way to do it is to just assemble them in HPS_IO and clock them in with the leading bit, kind of like PS2 mouse. That follows the same tactic I use with file loading.

alanswx:
I think a matching zip (or directory? See comments above) would be great. I think keeping the state and having checkboxes would be awesome. We can build a JavaScript editor to make the gg files eventually if we want or just use gamehacking

sorgelig:
I've added support for cheat codes to Main and tweaked Genesis core for it.
cheats zip should be placed to either the same folder or to /cheats/<system>/

Codes are always pushed all together with every enable/disable of any code. 
Currently up to 128 codes are pushed. How much codes will be accepted by specific core is up to the core.

Opinions and suggestions are welcome.

sorgelig:
@Kitrinx are you willing to propagate the changes to other cheat enabled cores? I think you are more familiar with cheats, so you can do this better.

James-F2:
Amazing work!
No problem with .zip detection.

gg_zip

Kitrinx:
@sorgelig This menu is excellent.
I will propagate the changes tonight after work. I will also see if I can do some optimizations for SNES.

sorgelig:
I will update SNES myself because i need to test other changes now too. 
You can add optimization later if it will be required.

sorgelig:
@Kitrinx gamehacking site has bug in generating zip. If hacks have the same name then only one gg will be generated. Ask the owner to add index to names or name of hack author, so all versions will be generated.
For example: R-Type III (SNES).

PR #76, opened 2019-05-08
Screenshot using Alt-Print Screen

alanswx:
This is a first proposal for a screen shot utility that saves PNGs inside:
screenshot/GAMEBOY/shot0001.png
etc.
It would be great if it knew the name of the current rom, but I didn't want to make it too much of a hack. It would be nice for this core, and others to have a simple method that returns the name of the current rom if it is relevant. (might not make sense on some of the computers)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,12 +29,13 @@ ST_SRC	= $(wildcard ./support/st/*.cpp)
 X86_SRC	= $(wildcard ./support/x86/*.cpp)
 SNES_SRC	= $(wildcard ./support/snes/*.cpp)
 LIBCO_SRC	= lib/libco/arm.c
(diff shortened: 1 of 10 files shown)


sorgelig:
I will check about rom name

alanswx:
Thank you!

sorgelig:
I'm thinking to remove Alt+PrnScr combination. Win+PrnScr will continue to work. Both Alt and PrnScr keys are used in some cores. In ao486 Win95 it's even used to make a window screenshot. Win key is not available for cores as itself.
Actually i'm thinking to use Win(Alt)+ScrLock. ScrLock is already used to switch the Joystick emulation so it's a good way to re-use it.

PR #78, opened 2019-05-11
Add INI option to set HDMI output to Limited RGB

CruelTott:
This tells the core to configure the HDMI output to Limited RGB range.

I've submitted a Pull Request against Menu_MiSTer for the FPGA side changes.


sorgelig:
I guess it's not really required to put to every core

CruelTott:
> I guess it's not really required to put to every core

The colors will always be slightly wrong on a Limited screen unless it's put in.
But it will most likely be hard to notice unless the core has many shades of color close to maximum darkness or maximum brightness.

Kitrinx:
It's probably a good idea to propagate it with /sys changes. Limited vs Full range colors is a fairly staple compatibility feature of HDMI if you're using it on a TV. Some TV's won't react well, will show the wrong colors, or sometimes will even create lag if they get full range.

PR #79, opened 2019-05-18
Allow both CPU cores during scripts execution

Locutus73:


Locutus73:
Hi, Sorgelig.
Today I released an updater which does parallel update/download processing and I made some performance measurements: I just discovered that the anything launched from the main MiSTer, i.e. scripts, is tied to the same single (already running at 100%) CPU core as the GUI, so scripts are slower when executed from main MiSTer than from SSH.
i.e. for a full global update with nothing new:
================================
SSH    PARALLEL_UPDATE=“false”    3:59
SSH    PARALLEL_UPDATE=“true”    0:57

GUI    PARALLEL_UPDATE=“false”    4:42
GUI    PARALLEL_UPDATE=“True”    1:22
================================
After applying this patch, main MiSTer Scripts menu GUI times are the same as SSH.

Regards.

Locutus73

PR #83, opened 2019-05-27
Framebuffer terminal Scripts menu (not finished)

Locutus73:
It sort of works, but when I exit the script and re-enter the GUI, MiSTer acts weirdly, just like if a button was pressed.


sorgelig:
I'm still working on this part. Let me finish it first :)

PR #84, opened 2019-05-28
changed Amiga keymap to also work on German keyboards

nretro:
Two keys on german keyboard (#,<) were not mapped. I also did put the keypad [ and ] keys on previously unmapped pgup and pgdown. Those keys are also mapped to F_13 and F_14 which probably few keyboards have.

### DIFF
--- a/input.cpp
+++ b/input.cpp
@@ -117,7 +117,7 @@ static int ev2amiga[] =
 	0x3c, //83  KEY_KPDOT		
 	NONE, //84  ???				
 	NONE, //85  KEY_ZENKAKU		
-	NONE, //86  KEY_102ND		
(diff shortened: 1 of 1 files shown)


nretro:
Opps, # was mapped before to CTRL. Since ctrl on PC keyboard is also mapped to CTRL, I guess nobody would need that particular mapping.

sorgelig:
F24 is related to Keyrah, IIRC
I accept your PR, but i may revert it later if i'll find incompatibility with US/UK layout or Keyrah.

nretro:
If it is needed for Keyrah, I would suggest to make it configurable. On Amiga '#' is needed for wildcards.

kowoba:
Thank you so much for this!

PR #87, opened 2019-06-04
Adds quiet boot option to hide menu duing countdown

Guddler:
Apologies if any of this is a bit hacky! The idea is basically to have an option similar to git where if a core is set to auto-boot and there is a FB background then the menu can be hidden and you effectively get a splash screen for the countdown. Any key interrupts the countdown and shows the menu.


Guddler:
I think my update is causing problems when cores first start. Minimig is massively slow until you bring up the menu and close it again, C64 core also being strange. I think the menu is still grabbing input. Probably the way I enabled keyboard input for the menu when it was hidden.

Investigating now...

PR #88, opened 2019-06-04
Revert "Adds quiet boot option to hide menu duing countdown"

Guddler:
Reverts MiSTer-devel/Main_MiSTer#87

You should probably revert this for now unless the fix is simple. The way I have forced user input to work while the OSD is hidden (to wake the OSD with any key) is causing problems with the cores when they start because the OSD is still grabbing user input and I haven't yet found the solution.


sorgelig:
ok.

Issue #89, opened 2019-06-06
Proposal: add disable_autofire to MiSTer.ini


bootsector:
Hi!

Some controllers/adapters lacking the HOME/MENU button, but providing combos for it (like START + SELECT), have trouble to get the menu to show up. Instead we get auto-fire enabled to either SELECT or START.

My proposal is to add a disable_autofire (defaults to 0) option to MiSTer.ini that will disable auto-fire and make those controllers having the HOME button key combo to work fine (at the expense of losing auto-fire).

Let me know if that's something you think it's fair so I can implement, test and open a PR.

Cheers!

bootsector:
Thanks for the feedback!

Yeah, a slightly different way to activate auto-fire like, for example, having LEFT + FIRE BUTTON + MENU instead of only FIRE BUTTON + MENU would work as well in my use case.

Thanks for looking at this!

sorgelig:
i will handle it.

sorgelig:
Implemented without needing this option.

PR #90, opened 2019-06-06
Added INI setting to disable auto-fire support. 

bootsector:
Allows for more compatibility with MENU/HOME button combos in some joysticks and adapters.

This should resolve https://github.com/MiSTer-devel/Main_MiSTer/issues/89

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,6 +19,7 @@
 
 # ignore directories
 Debug
+.vscode
 
 host
 *.lst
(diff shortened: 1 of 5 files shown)


sorgelig:
I think it's half-baked solution. What if someone has several gamepads and still want to use auto fire on gamepad where separate Menu button available?

sorgelig:
I have some plans to extend Menu combos. Currently all 4 dpad directions are used to adjust the speed while actually only 2 are enough. Another 2 can be used for other functions later. So, name it as "disable_autofire" is not future proof name. Need some name related to menu button - i didn't find it yet.

sorgelig:
Ok. i will re-work the code defining Menu button to allow define the combo. This will automatically disable auto-fire and other menu button related functions.
So, i close this PR.

PR #99, opened 2019-07-31
Hide hidden files in file browser

MP2E:
This hides system files that macOS and Windows add to the SD card. Also hides any folders designated as hidden by Linux.


MP2E:
Sorry for all the force pushes. Was getting some live feedback from the Classic Gaming Discord. Now it should be in good shape.

sorgelig:
actually if compare ".." first then you don't need to compare against ".." second time.

MP2E:
Good point! Opening a new PR to simplify it

PR #115, opened 2019-09-22
Update MiSTer.ini

PhantombrainM:
Add direct video mode and description


sorgelig:
I will add it myself.. Your description is not how it's supposed to be.

PR #116, opened 2019-09-23
Improvements to joystick mapping

Newsdee:
1) Display gamepad names instead of VID:PID
    (if recognized as one of the tested gamepads)

2) Allow some cores to derive mapping from SNES mapping
    (gives a better default mapping that respects physical button location across pads)
    (see attached image for example how the layout corresponds across controllers)

MiSTer_controllers


CodePoetSarah:
It would be good if we could provide files with settings for all makes and models of joystick, so as someone in the community finds a good setup for a particular joystick, they can share them.

Newsdee:
The idea for this was originated after doing some work to convert config files. Ultimately the current approach requires that we provide one file per joystick and per core, and the list quickly becomes large. 

My suggestion (which, granted, isn't fully implemented in this PR) is instead for MiSTer to have one internal "virtual gamepad" assumed to be 4 face buttons, Start/Select, and two to four shoulder buttons (basically, physical layout of SNES/PS/XBox).  Then, we can map physical controllers to this internal representation, and cores can configure themselves against the same API. This would eliminate the need of config file per core, except for the case of overrides.

sorgelig:
This assumes all users will be satisfied the predefined map. But when not, the system will have to provide the settings which will even more mess the internal configs. And it becomes worse when gamepad layout is different than SNES. Even Genesis gamepad already is not so obvious in map.
Also this approach will require to modify all existing cores. 

I use MiSTer myself. And current gamepad setup very universal. I can grab any gamepad and configure it right away without waiting when it will be added to support list.

sorgelig:
anyway, lets move to the forum where you've already posted.

PR #118, opened 2019-09-28
Menu navigation enhancement

rikard-softgear:
Pressing [UP] when on first menu item puts you at the last.
Pressing [DOWN] when on the last menu item puts you on the first.


James-F2:
That's a very useful feature, many people asking for that.
Thank you.

PR #124, opened 2019-10-20
Add SNES Virtual gamepad to central mapping

Newsdee:
1) Menu core will request A, B, X, Y, L, R, Select, Start
2) Tweaked OSD labels to be more intuitive
3) Moved input device struct to its own .h


sorgelig:
Please read my comment on forum.

PR #125, opened 2019-10-20
Added virtual SNES gamepad

Newsdee:
Added L, R, Select, and Start buttons to main menu core mapping.

### DIFF
--- a/input.cpp
+++ b/input.cpp
@@ -24,8 +24,8 @@
 #include "fpga_io.h"
 #include "osd.h"
 #include "video.h"
+#include "joymapping.h"
 
-#define NUMDEV 30
 #define NUMPLAYERS 6
 #define UINPUT_NAME "MiSTer virtual input"
 
@@ -988,41 +988,6 @@ enum QUIRK
 	QUIRK_DS4TOUCH,
 };
 
-typedef struct
-{
-	uint16_t vid, pid;
-	uint8_t  led;
-	uint8_t  mouse;
-	uint8_t  axis_edge[256];
-	int8_t   axis_pos[256];
-
-	uint8_t  num;
-	uint8_t  has_map;
-	uint32_t map[32];
-
-	uint8_t  osd_combo;
-
-	uint8_t  has_mmap;
-	uint32_t mmap[32];
-	uint16_t jkmap[1024];
-
-	uint8_t  has_kbdmap;
-	uint8_t  kbdmap[256];
-
-	uint16_t guncal[4];
-
-	int      accx, accy;
-	int      quirk;
-
-	int      lightgun_req;
-	int      lightgun;
-
-	int      bind;
-	char     devname[32];
-	char     uniq[32];
-	char     name[128];
-}  devInput;
-
 static devInput input[NUMDEV] = {};
 
 #define BTN_NUM (sizeof(devInput::map) / sizeof(devInput::map[0]))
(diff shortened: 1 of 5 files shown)


input.cpp
```diff
@@ -1732,6 +1697,7 @@ static void input_cb(struct input_event *ev, struct input_absinfo *absinfo, int
 		{
 			if (is_menu_core() || !FileLoadConfig(get_map_name(dev, 1), &input[dev].map, sizeof(input[dev].map)))
 			{
+				// nothing found so blank out everything
```
sorgelig: Here "magic" should happen when map will be created by associate core button names to virtual gamepad buttons.

Newsdee: So what we are saying is that we will use the names from _joy_bnames[][]_ in order to map from the internal mapping, right?

Just one caveat: cores name buttons differently so it's not as simple as mapping "A" to "A".
I had a look at several Joy strings for various cores, and we can do it as a lookup table from many names.

```suggestion
> SNES:    "J1,A(SS Fire),B(SS Cursor),X(SS TurboSw),Y(SS Pause),LT(SS Cursor),
>               RT(SS Fire),Select,Start;",

>  NES:     "J1,A,B,Select,Start,FDS,Trigger,Mic,
>               PP 1,PP 2,PP 3,PP 4,PP 5,PP 6,PP 7,PP 8,PP 9,PP 10,PP 11,PP 12;",

> Genesis: "J1,A,B,C,Start,Mode,X,Y,Z;"

> Neogeo:  "J1,A,B,C,D,Start,Select,Coin,ABC;",	

> Apple2:  "J,Fire 1,Fire 2;",
> GB:      "J1,A,B,Select,Start;",
> SMS: 	 "J1,Fire 1,Fire 2,Pause;",
> SAMCoupe:"J,Fire 1,Fire 2;",
> DK:      "J1,Jump,Start 1P,Start 2P;",
> C64:     "J,Button 1,Button 2,Button 3;",
> A2600:   "J1,Fire,Paddle1(x),Paddle2(y),Game Reset,Game Select;",
> TG16:    "J1,Button I,Button II,Select,Run,Button III,Button IV,Button V,Button VI;",
```

So if we want to avoid per-core specific logic, we can create a lookup table like this:

```suggestion
A = A, Fire, Fire1, Button 1, A(SS Fire), Jump, Button I
B = B, Fire2, Button 2, B(SS Cursor), Button II
X = X, Fire3, Button 3, X(SS Cursor), Button III
Y = Y, Y(SS Pause), Button IV
L = L, LT(SS Cursor), Z, Coin, Button V, FDS
R = R, RT(SS Fire), C, ABC, Button VI
Select = Select, Mode, Game Select
Start = Start, Start 1P, Pause, Run
```

Are you okay with this approach?

sorgelig: almost there. Just remove everything in brackets from compare. So A(something) will become simple A.
No need to guess very hard. If some core has very specific and uniq names, then left it undefined. Open the issue in such core, so later it will be updated with more traditional names.

Newsdee: I added a fallback logic where we just use mapping by position. It works well for the SNES core (only LT and RT are not recognized), but we can amend the core to fix the names.

I also added an option to slightly alter logic per core name (face buttons are in different order in these) which should be pretty universal (applies for SMS, Genesis, Neogeo at least).

Kitrinx: Matching things by matching text strings will never work well except for a few nintendo cores. Correct way to do is change the processing of menu joystick mapping strings to allow for spaces, so the core author can set up the default input to match an order in the joystick HPS_IO port., like: B,A,,Start,Select,FDS,MIC etc

sorgelig: This matching has no goal to work with any button name. So it tries to match most used names. If core wants default mapping, then button names can be corrected in the core. Of course very specific buttons like FDS,MIC won't be mapped by default.

sorgelig:
Unfortunately this PR broke many things in input, so current source code is broken. Wait for my fixes!

PR #127, opened 2019-10-23
Show gamepad in OSD when doing button mapping

Newsdee:
Add visual representation of the internal MiSTer gamepad when mapping from the menu core.
(it is not shown when mapping from regular cores)

This will help users map their controllers more intuitively.

OSDgamepad

### DIFF
--- a/menu.cpp
+++ b/menu.cpp
@@ -2200,7 +2200,44 @@ void HandleUI(void)
 
 			OsdWrite(3, s, 0, 0);
 			OsdWrite(4);
-			if (!line_info) OsdWrite(12);
+			
+			if (!line_info) {
+				if(is_menu_core() && joy_bcount && get_map_button() < DPAD_BUTTON_NAMES)
+				{
+					p = joy_bnames[get_map_button() - DPAD_NAMES];
+					// draw an on-screen gamepad to help with central button mapping
+					OsdWrite(10, "  \x86 L1\x88               \x86 R1\x88  ");
+					if (!strcmp(p, "L"))	OsdWriteDelay(10, "  \x86 \x1b \x88               \x86 R1\x88  ");
+					if (!strcmp(p, "R"))	OsdWriteDelay(10, "  \x86 L1\x88               \x86 \x1b \x88  ");
+					OsdWrite(11, " \x86\x81\x81\x81\x81\x81\x81\x81\x81\x81\x81\x81\x81\x81\x81\x81\x81\x81\x81\x81\x81\x81\x81\x81\x81\x81\x88");
+					OsdWrite(12, " \x83   U                 X   \x83");
+					if (!strcmp(p, "UP"))			OsdWriteDelay(12, " \x83   \x1b                 X   \x83");
+					if (!strcmp(p, "X (Backspace)"))OsdWriteDelay(12, " \x83   U                 \x1b   \x83");
+					OsdWrite(13, " \x83 L \x1b R  Sel Start  Y   A \x83");
+					if (!strcmp(p, "A (OK/Enter)"))	OsdWriteDelay(13, " \x83 L \x1b R  Sel Start  Y   \x1b \x83");
+					if (!strcmp(p, "Y"))			OsdWriteDelay(13, " \x83 L \x1b R  Sel Start  \x1b   A \x83");
+					if (!strcmp(p, "LEFT"))			OsdWriteDelay(13, " \x83 \x1b \x1b R  Sel Start  \x1b   A \x83");
+					if (!strcmp(p, "RIGHT"))		OsdWriteDelay(13, " \x83 L \x1b \x1b  Sel Start  \x1b   A \x83");
+					if (!strcmp(p, "Select"))		OsdWriteDelay(13, " \x83 L \x1b R   \x1b  Start  Y   A \x83");
+					if (!strcmp(p, "Start"))		OsdWriteDelay(13, " \x83 L \x1b R  Sel   \x1b    Y   A \x83");
+					OsdWrite(14, " \x83   D   \x86\x81\x81\x81\x81\x81\x81\x81\x81\x81\x88   B   \x83");
+					if (!strcmp(p, "DOWN"))			OsdWrite(14, " \x83   \x1b   \x86\x81\x81\x81\x81\x81\x81\x81\x81\x81\x88   B   \x83");
+					if (!strcmp(p, "B (ESC/Back)"))	OsdWrite(14, " \x83   D   \x86\x81\x81\x81\x81\x81\x81\x81\x81\x81\x88   \x1b   \x83");
+					OsdWrite(15, " \x8b\x81\x81\x81\x81\x81\x81\x81\x8a         \x8b\x81\x81\x81\x81\x81\x81\x81\x8a");
+				} else {
+					if(is_menu_core()){
+						//clear all gamepad gfx
+						OsdWrite(10);
+						OsdWrite(11);
+						OsdWrite(12);
+						OsdWrite(13);
+						OsdWrite(14);
+						OsdWrite(15);
+					} else {
+						OsdWrite(12);
+					}
+				}
+			}
 
 			if (get_map_vid() || get_map_pid())
 			{
--- a/osd.cpp
+++ b/osd.cpp
@@ -205,6 +205,15 @@ void OsdWrite(unsigned char n, const char *s, unsigned char invert, unsigned cha
 	OsdWriteOffset(n, s, invert, stipple, 0, 0, usebg, maxinv);
 }
 
+void OsdWriteDelay(unsigned char n, const char *s, unsigned char invert, unsigned char stipple, char usebg, int maxinv)
+{
+	if (s && s[0] && CheckTimer(scroll_timer)) // scroll if long name and timer delay elapsed
+	{
+		OsdWrite(n, s, invert, stipple, usebg, maxinv);
+		scroll_timer = GetTimer(SCROLL_DELAY2); // reset scroll timer to repeat delay
+	}
+}
+
 // write a null-terminated string <s> to the OSD buffer starting at line <n>
 void OsdWriteOffset(unsigned char n, const char *s, unsigned char invert, unsigned char stipple, char offset, char leftchar, char usebg, int maxinv)
 {
--- a/osd.h
+++ b/osd.h
@@ -24,6 +24,7 @@ void OsdSetTitle(const char *s, int arrow = 0);	// arrow > 0 = display right arr
 void OsdSetArrow(int arrow);
 void OsdWrite(unsigned char n, const char *s="", unsigned char inver=0, unsigned char stipple=0, char usebg = 0, int maxinv = 32);
 void OsdWriteOffset(unsigned char n, const char *s, unsigned char inver, unsigned char stipple, char offset, char leftchar, char usebg = 0, int maxinv = 32); // Used for scrolling "Exit" text downwards...
+void OsdWriteDelay(unsigned char n, const char *s="", unsigned char inver=0, unsigned char stipple=0, char usebg = 0, int maxinv = 32); //same as OsdWrite but with a certain delay. Used for "blinking" text lines
 void OsdClear(void);
 void OsdEnable(unsigned char mode);
 void InfoEnable(int x, int y, int width, int height);


sorgelig:
i don't see the need of such "picture" because of:
1) Virtual layout is quite simple, so anyone understand it.
2) There is no hard mapping of buttons. For example A/B X/Y can be swapped on specific gamepad which will confuse the user. Some gamepads have quite different layout which from other side doesn't break the functionality. User may have his own vision where each button should be located on his gamepad.

Newsdee:
It helps understand the mapping process, especially when the user physical
gamepad is not "snes" shaped.

I added this after user feedback from various people that found the "magic"
confusing (on Discord). This was seen aa a solution that was easy to
understand without docs.

Kitrinx:
The letters are not important and are in fact misleading. A on xbox is different than A on snes is different than A on genesis. What is important is that the user understands the physical location of where the button is being mapped to, and that why the picture is important. The idea is to create mappings based on the physical layout of a pad, consistently. The actual labels are arbitrary.

sorgelig:
I still don't get it. When you enter this dialog you already have the gamepad in hands and see these labels on it. And just follow the letters you see on your gamepad, not this picture which may not represent your actual gamepad. I even don't say half of buttons/sticks are not shown on this pic.

klaritan:
I think the point is to help the users see what the original controllers looked like and bind them like so as close as possible.
But if we're doing that, we'd need to have ascii gamepad for all cores with very different layouts (e.g. SNES vs. Genesis). I like the direction it's going, but if we only have SNES-style layout, it could actually be more confusing.

Kitrinx:
The gamepad in your hands has some random labels. For example, south-most button: Xbox will be A, Switch pad will be B, Playstation will be X. How do you ask the user to press the south most button? If you give them a picture that shows where "A" is, and then tell them to press "A" they can understand what you want them to press. If you just say "A" and they have an xbox controller, they will press the wrong button, and then NES will have buttons mapped backwards. Do you understand?

sorgelig:
OSD asks to press the button labeled "A" not the button located on south! So depending on your gamepad or your wish you can press either button labeled "A" or button you wish to be used as "A".
That's why picture is misleading. It implies physical location which is wrong by concept.

sorgelig:
> Playstation will be X

But X on Playstation is not a letter but sign. It's still A or B depending on region. But anyway it just confirms the picture is wrong by many means.

Newsdee:
Mapping physical location is part of the intended concept. It makes it much more intuitive than trying to figure out where "your physical A" should be for it to work nicely by default for most cores.

Remember that the purpose overall is to reduce config files and overrides. Using this picture makes it more obvious for users to get it right.

The picture is "right" in that it represents the MiSTer internals as an abstraction. What user inputs by mapping is saying "these are the buttons I want for face buttons", "these are the buttons for shoulder buttons", etc.

Newsdee:
Also please note that we flash the active button to make it more obvious where it is. So even if the user ignores labels, they can quickly map their gamepads just by following visual cues.

James-F2:
> OSD asks to press the button labeled "A" not the button located on south! So depending on your gamepad or your wish you can press either button labeled (on your physical gamepad) "A" or button you wish to be used as "A".
> That's why picture is misleading. It implies physical location which is wrong by concept.

This contradicts the exact reason of having a virtual snes pad layout.
Physical position is the important part here, not the names of the buttons.

sorgelig:
> This contradicts the exact reason of having a virtual snes pad layout.
> Physical position is the important part here, not the names of the buttons.

no, it doesn't contradicts. Virtual gamepad is for defining set of common buttons. Where physically these buttons will be is up to the user. If someone is get used to have A button on south, then he can assign it there and get the Main Fire/Confirm on the south in all cores. If someone wants to get it on east, then he can define it there. It doesn't break the concept. It makes it more universal and adoptable.

James-F2:
> Virtual gamepad is for defining set of common buttons. Where physically these buttons will be is up > to the user.

There is no point to having virtual gamepad then,, since the core goes by physical position.
The idea is to have the correct console layout in relation to the virtual gamepad.

Newsdee:
The user is still free to map to their preference.

But the purpose of having an internal MiSTer gamepad is also to imply some positioning. If we have that, then defaulting can be made more accurate by default, which is the whole point for me. 

It makes documenting further mapping predictable. It is easier to say and use "this button is East like SNES A" than "I use A for East and hope it is, possibly, in SNES layout".

sorgelig:
> There is no point to having virtual gamepad then,, since the core goes by physical position.
> The idea is to have the correct console layout in relation to the virtual gamepad.

it's up to you. You can treat it as no point. So this concept is not for you then. And the point is to have the list (not the really fixed to places) of buttons, so you won't have to define them in every core.
Different consoles may have different layout. For example genesis has ABC XYZ layout which cannot be represented by SNES layout. But having enough buttons in virtual gamepad even Genesis will get all of them defined. Physically some systems may have different position for button A. If for example some system had A on the north, it doesn't mean it will be automatically mapped to X on virtual gamepad. It will be still on A depending which button you choose. For those who insist to have A on north in particularly this core still have option to define the buttons in the core. 99% users won't care of original gamepad layout and will be happy to have A on button they defined as system-wide.

James-F2:
> 99% users won't care of original gamepad layout and will be happy to have A on button they defined as system-wide.

Exactly the opposite.

Kitrinx:
That's not true at all. When you play snes or nes games, the layout of the buttons matters tremendously. The point here is to abstract the mapping into a standard layout so that the correct *physical* layout can be inferred from that mapping. That's why it's done by position, not by letter or anything else. Strictly physical position. Then you can start TG16, and I and II will be the right way, or gameboy, and A and B will the be the right way, etc.

Newsdee:
Genesis 3 button and 6 button can be mapped intuitively from SNES layout. 

As examples: SNES had 6 face button gamepads which had L and R on the right, and M30 has a SNES equivalency where it gives Z as L and C as R.


Sure you can't map to both 3 button and 6 button once, but it remains an intuitive default that can be easily explained.

sorgelig:
No, it's not how this feature will work.
I know there are 2 opposite concepts. MiSTer will use button names to make default map. This is even technically achievable than physical map.
You can re-define the buttons if you don't like the default map.

Newsdee:
I am confused as to why do you consider the physical layout as not achievable? That is the point of having a standard.

sorgelig:
> That's why it's done by position, not by letter or anything else.

Please read the code. You will see the mapping is done by button names, not by physical positions. An this is the only logical way to map. You can't know where button A physically located on original controller. And it even will be more confusing when you ask for A button in Menu core, and then in some specific core you will find the A button is in other place. Game asks you to press "A". You press A but nothing happens.. And only after angry complain in forums and FB you get reply "oh, this system has button A in place of button X". WTF?
Trying to map it physically regardless button names is contrintuitive and confusing.

sorgelig:
> As examples: SNES had 6 face button gamepads which had L and R on the right, and M30 has a SNES equivalency where it gives Z as L and C as R.

sure, and ABXY from Genesis are mapped exactly to ABXY of virtual gamepad. And it's different from original Genesis as A/B and X/Y should be swapped to be matched. This is what i'm talking about. I happy with label matching. If i will find it inconvenient i can always re-map it. But i want to have A button in all cores in the same place, not moving around. Same as Fire 1 i like to be on A and Fire 2 on B. Not like Fire 1 on Y and Fire 2 on A which could be more correct from positional point of view.

Newsdee:
I see your point, but what we are after is different, it's to make A move where it should be per core, and match the common controllers of that system.

We could make the behavior change by user INI settings, so user can specify their own preference?

sorgelig:
Physical position is just suggestion, not requirement. If you use M30/Genesis gamepad, then most likely you will use Z for L and C for R, and most likely you won't like to change ABXY to other positions than they are labeled on M30. Otherwise you will always fight yourself to press A labeled as B and B labeled as A and same for X/Y..

Newsdee:
Also: we can solve the intuitiveness of mapping by adding a controller test screen. Then user knows what is actually happening.

You can send user complaints to me if that is the concern :)

sorgelig:
Please read my messages above. There is no way to know where physically buttons are located on original gamepad.
So, for now MiSTer will use label matching according to labels asked in Menu core.

sorgelig:
Can new way through virtual gamepad provide ready to play environment for newbie users? - YES
Can it suit all possible wishes? - NO
Can it be fixed? -YES, use "Define XXX buttons"  to have layout you like.

The main reason for new gamepad definition is to provide the button definition for all cores as a ready to use solution, so newbies will stop to complain "Ahh, my gamepad doesn't work in XXX core."
It's not a replacement for "Define buttons" option.

Newsdee:
> There is no way to know where physically buttons are located on original gamepad.

Yes there is: this is exactly why we ask users to map to a SNES layout . They declare what (in their opinion/preference) matches that physical layout.

Can labels only be ready to play? No, because it will result in unintuive button placement for many cases as console games expect their original controllers are being used.

The new gamepad definition aims to reduce the need for specific mappings.

sorgelig:
> Can labels only be ready to play? No, because it will result in unintuive button placement for many cases as console games expect their original controllers are being used.

I play games with layout i define and have no problems. So you are wrong.

> Yes there is: this is exactly why we ask users to map to a SNES layout . They declare what (in their opinion/preference) matches that physical layout.

It's only valid for SNES core.

sorgelig:
> The new gamepad definition aims to reduce the need for specific mappings.

Not correct. New definition aims to provide SOME button definitions so any core can be controlled by gamepad.

sorgelig:
1) There is not enough info from core where each button must be placed. Only name is provided, so only by name it can be mapped.
2) If names have no meaning (A can be anywhere depending on core) then in Menu core we cannot name the buttons as A,B,... We have to name neutrally like compass sides or similar.

So technically new way has no any relation to physical map. It's all about labels map. As a developers you should understand from the code what all this is about. I don't understand why you switched off the logic and pretend to be a newbie with wish "I want a magic"

Kitrinx:
You can make menu maps like B,A,,,Select,Start for NES, with spacers to tell it how to line up with the default mapping.

sorgelig:
It will require to modify all cores. Unmodified cores will get strange default layout. Probably it will be implemented sometime later. Currently you solve one problem but produce others difficulties.
Imagine button definition on OSD where you need to define B and then A, Then may be start, then X.. Awkward, isn't?
Putting so much effort to feature which is basically diminished by "Define buttons" option is irrational. Redundant complexity makes things worse.

Newsdee:
I understand we have a difference of opinion here. I assumed it was a bit clearer.

If I undersrand you correctly, you are coming from the perspective of keeping button names the same regardless of where they are physically. So if e.g. the screen says "Press A" for Neogeo then user looks down and finds A, and done. If a controller has a different name (e g. playstation) they decide which label matches A (say Triangle). 

I would say this kind of mapping is best for a user with no direct experience of retro systems. It sacrifices physical accuracy for intuitiveness in finding buttons.

There is another kind of user, which has more retro experience . They are aware of different gamepads configurations, and just wants an authentic experience without worrying too much about setup. This is especially relevant if you have experienced playing the original system with original controllers.

Would you accept making this optional? 

We can cater for both atyles depending on user preference. We need a good option for the ini e.g. :

gamepad_layout=0  - map by label
gamepad_layout=1 - map to physical SNES layout
(we can have more for variations in behavior)

We can always have option 0 by default for a core, so your original intent remains in place, until a core author decides to allow aupport for mode 1. But then it is only used if the end user declares thenselves to want that behavior.

Hopefully that will avoid surprises so even FB users that aren't very technical can use it nicely.

If we have that option we greatly eliminate the need to map per core. I expect most intermediate users would want a "natural physical layout" for a core.

Newsdee:
I should add that I am fine relabeling face buttons "North, South, etc" instead of SNES names if that makes it more intuitive. We can do both e.g. "A (East)".

sorgelig:
MiSTer provides an option to define the buttons per core. No technical knowledge is required for that. Experienced retro users definitely can understand how to press menu button and enter the button definition. Current SNES layout already fits most cores to their physical buttons (if it possible). So we are basically talking about couple cores where buttons may be shuffled from their original physical position. So it's just a matter of re-defining the gamepad in these couple cores. It's much easier than mess with ini settings and make input subsystem more complex.
Problem is greatly exaggerated than what it really is.

Newsdee:
Adding more buttons as we just did is a huge step, yes. But when I talked about SNES layout I always had the physical placement in mind as well.

It isn't just about remapping a couple of cores, but each core per controller. As a developer it seems completely inefficient to have to remap when I know a few options can drive all/most cores nicely from one mapping.

I am undestanding that your concerns are about users being surprised and complaining, and also about adding complexity to the framework.

But this code can live encapsulated (it already is to some extent) so it should not affect other parts of the stack, unless I'm missing something? Can we cater for that in a way that we can preserce this feature?

I can accept that you tell me advanced mapping may confuse completely new (non retro) users. But there is also a large community of retro users as well that would want that kind of behavior.

To me, relying on a physical layout gives image/visual feedback which is understood by muscle memory rather than the brain (reading). 

I think between us we represent different point of views of various users. So hopefully we can provide a wide range of options.

I can commit to document this all very nicely and make it as obvious as possible with screenshots, videos, etc. If that helps?

sorgelig:
While preferences for each people can be different, there is a basic common sense.. If you will tell to user that button A will be on B and B will be on X, X will be on Z and Z will be on A, then he will tell "What??". And this is per core shuffling. Sure some one would like it, but it's NOT common sense. While different system had different layouts, user will have to operate by a single gamepad in most cases. It will be a pain to remember where is specific button on each system while sticking to labels will make it common.
As for specific layout, then it's more common to have different gamepads for different cores. So you may have SN30Pro for SNES and some other cores with similar layout, and M30 for Genesis and other cores. Imagine you have M30 gamepad and to match a genesis core layout you will need to press A instead of B and B instead of A while defining it in Menu core. same for X/Y swap.. While labeling matching you define ABXY as they are asked in Menu core and Genesis core will match the physical M30 layout.

sorgelig:
@Newsdee basically all your talk about "what if i'm lazy to define the keys for core". And you are talking about this like you cannot define the layout.
My intention is to provide a working gamepad for all cores. May be not Pro-oriented, but enough for new users. No more.
Advanced users HAVE to use settings. This is how things are working.

Newsdee:
Please note, we do not need a rewrite for positional config. I can drive it from the current joy_map function with some extra flags, assuming user followed mapping using the SNES layout presented in UI.

Cores could optionally request new mapping with a secondary joy string that is used only for defsult malling eg:

Core labels (for override mapping):
A,B,C,Start,Mode,X,Y,Z

Default map labels (to match MiSTer internals):
B,A,L,Start,Select,Y,X,C

The existing code works same as today, but accepts an additional joy_balias that a core could define.

I think that would be minimally invasive from current code and get what we are after at low risk.

sorgelig:
i will think about it

Newsdee:
I understand your intention.

What I am asking is to extend it slightly, so as to cater for advanced users who have a predictable want (match physical layout). 

It is not even really advanced users, but owners of retro hardware vs. people who only used PCs.

If your concern is the work needed fo refactor the code, I think we can avoid that with how I suggested above.

sorgelig:
ok. i will define additional option for cores used for physical mapping in case some INI option is enabled.
I have a headache already from this conversation..
Actually sometimes is better just simply follow original plans without looking to the sides.. When hobby doesn't bring good feeling, then it will risk to be abandoned.

Newsdee:
Sorg, I am in your side and trying to work with you here, offering to shoulder user feedback and avoid, as much as I can, this being more work for you. 

I take your concerns seriously and want to cater for them. I am sorry if my trying to understand things fully and present my perspective came across in the wrong way. As you know, I've been around for a while and am finally starting to contribute more to what I can. 

PM me (forums or FB) or let me know if you prefer that we discuss things in any other way. I'd be also glad to chat about other ideas in return. My doors are open.

bootsector:
I think adding this drawing of the Virtual Pad helps a lot for the joypad first mapping.

I'd like to add my two cents for the overall solution though: 

1 - Put a drawing of the MiSTer Virtual Pad in the Wiki with buttons on it
2 - First time mapping MiSTer should ask buttons as per buttons from Virtual Pad drawing
3 - Cores will have defaults based on Virtual Pad buttons to the system's real buttons. This won't make everyone happy, so see below!
4 - Users should be able to override core's defaults by remapping those (just the same as today!)

sorgelig:
btw, i don't see any flashing while defining the keys in Menu core

sorgelig:
ok. i see where is the problem

PR #136, opened 2019-11-07
Save input map files to /config/inputs/

Newsdee:
(old folder is still compatible, will revert if file not found)


sorgelig:
Generally i'm fine with subfolder, but please remove "skip/remove" from menu.cpp. Undefine is more correct term i think. It's not a skip and not remove.

Newsdee:
Done, I reverted it back.

PR #143, opened 2019-12-05
allow loading of WAVE tracks

Kitrinx:


sorgelig:
this is half solution. wav has header which needs to be skipped.

PR #175, opened 2020-02-28
option to hide joystick mappings, games path change

Kitrinx:
- Adds an options to hide controller popup for people who capture video from mister
- Changes the root dir search order to after usb/cifs, (just before /games)


sorgelig:
I've implemented it slightly differently. Folder priority is also changed.

PR #181, opened 2020-03-16
allow volume menu to work without ini option enabled

Kitrinx:
I can't see any reason to disable the volume menu by default. Please correct me if there is a reason I don't see. Keys remain disabled by default, and enabled by INI option.

### DIFF
--- a/menu.cpp
+++ b/menu.cpp
@@ -2327,7 +2327,7 @@ void HandleUI(void)
 			memset(bar, 0x7f, 8 - m);
 		}
 
-		OsdWrite(13, s, menusub == 0, !cfg.volumectl);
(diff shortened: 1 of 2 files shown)


sorgelig:
Unfortunately, there are many strange users who somehow lower the volume till zero and then start to complain "arghhhh.. i have no sound!"
This is why control is disabled by default.

Kitrinx:
They would have to do so willingly from the menu, with the keys being disabled from default. If a user cannot operate a simple volume slider, is this really our issue? Many more users can't figure out how to enable volume slider, which is arguably much more confusing since the option doesn't specify that it enables this.

sorgelig:
yeah, OSD option has been added later, so now should be safe to enable it permanently

PR #188, opened 2020-04-07
Disabled merging for Mayflash 2 port SNES controller and Xinmo 2 player usb encoder

mlc4050:


sorgelig:
Mayflash has many different adapters. Are you sure all of them shouldn't be merged?

mlc4050:
A good point.  I supposed it should look at PID as well.

PR #203, opened 2020-04-22
add support for JammaSD input device

vikbez:
JammaSD is a device that is commonly used in arcade machines to convert input buttons to keyboard keys (for using a computer inside a cabinet)

I coded something that works so far: emulating 2 joysticks depending on what input the mister receive from the JammaSD;

I don't know if it's the right way to do it and the right place to put the test, but it works so far.

There are more devices working like that (Jpac for example); I took a long time to think how this should be implemented and made this, but I could be wrong.

### DIFF
--- a/input.cpp
+++ b/input.cpp
@@ -816,6 +816,51 @@ static int ev2archie[] =
 	NONE  //255 ???
 };
 
+// jammasd have shifted keys: when 1P start is kept pressed, it acts as a shift key,
+// outputting other key signals. Example: 1P start + 2P start = KEY_ESC
+static char jammasd2joy[35][3] =
+{ // key,      player, button_remap
+    {KEY_5,         1, 0}, // 1P coin
+    {KEY_1,         1, 0}, // 1P start (shift key)
+    {KEY_UP,        1, 0}, // 1P up
+    {KEY_DOWN,      1, 0}, // 1P down
+    {KEY_LEFT,      1, 0}, // 1P left
+    {KEY_RIGHT,     1, 0}, // 1P right
+    {KEY_LEFTCTRL,  1, 0}, // 1P 1
+    {KEY_LEFTALT,   1, 0}, // 1P 2
+    {KEY_SPACE,     1, 0}, // 1P 3
+    {KEY_LEFTSHIFT, 1, 0}, // 1P 4
+    {KEY_Z,         1, 0}, // 1P 5
+    {KEY_X,         1, 0}, // 1P 6
+    {KEY_C,         1, 0}, // 1P 7
+    {KEY_V,         1, 0}, // 1P 8
+    {KEY_9,         1, 0}, // Test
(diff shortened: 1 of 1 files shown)


sorgelig:
with such patch you won't be able to define keys per core

vikbez:
What about something like that then ?

sorgelig:
I plan to publish info how i've adapted the arcade cabinet's inputs for MiSTer using Arduino Micro.
Basically you don't need any extra components. Just buttons (and spinner) connected to arduino board. And it works fine without any changes in code.
Then may be you can see if you want to go that route. 
Modern arcade cabinets input board are full of weird things.

vikbez:
Well I'm not doing this for me (I used a teensy in my cabinet).
I  wanted to help people around me that all have a jammasd and want to play with a mister to have a "easy" way of doing it.
Does a PR with this will be accepted if it seems correct to you or do you not prefer having this in the main_mister code ?
Maybe you have a better idea for the implementation ?

mcarquigny:
For members of the communities I belong to JammaSD is almost the default configuration in most cabinets.
I don't know if the solution proposed by vik "as it is" is sufficient or if it needs more work, but having this compatibility would be a real plus (at least for me and I guess for a lot of people) !

sorgelig:
It's unclear if all jammasd controls use the same vid/pid, otherwise there is no way to distinguish it from normal keyboard.

sorgelig:
I have idea how to add it. But i have no jammaSD to test.

what mean these records?
```
{KEY_BACKSPACE, 1, 0},    // 1P up shifted
    {KEY_P,         1, 0},    // 1P down shifted
    {KEY_ENTER,     1, 0},    // 1P left shifted
    {KEY_TAB,       1, 0},    // 1P right shifted
    // {KEY_5,         1, 4}, // 1P 1 shifted (defined above)
    // {KEY_6,         1, 5}, // 1P 2 shifted (defined above)
    {KEY_ESC,       2, KEY_1} // 2P start shifted
```
are these keys are physically separate? Wat meas "1P shifted"?

sorgelig:
oh, i've saw your comment in the code.

vikbez:
These records are for knowing if the input comes from player 1 or 2; (the 3rd column is useless for P1, I have to improve the code if we go that route)

You can find the jammasd outputs here (page 2) https://www.webasd.com/public/doc/ASD275D/JammASDusb_Pinout.pdf

Ok you saw the comments. For now when shifting I did the code to behave like the input is the same key.

sorgelig:
I've pushed the change to support JammaSD.
you need to add into MiSTer.ini:
```
jammasd_vid=0x04D8
jammasd_pid=0xF3AD
```
then MiSTer should see it as 2 joysticks which you have to map in Menu core initially.

vikbez:
I tested and it works. Closing this ! Thanks

PR #213, opened 2020-05-03
Comments: converted mixed codepage mess to UTF-8

sonic74:


sorgelig:
It may get corrupted again and not all editors are UTF-8 aware or properly handle mixed text. Basically it needs either to be translated to english or all non-english comments removed.

sonic74:
> It may get corrupted again 

The point of UTF-8 is that it never gets corrupted again. 

> and not all editors are UTF-8 aware 

Even console editors like vim and emacs are already UTF-8 aware. 

> or properly handle mixed text. 

The text was mixed codepage 866 and 1251 before, now it's pure UTF-8.

> Basically it needs either to be translated to english or all non-english comments removed.

Comments in a foreign language are much better than no comments at all. Even without speaking russian, I can make out the meaning of some of them.

PR #219, opened 2020-05-11
Minor timing adjustments

GromRGB:
Hi,

Clould you see that rotation_ms = sec_per_revolution * 13.33 ? The array in seektime.cpp can be simplified.
The 13.33 can be replaced too by 1000.0/75.0 to show that the CDRom read 75 sectors in 1s.

After function get_cd_seek_ms could be replaced by  (int) get_cd_seek_latency to remove (float) outside the function. (float) is the devil :)

Finally, I think (2 * 1000 / 60) is an integer operation. 
I think it would be better to add .0 after all float numbers.
Example : milliseconds = (36.0 * 1000.0 / 60.0) + sector_list[target_index].rotation_ms * 0.5 + (track_difference - 644.0) * 16.66 / 195.0

Regards

### DIFF
--- a/support/pcecd/pcecd.cpp
+++ b/support/pcecd/pcecd.cpp
@@ -24,8 +24,14 @@ void pcecd_poll()
 
 	if (CheckTimer(poll_timer))
 	{
-		poll_timer += 13 + (!adj ? 1 : 0);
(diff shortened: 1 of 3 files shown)


GromRGB:
Sorry, I just want to make a comment for dshadoff

PR #234, opened 2020-05-24
Reduce CPU load on ARM side. MiSTer runs cooler.

samuel-giroux:
Before, the CPU load of MiSTer was 48%, now it's 2% as reported on top command.
I tested the temperature with a heatsink, no fan, running the NeoGeo core with Metal Slug X running in the background.
After 5 minutes, MiSTer_20200519 was 5.11 degrees Celsius warmer than before ambient temperature.
Whereas after 5 minutes, this commit was only 2.78 degrees Celsius warmer than before ambient temperature.

### DIFF
--- a/file_io.cpp
+++ b/file_io.cpp
@@ -1068,12 +1068,12 @@ struct DirentComp
 	bool operator()(const direntext_t& de1, const direntext_t& de2)
 	{
 
-#ifdef USE_SCHEDULER
-		if (++iterations % YieldIterations == 0)
-		{
-			scheduler_yield();
-		}
-#endif
+//#ifdef USE_SCHEDULER
+//		if (++iterations % YieldIterations == 0)
+//		{
+//			scheduler_yield();
+//		}
+//#endif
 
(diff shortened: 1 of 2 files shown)


sorgelig:
ARM in Cyclone V has no power management circuit, therefore it doesn't matter if you call sleep or not - CPU always work. Sleep in this context means empty loop. So instead of doing empty loop, let the CPU do some useful task.
5.11 or 2.78 with "loading drop" from 48% to 2%..  You don't need better proof of useless tweak.
CPU loading with absent PM block just shows how much CPU is wasted. 2% means 98% of CPU performance is wasted for empty loop.  What it means? It means MiSTer will be more laggy for input, laggy for disk operations from FPGA core, and other problems.

sorgelig:
Btw, for cooled system, it doesn't matter how hot it becomes after N minutes. It's important temperature change direction. If it increasing, then cooler system failed. So you need changes of 0 or minus. Only in this case you can tell, the problem is solved. 
Otherwise it doesn't matter if system will reach the critical temperature after 5 minutes or 10 minutes.

PR #244, opened 2020-06-07
Adding initial J-PAC/I-PAC support

bdlou:
This change uses the work that was done for JAMMASD support to allow for J-PAC and I-PAC to be supported in a similar fashion.  I've mapped all of the keys defined in the Default Key code table for the J-PAC and I-PAC (except for tilde, as I was not clear looking through the code what the key code for this is).  This was tested using a pre-2015 J-PAC with a mini-USB connected to an unmoddified standard JAMMA harness with buttons P4 - P6 wired from a CPS2 kick harness to the J-PAC.  Once configured, both joysticks, all 6 buttons on both players, and the shift keys are all enabled.

### DIFF
--- a/MiSTer.ini
+++ b/MiSTer.ini
@@ -106,6 +106,13 @@ controller_info=6
 jammasd_vid=0x04D8
 jammasd_pid=0xF3AD
 
+; J-PAC keys to joysticks translation
+; Provide the VID and PID for your J-PAC/I-PAC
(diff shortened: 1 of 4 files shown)


sorgelig:
actually i don't see any noticeable difference from jammasd. Just some additional service keys which can be simply added to jammasd.
I will merge jpac/ipac to jammasd so it will be a single option.

Issue #242, opened 2020-06-07
Bluetooth controller D-pad stops working after Bluetooth keyboard goes to sleep


c0d3h4x0r:
I use a DIERYA DK61 Bluetooth keyboard and an 8-bitdo M30 Bluetooth gamepad with a Bluetooth USB dongle with my MiSTer.

Whenever I have both devices on and connected and working, and then I start playing a game (using any core) using the gamepad, the moment the keyboard goes to sleep due to inactivity, the D-pad on my controller stops responding.  The buttons on it still work, however.  To fix it, I have to wake the keyboard up so it reconnects.  The only way I have found to avoid this is to manually power down the keyboard before launching a core — but this is not a viable workaround when using computer cores (such as Minimig) to play a game.

Is there some way the Bluetooth handling can be fixed so that when my keyboard goes to sleep, the game controller will keep working?

sorgelig:
This is specific to your BT devices. Never had such problem. May be keyboard sends some codes confusing the dongle or system when going to sleep. 
I only can suggest to use other BT keyboard.

sorgelig:
I close this issue as it's definitely not related to Main.
Any linux related issues should be discussed on forum. In most cases they are not related to Main (such as this one).

c0d3h4x0r:
Definitely not a problem with my Bluetooth keyboard, as this problem does not happen when using thr same keyboard, controller, and Bluetooth dongle with my HTPC or my Raspberry Pi.  It only happens with MiSTer, so it’s definitely a MiSTer problem of some kind.

sorgelig:
You don't understand what is open source hobby project. It's not like "i have problem - you have to fix." I have no devices you use, so most likely i won't fix it, as i have no such issue.
Tested keyboards and gamepads are discussed on forum. If you want to provide fixes for other devices - you are welcome.

c0d3h4x0r:
I understand perfectly well that this is an open source hobby project.  I never said you have to fix it.  I just said it’s definitely a MiSTer problem.  No need to take it personally.

c0d3h4x0r:
[...]

PR #249, opened 2020-06-21
Fix issue #242

c0d3h4x0r:
Fix issue #242 by preserving all still-connected input devices in their existing array positions whenever a device goes to sleep or gets unplugged.

### DIFF
--- a/input.cpp
+++ b/input.cpp
@@ -1,3 +1,5 @@
+#include <set>
+#include <string>
 #include <stdio.h>
 #include <fcntl.h>
 #include <unistd.h>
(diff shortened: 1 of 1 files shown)


sorgelig:
Commit is large and i don't get the idea.
If the whole point is to try to keep the devices on their positions after re-enumerating, then it's useless.
Your fix it seems just masking the issue, not solving.

c0d3h4x0r:
> If the whole point is to try to keep the devices on their positions after re-enumerating, then it's useless.

No, it is **not** "useless".  The positions of devices changing is the root cause of the problem, and preserving their positions is the correct solution.

> Your fix it seems just masking the issue, not solving.

Incorrect.

However, I am closing this PR because I came up with a different approach (I created a different PR for it) that is less churn and preserves **only** the positions.

PR #250, opened 2020-06-22
Fix for Issue #242 (second approach)

c0d3h4x0r:
When flushing/recreating devices, preserve array positions of devices that are still connected.


c0d3h4x0r:
Per other conversion, I have a third fix approach that's even better, so closing this PR.

PR #251, opened 2020-06-22
Fix for Issue #242 (third approach)

c0d3h4x0r:
Fix for Issue #242 -- preserve core/virtual input device numbers across Linux input device flush/recreation.

### DIFF
--- a/input.cpp
+++ b/input.cpp
@@ -1689,23 +1689,30 @@ static void input_cb(struct input_event *ev, struct input_absinfo *absinfo, int 
 
 	if (!input[dev].num && ((ev->type == EV_KEY && ev->code >= 256) || (input[dev].quirk == QUIRK_PDSP && ev->type == EV_REL)))
 	{
-		for (uint8_t num = 1; num < NUMDEV + 1; num++)
+		bool num_used[NUMDEV + 1] = { false };
+		for (int i = 0; i < NUMDEV; i++)
 		{
-			int found = 0;
-			for (int i = 0; i < NUMDEV; i++)
+			// paddles/spinners overlay on top of other gamepad
+			if ((input[dev].quirk == QUIRK_PDSP) && (input[i].quirk == QUIRK_PDSP))
 			{
-				// paddles/spinners overlay on top of other gamepad
-				if (!((input[dev].quirk == QUIRK_PDSP) ^ (input[i].quirk == QUIRK_PDSP)))
-				{
-					found = (input[i].num == num);
-					if (found) break;
-				}
-			}
-
-			if (!found)
-			{
-				input[dev].num = num;
+				input[dev].num = input[i].num;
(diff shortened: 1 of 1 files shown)


input.cpp
```diff
@@ -2752,6 +2766,10 @@ int input_test(int getchar)
 							input[n].vid = id.vendor;
 							input[n].pid = id.product;
 
+							auto result = devnums.find(device_key_string(id.vendor, id.product));
+							if (result != devnums.end())
+								input[n].num = result->second;
```
c0d3h4x0r, line 2771: With `num` being set here, it's now possible for the loop in `input_cb()` to assign an already-used `num` value to an additional device, per this assignment at line 1707:

```
			if (!found)
			{
				input[dev].num = num;
				break;
			}
```
So in addition to the changes I've made already, that loop will need to avoid assigning `num` values that are already in use by other input devices.

c0d3h4x0r: Actually, the longer I look at that preexisting nested `for` loop, the more I realize that what it's doing doesn't make much sense.  I think it's already assigning `num` values incorrectly, even without my change.

c0d3h4x0r: I rewrote the loop to do the right thing, and I've committed that addition to this pending PR.

That fixed the last issue I was seeing.  (Sometimes when I reconnected one device out of my three [keyboard, gamepad1, gamepad2], it wouldn't work as expected (because it was assigned the wrong `num`), or its input would stutter/lag (because another unrelated input device was already using the same `num`).

input.cpp
```diff
@@ -2717,12 +2732,18 @@ int input_test(int getchar)
 	if (state == 1)
 	{
 		printf("Open up to %d input devices.\n", NUMDEV);
+
+		// Remember which devices mapped to which core device numbers.
+		std::map<std::string, unsigned int> devnums;
```
c0d3h4x0r, line 2737: Actually, there's one more problem -- device numbers aren't globally/permanently remembered during the lifetime of the MiSTer process, so as different devices get disconnected/reconnected, the mapping of real device to virtual device changes around... and sometimes the keyboard ends up mapped to a `num` that was previously used for gamepad1, so then neither of the gamepads work.  I think the right thing to do is to make this a global/static, and then move all the logic into that same loop I just rewrote, so all the logic for assigning input device `num`s is all in one place.

sorgelig:
fixed in today's release.

input.cpp
c0d3h4x0r: I've committed another change to this pending PR which implements that idea.  Now, no matter what order I disconnect/reconnect devices, the same device always maps to the same virtual device, as the user would naturally expect.

sorgelig: It's already fixed in release. If assigned device is removed then it's removed from player assignment forever.

Issue #258, opened 2020-06-29
Disconnected USB or Bluetooth input devices do not map to virtual devices correctly after reconnecting


c0d3h4x0r:
(This is a follow-on to #242 )

**Repro Steps**
This is only one confirmed set of repro steps; others are likely also possible.

1. Turn on MiSTer w/ USB Bluetooth dongle attached.
2. Turn on previously-paired 8bitdo M30 Bluetooth gamepad, and it reconnects to MiSTer.
3. Turn on previously-paired Bluetooth keyboard, and it reconnects to MiSTer.
4. Use the gamepad to launch the Minimig core from the OSD.
5. Start playing an Amiga game.
6. Pause the game (or just walk away from it) for a while, long enough for both the keyboard and gamepad to fall asleep.
   * This is to simulate a common real-life interruption.
7. Come back to MiSTer.
8. Wake up gamepad, it reconnects.
9. Wake up keyboard, it reconnects.
10. Try to resume playing the game.

**Expected Result**
Keyboard and gamepad should still be recognized by the running Minimig core and should both still be able to open and operate the MiSTer OSD.  Mappings between real devices and virtual devices should remain unchanged.

**Actual Result**
Neither the keyboard or the gamepad are recognized anymore by the running Minimig core.  The keyboard can no longer open or operate the MiSTer OSD.  The gamepad can open and operate the OSD, but not the running core.  By toggling the "joystick swap" option in the core's chipset settings in the OSD, the gamepad is suddenly recognized again by the running Minimig core.  It's as if the mapping between real devices and virtual devices has been changed.

c0d3h4x0r:
The fix for #242 can be seen here:
https://github.com/MiSTer-devel/Main_MiSTer/commit/03d824a6b34b4eb8e6a8749d61b24b185c879364

It looks like that fix specifically addressed disconnection/reconnection of gamepads, paddles/spinners, and lightguns, but not keyboards.  Also, it uses the `input[i].devname` property for matching during the "restore players" loop, which is unsafe.  That property (with values such as `/dev/input/event1`) does not remain constant for a given physical input device across disconnect/reconnect, so it cannot safely be used for unique physical device identification and matching.  Only the Vendor ID + Product ID combination can be used to uniquely identify a physical device across disconnect/reconnect.

[...]

PR #259, opened 2020-06-29
fix for issue #258

c0d3h4x0r:
use vid+pid for matching unique physical device across disconnect/reconnect

### DIFF
--- a/input.cpp
+++ b/input.cpp
@@ -2952,7 +2952,9 @@ int input_test(int getchar)
 				devInput *player = (input[i].quirk == QUIRK_PDSP) ? player_pdsp : player_pad;
 				for (int k = 0; k < NUMPLAYERS; k++)
 				{
-					if (strlen(player[k].devname) && !strcmp(player[k].devname, input[i].devname))
+					if (strlen(player[k].devname) &&
+						player[k].vid == input[i].vid &&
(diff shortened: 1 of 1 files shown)


sorgelig:
This is not correct way. It will break the logic. If you have several gamepads of the same model, then they will have the same vid+pid. The only possible way to avoid that issue is don't let your gamepad to go to sleep, i.e. if you play the game, then play. Once gamepad mapped to player is disconnected, it's un-mapped. Actually it's not fatal. The case you described in issue is about playing alone. So, once you return back to game, use the same gamepad (don't touch keyboard or other sleeping devices) and it will be automatically assigned to the same (first) player.

sorgelig:
i close this PR as the method of restoring the map is unacceptable. It completely breaks the case when you have several gamepads with the same vid/pid and all then will be mapped to the same player.

c0d3h4x0r:
The scenario (one person playing alone, using a Bluetooth keyboard and one Blueyooth gamepad) is badly broken.  This is a real problem that needs to be fixed.  Other people in the Classic Gaming Discord have been complaining about it too.  Please do not close those; instead, let’s work together to figure out a way to fix it.  This issue makes MiSTer basically unusable for many people.

sorgelig:
There is a discussion in your issue. So i've posted there.

PR #262, opened 2020-07-01
Implemented type-to-filter functionality in long game lists

OskarSigvardsson:
Hello! My name is Oskar, and I'm a programmer from Sweden. I think this project is incredible, and I'd like to contribute in any way I can. I don't have any experience with hardware definition languages, so I can't exactly help out with the cores, but I'm an experienced C++ programmer and has worked some with UI/UX design, so I figured I can maybe contribute some small stuff to the menu program at least. 

One thing I noticed early when firing up my MiSTer is that the keyboard doesn't behave as expected when you're in a long list of games. If you want to play "Crash Bandicoot" for the GBA, you naturally start typing "crash..." on your keyboard. This would have the very unexpected effect of first scrolling to the letter "C", then to the "R", then to "A", etc. This is not expected behaviour and is kind of annoying. Many of these lists can get quite lengthy, and I think there's a real need to be able to filter them a little bit quicker. 

This PR implements a "type-to-filter" functionality, whereby when you start typing, it will filter the list to the prefix of the word you've typed. So if you type "crash", it will filter to games that begin with "crash" (if you type just a single letter, it behaves the way it's always done, jumping to that section of the list). You can reset the filter by pressing backspace, or by waiting a little bit (currently 2000 ms) and type some other letter, which will "reset" the filter. Here's a gif that shows the idea: 

menu

The functionality is implemented using the "prefix" argument to ScanDirectory. I'm considering implementing a slightly more advanced filter, where you can type "su ma" and it shows only the "Super Mario" games, but I thought I'd start here to gauge if this was something you feel would be a good contribution to the project. 

I've tried to make the change as small and surgical as possible, and I've tested to make sure that it's stable and bug free. It currently doesn't work on file lists which contains directories, because that would require a more extensive change to ScanDirectory. I'm perfectly willing to do that, but even without it, I still think this is a good feature (it's more meant to be used in the longer file lists). 

I hope you find this a good change, I certainly think it improves the experience of using the menus in a small way. I'm very happy to make whatever changes you think are needed in order for the pull request to be merged.

### DIFF
--- a/menu.cpp
+++ b/menu.cpp
@@ -3992,6 +3992,10 @@ void HandleUI(void)
 
 		if (flist_nDirEntries())
 		{
+			static char prefix[256];
+			static unsigned long prefix_typing_timer = 0;
+			int prefix_len = strlen(prefix);
+
 			ScrollLongName(); // scrolls file name if longer than display line
 
 			if (c == KEY_HOME)
@@ -4019,6 +4023,17 @@ void HandleUI(void)
 				menustate = MENU_FILE_SELECT1;
 			}
 
+			if (c == KEY_BACKSPACE && prefix_len > 0)
+			{
+				memset(prefix, 0, 256);
+				prefix_typing_timer = 0;
+				
+				printf("Prefix is: %s\n", prefix);
+				ScanDirectory(selPath, SCANF_INIT, fs_pFileExt, fs_Options);
+
+				menustate = MENU_FILE_SELECT1;
+			}
+
 			if (down) // scroll down one entry
 			{
 				ScanDirectory(selPath, SCANF_NEXT, fs_pFileExt, fs_Options);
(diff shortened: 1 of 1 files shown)


sorgelig:
Hi!
I like your patch. It looks tidy. So what happens if there are sub-directories?

OskarSigvardsson:
Then it basically doesn't do anything different than before: I'm relying on the preexisting `prefix` parameter to ScanDirectory, and it seems like that's totally ignored for directory entries. I'm thinking that's not a huge deal, since it's pretty rare you have massive lists of directories, but you can easily have lists of thousands of individual games. 

As I said, I'd be happy to go and do that fix so that directories work with `prefix` as well, but that's a slightly larger project, and I personally think this PR works without it.

sorgelig:
No problem. I will play with your patch and see if it needs to be extended to dirs.
Thanks!

OskarSigvardsson:
Awesome! Glad to have contributed!

sorgelig:
I think i will extend it to filter by any position of typing word (probably only when more than 1 symbol typed). Sometimes i remember only approximate name of game. Sometimes list may include "the" or "a" in the beginning.

PR #263, opened 2020-07-02
Use proper filter instead of prefix to filter file lists

OskarSigvardsson:
Update on the merged pull request from yesterday: I added a `filter` parameter to ScanDirectory that I use instead of the prefix that was already there. This filter parameter filters on any part of the word (per @sorgelig's suggestion), so if you type `ADV` it will include both `Advance Wars` and `Ace Combat Advance`. It also works on both files and directories (it's neat to just type "snes" at the console core selection screen). I was a little bit worried that the performance wouldn't be acceptible, but it seems to work pretty in the folders I've tried. 

I opened a new PR since the PR from yesterday was closed and merged, I hope that's the right way to do it.

### DIFF
--- a/file_io.cpp
+++ b/file_io.cpp
@@ -1254,7 +1254,7 @@ static void get_display_name(direntext_t *dext, const char *ext, int options)
 	if (fext) *fext = 0;
 }
 
-int ScanDirectory(char* path, int mode, const char *extension, int options, const char *prefix)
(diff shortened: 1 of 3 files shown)


sorgelig:
single letter still just jumps to first item of that letter without filtering?

OskarSigvardsson:
Yes, it only uses filter for 2 or more letters.

PR #265, opened 2020-07-03
Fixing new bug on reboot functionality

theypsilon:
Rolling back first change of 29e9150ca7864a63aff2a6cc53534645ac19db34

It caused a bug on reboot functionality

After reboot you wouldn't see files in the OSD, and you would have to do Cold reboot or power cycle your MiSTer.

This commit fixed it.

### DIFF
--- a/file_io.cpp
+++ b/file_io.cpp
@@ -91,9 +91,9 @@ static int FileIsZipped(char* path, char** zip_path, char** file_path)
 static char* make_fullpath(const char *path, int mode = 0)
 {
 	const char *root = getRootDir();
- 	if (strncasecmp(getRootDir(), path, strlen(root)))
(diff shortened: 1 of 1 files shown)


theypsilon:
I just realized there is something more subtle with this change. Now after that fix, scripts can't be launched. So please don't merge this PR at the moment.

theypsilon:
So I basically reverted that whole commit https://github.com/MiSTer-devel/Main_MiSTer/commit/29e9150ca7864a63aff2a6cc53534645ac19db34 and I confirm it works for the reboot bug and the scripts problem I had in my previous fix. But I'm guessing I broke something else because investigating, I found this message: https://misterfpga.org/viewtopic.php?f=4&t=653&start=25

Then, this is not a proper solution, and I'm closing this PR.

PR #287, opened 2020-09-15
fix compilation warning (misleading-indentation)

frederic-mahe:
```
support/arcade/mra_loader.cpp: In function ‘int xml_send_rom(XMLEvent, const XMLNode*, SXML_CHAR*, int, SAX_Data*)’:
support/arcade/mra_loader.cpp(651, 3): warning: this ‘if’ clause does not guard... [-Wmisleading-indentation]
   if (result<0)
   ^~
support/arcade/mra_loader.cpp(653, 4): note: ...this statement, but the latter is misleadingly indented as if it is guarded by the ‘if’
    if (result==-1)
    ^~
```

### DIFF
--- a/support/arcade/mra_loader.cpp
+++ b/support/arcade/mra_loader.cpp
@@ -648,14 +648,14 @@ static int xml_send_rom(XMLEvent evt, const XMLNode* node, SXML_CHAR* text, cons
 		 */
 		{
 			int result = buffer_append(arc_info->data, text);
-  			if (result<0)
+			if (result<0)
 			{
-                		printf("buffer_append failed %d\n",result);
-                		if (result==-1)
-                  			printf("-1 no data given\n");
-                		if (result==-2)
-                  			printf("-2 could not allocate\n");
-              		}
+				printf("buffer_append failed %d\n",result);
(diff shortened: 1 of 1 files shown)


support/arcade/mra_loader.cpp
```diff
@@ -649,11 +649,13 @@ static int xml_send_rom(XMLEvent evt, const XMLNode* node, SXML_CHAR* text, cons
 		{
 		int result = buffer_append(arc_info->data, text);
 		if (result<0)
+			{
```
markun, line 652: Perhaps you should also fix the indentation (the 2 lines above and everything inside the new scope you created). Right now it is not consistent with the indentation in the rest of the file.

frederic-mahe: the ugly mix of tabs and spaces should be fixed now.

frederic-mahe:
sorry, I've used the GitHub online editor to make the changes, hence the mix of tabs and spaces. Indentation should be fixed now, and it should match the indentation style of the rest of the file.

sorgelig:
meaningless

PR #288, opened 2020-09-15
fix compilation warning (misleading-indentation)

frederic-mahe:
```
menu.cpp: In function ‘void HandleUI()’:
menu.cpp(1174, 5): warning: this ‘if’ clause does not guard... [-Wmisleading-indentation]
     if (user_io_core_type() == CORE_TYPE_SHARPMZ)
     ^~
menu.cpp(1183, 2): note: ...this statement, but the latter is misleadingly indented as if it is guarded by the ‘if’
  switch (menustate)
  ^~~~~~
```

### DIFF
--- a/menu.cpp
+++ b/menu.cpp
@@ -1167,35 +1167,35 @@ void HandleUI(void)
 		}
 	}
 
-    // SHARPMZ Series Menu - This has been located within the sharpmz.cpp code base in order to keep updates to common code
-    // base to a minimum and shrink its size. The UI is called with the basic state data and it handles everything internally,
-    // only updating values in this function as necessary.
-    //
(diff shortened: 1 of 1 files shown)


menu.cpp
```diff
@@ -1180,8 +1180,8 @@ void HandleUI(void)
 			       menu, select, up, down,
 			       left, right, plus, minus);
 
-	switch (menustate)
-	{
+    switch (menustate)
+    {
```
markun, line 1184: replacing tabs with spaces does not seem to be the right fix here since almost all of the code is indented with tabs

frederic-mahe: you are right, sorry about that.

sorgelig:
meaningless

PR #289, opened 2020-09-16
update GCC version

frederic-mahe:
The toolchain seems to use gcc 6.3 now.


sorgelig:
can i decide which GCC version to use?

PR #290, opened 2020-09-16
fix check which was always returning false

markun:


markun:
looking at the code above, this check is redundant and it doesn't matter that it was broken.

sorgelig:
that's is really bug, but currently i have no time to check that. 
It definitely should be fixed.

PR #298, opened 2020-10-25
snes: add special LoRom mapping (Derby Stallion 96 and Sound Novel Tsukuru)

srg320:


paulb-nl:
has_bsx_slot is always true?

PR #301, opened 2020-10-30
Adjusted autofire rate to allow 8hz increments and display the rate a…

ConHuevosGuey:
Adjusted the autofire increment rate to 8hz intervals and the displayed rate to the rate of fire rather than the rate that it is toggled.

### DIFF
--- a/input.cpp
+++ b/input.cpp
@@ -37,7 +37,7 @@ int  joy_bcount = 0;
 
 static int ev2amiga[] =
 {
-	NONE, //0   KEY_RESERVED
+	NONE, //0   KEY_RESERVED 
 	0x45, //1   KEY_ESC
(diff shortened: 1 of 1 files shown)


sorgelig:
Old display was correct unlike new one. The rate in milliseconds should include a whole period (press+release). So 16ms initial is the half-period of autofire. The whole period is 32ms (not hz!). I will correct it.

sorgelig:
after accepting the PR i was thinking more. 32ms as highest rate is too fast! NTSC consoles used 30Hz for video which is 32ms per frame. So with 32ms rate it means either pressed or released in every frame. So it will be either always pressed or always released from console view.

paulb-nl:
I don't know where you get 30Hz from. The consoles use 60Hz video.

PR #305, opened 2020-11-23
improvements

Pasha-From-Russia:
When you play the game, you cannot see what time is it (unless you restart the MiSTer)
Added time string to "Information" box (shows only when core is running)

did some code alignment fixes (spaces -> tabs) too


sorgelig:
The place you've used for time is actually occupied.
Anyway i don't see a strong need for time. You can check the time on watches or phone or whatever surrounding devices.

Pasha-From-Russia:
> The place you've used for time is actually occupied.
> Anyway i don't see a strong need for time. You can check the time on watches or phone or whatever surrounding devices.

occupied by what? Battery info fits the box

Pasha-From-Russia:
> The place you've used for time is actually occupied.
> Anyway i don't see a strong need for time. You can check the time on watches or phone or whatever surrounding devices.

of course I can check the time anywhere else, but what about usability?

sorgelig:
I don't see great usability when you have to open OSD and then go to another page to check the time.

Pasha-From-Russia:
> I don't see great usability when you have to open OSD and then go to another page to check the time.

sounds reasonable. Ok, I'll rework that.
And you didn't tell me what occupies the place I used for the time

Pasha-From-Russia:
> I don't see great usability when you have to open OSD and then go to another page to check the time.

well, am I right that menu.rbf does all the framebuffer drawing stuff and draws time header (lines 16,17, 18) by itself so in order to make the "%Core%" page look like main page (with a header that shows the time) it is required to update all the cores?... :(

Pasha-From-Russia:
@sorgelig yet still, I deleted the modification commit, but you can merge the rest 2

sorgelig:
Menu core is basically empty and has a lot of free resources, that's why it's not a problem to consume more BRAM from OSD. For normal cores every bit should be saved and framework should consume as less resources as possible. That's why header is not available in normal cores. And it's done on purpose to save the resources.
No need to try to output time to OSD in every core. It's useless feature wasting the resources and/or OSD place.

Pasha-From-Russia:
i'll stick to the patch made in my fork then, ASA nobody else needs that feature... Wasn't going to alter the cores...

PR #315, opened 2020-12-27
Support CHD CD images

zakk4223:
Add support for reading CHD cd images to ao486, TGFX16-CD and MEGACD cores.

Dependencies added due to this: libchdr, flac, lzma. (libchdr also requires zlib, but existing miniz is used instead)

Timing/performance: CHD stores data in hunks, each hunk contains 8 sectors. Retrieving audio data takes around 1.3-1.5ms per hunk. So approximately 1.5ms of access time every 106ms of audio. I've encountered no audio playback issues in any of my testing.

If there are any cd images that are known to be problematic, please let me know and I will test them.

You can use the mame 'chdman' utility to convert both bin/cue and iso images to chd.


zakk4223:
One additional note: the change to menu.cpp can be reverted once the individual cores are updated to include CHD in the extension string for cd image mounting.

sorgelig:
lemme double check: did you remove bin/cue support?

zakk4223:
No, did not remove bin/cue. Made sure that still worked when I was done

KremlingKuthroat19:
I couldn't agree more! CHD is the easiest way to manage and compress CD games. It saves gobs of data (around 40-50%) and neatly compresses each iso/bin per CD into a single file. It's the best compression method I'm aware of and if it's good enough for the MAME team then it's good enough for me.

flynnsbit:
@zakk4223 Did you by chance also test using the imgset command in AO486?  Does it still work and CHDs mounted?

Kitrinx:
Sega CD in particular is extremely sensitive to read timing. Zipped cd's often crash on this system due to slow reads, so this is probably the best one to test for speed stability of CHD.

sorgelig:
good, i will merge it after some my local changes.

zakk4223:
> @zakk4223 Did you by chance also test using the imgset command in AO486? Does it still work and CHDs mounted?

Yes, imgset works to mount chds. I think the only issue you'll have with the exodos/tdl stuff is there's no good way for you know if the user converted a cd image to CHD, so it may complicate your scripting on the DOS side.

flynnsbit:
> > @zakk4223 Did you by chance also test using the imgset command in AO486? Does it still work and CHDs mounted?
> 
> Yes, imgset works to mount chds. I think the only issue you'll have with the exodos/tdl stuff is there's no good way for you know if the user converted a cd image to CHD, so it may complicate your scripting on the DOS side.

The scripts can be easily changed to reference a CHD instead and then pushed with the github updates. so at least the pack won't have to be rebuilt.  If imgset could give some error control back if a file doesn't mount, then I could script around that.

  We can back port the cd converter in the ExoDOSConverter so that the images can be created at the source at the very least.

sorgelig:
cannot build. I don't see how it's adopted to miniz. It still requires zlib sources

sorgelig:
ok, change zlib.h reference to miniz.h allowed it to be compiled.
Converted Sonic USA to chd and got decompression error on MiSTer.

zakk4223:
I see what it is. I'm using the misterkun/toolchain docker image (from https://github.com/MiSTer-devel/Main_MiSTer/wiki/Native-ARMv7-Toolchain-on-x86-64) to build. It contains zlib.h in /usr/include. I'd neglected to change the include in libchdr so it was using 'native' zlib instead of miniz. Mister already links to -lz as part of the IMLIB2_LIB requirements, so it all linked correctly and worked. Seems the version of miniz in the Mister source fails to work for some reason.

I did a bit more experimenting and it seems like the newest version of miniz works. 

So either do whatever is required for #include "zlib.h" to work or update miniz. Seems to me that if -lz is already being linked in it might just work out better to use that (unless you think you will remove it one day)

sorgelig:
Please fix the problem as you an author of patch and know how it works. Make sure it doesn't use something outside of included sources.

zakk4223:
I have two possible solutions I can commit: Since you are the maintainer I'd like to know which you'd prefer.

1. Update miniz to the latest version
2. Add the minimum amount of defines/function prototypes to libchdr_chd.c so that the already linked libz.so (from lib/imlib2) is usable.

Updating miniz feels cleaner to me, but I don't know if you have reasons to use the version that is currently included in MiSTer. I did make sure that after upgrading miniz that zip file navigation still works, so it doesn't seem to break anything. (Is there anywhere else miniz is used?)

sorgelig:
there is no reason to keep current version of miniz. So feel free to update it.

PR #316, opened 2020-12-29
Miniz update

zakk4223:
Update miniz to latest (2.1.0) release.
Rever libchdr to use inflateReset


zakk4223:
I also switched my build environment to use the linaro release on the wiki instead of the docker image so there shouldn't be any surprise external dependencies this time

sorgelig:
Where is inflateReset() defined? libz? So you want to tell you pass foreigner structure to libz while using miniz? This is definitely wrong

zakk4223:
It's part of miniz. Please see miniz.h

sorgelig:
oh, i see. they added this function

PR #328, opened 2021-01-12
Added a 2048x240p @120hz video mode for 240p support on PC CRT

mathieulh:
This is used to display 240p (including SD CRT like native scanlines) on VGA and higher PC CRT monitors. This has been tested successfully on multiple PC CRT monitors.


sorgelig:
Numerical hard coded resolutions better to keep for standard resolutions. 
This is quite specific resolution. It's not a problem to use it as a custom resolution. 
```
video_mode=2048,40,442,200,240,3,10,8,85909
```
Anyway, no one will understand what is this strange resolution for without long explanation. That's what Wiki for. You can create a good article there and provide a custom resolution string. So anyone can copy paste and use it.

PR #330, opened 2021-01-13
Added a true 320x240 @120hz mode (for PC CRT) as well as a 1600x1200 mode (works on compatible LCD/OLED panels too)

mathieulh:
Please disregard this Pull Request if you don't feel like users benefit from these video modes being added.


sorgelig:
From all modes only 1600x1200 is good to have in numerical modes.
I'm not against to use non-standard modes. This is why i've added custom video mode functionality. 
Nothing prevent you to use them. Its quite niche use by some users who understand the matter. You can make a Wiki page to explain the feature and add those modes lines there to use.

piso77:
Completely OT: where did you get the numbers/timings for 240p mode?
Or did you derive it? And how? Trying to learn something new here... :)

http://www.tinyvga.com/vga-timing <= this only list 640x480 and above.

mathieulh:
@piso77 you can use this site to get most of the timings https://tomverbeure.github.io/video_timings_calculator
MiSTer is compatible with CVT and CVT-RB (in fact the default modes use the later), CVT is more compatible with panels/screens than CVT-RB is, but CVT-RB requires lower pixel clocks.

This site calculates pixel clocks slightly too low so you either have to calculate that value by hand or use either CRU or the nvidia control panel custom resolution utility to get the proper pixel clock value.

You can also do without this website entirely and use CRU or use the Nvidia custom resolution panel and calculate the vertical and horizontal back porches by hand (total pixels - (sync + front porch + act)).

PR #332, opened 2021-01-14
Use KPENTER as select on osd menu.

bellwood420:
I bought a compact USB keyboard shown below for my MiSTer.
Its ENTER key seems to emit code which is treated as KEY_KPENTER by Linux when NumLock LED is on.
So that means the ENTER key cannot be used as select on osd menu when toggled to Joystick 1 emulation mode. 

picture

### DIFF
--- a/menu.cpp
+++ b/menu.cpp
@@ -1153,6 +1153,7 @@ void HandleUI(void)
 			break;
 		case KEY_ENTER:
 		case KEY_SPACE:
+		case KEY_KPENTER:
 			select = true;
 			break;
 		case KEY_UP:


sorgelig:
btw, you can use ScrLck key to switch joystick emulation as well.
I will merge your PR later.

bellwood420:
Thanks for info, but I already tried ScrLck.
Even if ScrLck is used to switch joystick emulation, my keyboard emits KEY_KPENTER when NumLock LED is on.
NumLock LED may be a condition with my keyboard.

I tested and this fix worked well at least with my keyboard.
If there are no problems, I would be happy for the fix to be merged.

sorgelig:
Of course you need to switch NumLock off.
Your keyboard emulates a key pad when Numlock is ON.
I've suggested ScrLck to keep NumLock state off.

bellwood420:
I'm sorry but there seems to be misunderstanding.
Sorry for my poor explanation.
Please let me be clear. 

NumLock state in my keyboard seems to be determined by its LED state.
Just pressing NumLock key doesn't seem to affect NumLock state.

The following are the results in my environment.

- NumLock state turns on when toggling to Joystick 1 emulation mode(where NumLock LED is on) by pressing ScrLck.
- NumLock state turns off when toggling to Joystick 2 emulation mode(where NumLock LED is off) by pressing ScrLck.
- NumLock state turns on when toggling to Joystick 1 emulation mode(where NumLock LED is on) by pressing NumLock.
- NumLock state turns off when toggling to Joystick 2 emulation mode(where NumLock LED is off) by pressing NumLock.
- NumLock state is always off in MENU core where NumLock LED is always off even if pressing NumLock.
 
NumLock, ScrLck LEDs are controlled by Main_MiSTer to show joystick/mouse emulation state.

I cannot keep NumLock off by using ScrLck, but the fix helps.

sorgelig:
i see. As i've said i will merge it later. I need to finish my modifications first.

bellwood420:
Thank you for understanding.
I'm in no hurry, just wanted to be clear.

PR #342, opened 2021-02-03
Support for reassigning player number for joypads and spinner/paddles

zakk4223:
Replace the 'Reset player assignments' with 'Change player assignments' menu.
In this menu you can change what 'player slot' any controller is assigned to by left/right or OSD 'select'. Full reset of all player assignments is also available in the menu, with the exact same functionality as the previous 'Reset player assignments'

When an input for a pad or spinner is detected the corresponding menu entry is prepended with an asterisk to indicate which one it is (to help users reassign the correct gamepad)

Primary motivation for doing this is that a number of arcade games have fairly important gameplay differences when the Player 2 side is used. If you only have one gamepad there's was no way to easily play as P2; with this you can just remap to P2. As a nice side benefit you can now pick your character in three player Gauntlet games :)

### DIFF
--- a/input.cpp
+++ b/input.cpp
@@ -1082,11 +1082,79 @@ static int mapping_type;
 static int mapping_count;
 static int mapping_clear;
 static int mapping_set;
+static int last_input_dev = 1<<31;
+static int last_pdsp_dev = 1<<31;
+
+static int player_num_remapping = 0;
+static int remapping_spinner_input = 0;
 
 static uint32_t tmp_axis[4];
 static int tmp_axis_n = 0;
 
 static int grabbed = 1;
+
+int get_dev_num(int dev)
+{
+	return input[dev].num;
+}
+
+
+int get_pad_mask()
+{
+	int ret = 0;
+	for (int i = 0; i < NUMDEV; i++)
+	{
+		if (input[i].num && input[i].quirk != QUIRK_PDSP && input[i].quirk != QUIRK_MSSP)
+		{
+
+			ret |= 1<<i;
+		}
+	}
+
+	return ret;
+}
+
+int get_pdsp_mask()
+{
+	int ret = 0;
+	for (int i = 0; i < NUMDEV; i++)
+	{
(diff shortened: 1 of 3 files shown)


zakk4223:
The last active input detection in input.cpp (triggered by entering the remap menu) works for all the input devices I have. One thing I am missing is a Mr Spinner in paddle only mode, so if someone could test that I would appreciate it.

If there's a better way to do the input detection I would be interested in any improvement ideas.

sorgelig:
99% arcades are non-multiplay. So most arcade cores written that way when you can play for both players by single gamepad when one user pass the gamepad to another.
For me it looks over complicated, not a MiSTer way. Also large changes in input.cpp usually leads to a lot of bugs as logic there is very complicated. Even i forget sometimes how it works :)

zakk4223:
So the only way to deal with playing as non player 1 in games that have p1/p2 differences is to plug in another controller or bother the core developer to add a player mapping option?

sorgelig:
I didn't get your question... Which p1/p2 differences you mean? I didn't see such arcades

zakk4223:
DoDonPachi has scoring differences between player sides, as does a number of other games that the Cave core will support in the future.  Four player gauntlet assigns the characters based on the player slot, so if three people are playing no one can pick Elf. Beat'em Ups like The Punisher have different characters (with different move sets) depending on if you play on the P1 or P2 side.

It's not too common in the early to mid 80's games, but it starts to show up more in late 80s and 90s. Especially in shmups and beat'em ups. In some cases it is purely cosmetic, and in some cases it has game play changes. I'm not sure the extent of how many games do this is really documented any where. 

Feels like the Mister is getting closer and closer to having more 90s arcade games and this may come up more. I'd rather have a consistent interface to changing player assignments than to have individual cores all do it somewhat differently...

sorgelig:
So couple games you mentioned can include special setting in the core.
For example some consoles and computers have option "Swap joysticks". No problem to add such function in arcade.

PR #382, opened 2021-04-15
progress_info and missing controller_info check

funkycochise:
conditional controller_info to display Info(") in user_io.cpp
new progress_info settings in Mister.ini to show/hide rom loading progress bar for mras and neogeo core
default shows progress bar if progress_info doesn't exist in Mister.ini

example in Mister.ini to display progress_info use

Compiled Mister Core


sorgelig:
I don't see a reason why progress should be hidden as nothing else is happening.
Controler info message you disabled has nothing to do with controller info. These messages appear when you press the button to switch keyboard mode. This is important message to know the keyboard behavior.
Muting it is inappropriate.

PR #383, opened 2021-04-15
O_CLOEXEC flags to fix file descriptor leaks

### LINKED ISSUE #379: Freeze of the MiSTer app after changing core a lot of times
I got an [issue report for mbc](https://github.com/pocomane/MiSTer_Batch_Control/issues/7) about a freeze after a lot of rom changes. After a bit of investigation, I found it is probably a bug of the MiSTer app. I can reproduce it with the following command via ssh:

for (( ; ; )); do echo "load_core /media/fat/_Console/SNES_20210222.rbf" > /dev/MiSTer_cmd; date +%H:%M:%S; sleep 7; done;

After about 5 minutes, the MiSTer freezes and it does not accept anymore commands via keyboard or joypad. To recover, you have to relaunch the MiSTer app, i.e. in a ssh:

killall MiSTer
/media/fat/MiSTer &
disown

zakk4223:
Resolves #379

### DIFF
--- a/DiskImage.cpp
+++ b/DiskImage.cpp
@@ -1156,7 +1156,7 @@ void TDiskImage::Open(const char *filename, bool ROnly)
 	}
 	FType = typ;
 
-	int hfile = open(filename, O_RDONLY);
+	int hfile = open(filename, O_RDONLY | O_CLOEXEC);
 
 	if (hfile < 0)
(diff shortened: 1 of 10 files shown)


sorgelig:
Thanks for testing.
I will rework the memory part.

PR #385, opened 2021-04-20
MiSTer.ini: fix typo in default filters comment

tjanas:
Also update example vfilter_default to reference an existing file from Filters_MiSTer repo

### DIFF
--- a/MiSTer.ini
+++ b/MiSTer.ini
@@ -145,6 +145,6 @@ shared_folder=
 ;spinner_pid=0x0005
 ;spinner_throttle=-50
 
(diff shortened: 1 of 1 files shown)


rsn8887:
This doesn't work for me. Because of the space in the folder name, I have to put single quotes around the filter path and filename.

tjanas:
I am using the following in my MiSTer.ini, and it is correctly applying it for all the cores where I did not specify a video filter (note that this filename has spaces but I am not putting quotes around the filename):
`vfilter_default=Interpolation (Sharp).txt`

PR #397, opened 2021-05-14
Optimized arcade rom deinterleaver.

blackwine:
This patch optimizes loading of `<interleave>`'d sections of MRA file.
Some of the biggest CPS2 romsets load about 4 times faster than previously. Measured from start to end of load including unzipping. Not counting sending time.


blackwine:
There may be an issue in a corner case. Closing it for now, will reopen after ensuring everything is perfect.

PR #416, opened 2021-06-24
Writes path info of menu selection

kaicherry:
Writes to current path and "full path" of whatever is highlighted in the menu for external scripts to monitor.

This was implemented specifically to support a dynamic marquee solution for MiSTer in arcade cabinets, but in a way that it is not "vendor specific". See https://pixelcade.org/mister/ for an example of a practical application.

### DIFF
--- a/menu.cpp
+++ b/menu.cpp
@@ -4414,6 +4414,15 @@ void HandleUI(void)
 		if (flist_nDirEntries())
 		{
 			ScrollLongName(); // scrolls file name if longer than display line
+			
+			//Write out paths infos for external integration
+			FILE* filePtr = fopen("/tmp/CURRENTPATH","w");
+           		FILE* pathPtr = fopen("/tmp/FULLPATH","w");
+           		fprintf(filePtr, "%s", flist_SelectedItem()->altname);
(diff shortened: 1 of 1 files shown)


sorgelig:
not every user like a spy behavior

kaicherry:
I genuinely do not understand what you mean by that. This was specifically designed to mimic the behavior the software has, eg writing the current core name to /tmp/CORENAME.

This is all this code does. I don't understand the 'spy behavior' connotation. These files are "monitored" the way everything else is...via inotify-type scripts.

alinke:
Hi Alexey, I am the creator of Pixelcade, a line of LED and LCD based marquees for arcade machines http://pixelcade.org. We have had many asks from MiSTer users for Pixelcade marquee support, Pixelcade/MiSTer beta user here ==>https://www.youtube.com/watch?v=hhrPBrxSyRQ. This PR simply adds to the file path the current game that has been selected in addition to the console that is already there so there should be no concerns there. This is the only way we know of to add Pixelcade support to MiSTer so please consider merging this PR and of course let me know if you have any questions. Thanks much, Al

PR #454, opened 2021-09-04
Align AL_GETATTR reply size with misterfs redirector

p2mate:
misterfs expects a 9 byte reply while we were sending 10 bytes.

Signed-off-by: Peter De Schrijver <p2@psychaos.be>

### DIFF
--- a/support/x86/x86_share.cpp
+++ b/support/x86/x86_share.cpp
@@ -875,10 +875,9 @@ static int process_request(void *reqres_buffer)
 		*buf++ = sz >> 16;
 		*buf++ = sz >> 24;
 		*buf++ = (mode & S_IFDIR) ? FAT_DIR : 0;
-		*buf++ = 0;
 
 		res = 0;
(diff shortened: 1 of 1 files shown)


sorgelig:
actually it should be 10. And rather code in misterfs should be corrected to:
```glob_intregs.w.ax = ((unsigned short *)answer)[3];```
but actually it has no difference in work and bigger size doesn't corrupt anything as buffer is fixed.

p2mate:
the bug in misterfs is that it rejects answers with length 10. See https://github.com/MiSTer-devel/ao486_MiSTer/blob/8fd09853cf4f02be55a294f4120b8b128dadd940/sw/MiSTerFS/MiSTerFS.C#L485. So setting reslen to 9 is the important bit of the patch, not writing the last byte is more a consistency thing.

sorgelig:
i see

sorgelig:
ok. accepted fix in misterfs

PR #468, opened 2021-09-25
Support for unoptimized mister builds.

wickerwaka:
Pass DEBUG=1 to make to build and unoptimized version.
Seperate build directories are used for debug and release builds.
Don't strip the debug executable.
Change an `inline` to a `static inline` in coretypes.h to ensure symbol is found in debug builds.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,77 +14,118 @@ else
 	Q := @
 endif
 
-INCLUDE	= -I./
-INCLUDE	+= -I./support/minimig
-INCLUDE += -I./support/chd
-INCLUDE	+= -I./lib/libco
-INCLUDE	+= -I./lib/miniz
-INCLUDE	+= -I./lib/md5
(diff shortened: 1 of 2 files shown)


sorgelig:
I don't see need of separate build version. If you want to debug on MiSTer then you can simply use MiSTer.elf renamed to MiSTer - it contains symbols.
"Unoptimized" build will obviously work differently and won't give the same result as normal version. MiSter is not software emulator, it includes FPGA and it needs realtime communication which is quite different between different versions.

wickerwaka:
Yeah, the reason for the unoptimized build is to allowed better single stepping through. I haven't run into any timing related issues in the unoptimized builds, but I'd be interested in knowing what areas might be more sensitive than others.

The separate build directories aren't necessary, just a convenience so both debug and non-debug versions could live side by side with the added benefit of not cluttering up the main directory with intermediate files.

Happy to just keep this change local.

PR #473, opened 2021-10-09
lastgame bootcore feature (WIP)

wickerwaka:
This still needs a lot more testing by me, I'm submitting this PR in order to get feedback on whether there is interest in having this feature, what additional things it needs (e.g. should it handle mounting disk images also?) and feedback on the changes to user_io and menu.

# Description
This change expands the `bootcore` ini option to support a new mode: "lastgame". The lastgame mode will load the previously loaded core (just like "lastcore") but it will also attempt to load the ROM that was previously being played on that core also.

## user_io and menu.

The work of parsing the confstr for file and mount descriptions and handling the loading/mount of them has been moved from the menu code to user_io.

`user_io_parse_selection` parses a confstr entry and fills a `uio_selection_descriptor` with the necessary information for the file or image. The parsing code in the menu system has been replaced by this and several of the globals and function statics (`opensave`, `load_addr`, `ioctl_index`, `store_name`) have been replaced by a static `sel_desc` instance of a `uio_selection_descriptor` which now contains the same information. The `user_io_load_or_mount` is new functionality, it works in reverse, given a filepath it determines which confstr entry matches it's extension and loads or mounts it appropriately.

The code in the `MENU_GENERIC_FILE_SELECTED` and `MENU_GENERIC_IMAGE_SELECTED` that was responsible for loading/mounting the selected file(s) has been moved to `user_io_load_file` and `user_io_mount_image` respectively. The menu code is still responsible for handling the "UI" specfic tasks like recording recent files.

These changes to the menu implementation this need a lot more testing. For instance, the archie and minimig specific code were setting `ioctl_index` but it seems like this was not actually being used for anything and mounting multiple disk images on both of these cores works without issues.

## bootcore
Rewrote most of this because it needs now needs to handle more state and I didn't want to work around how some of it previous worked. Internal functions are no longer in the header and the bootcore state is handled internally (moved from menu.cpp) and doesn't modify the cfg data anymore. The `lastcore.dat` data that gets written out is not a structure that contains the core path along with the name of the core and the last file that was loaded for the core. When a file is selected and when store_name is false (since those files would be loaded automatically when the core starts anyway) that path is written to the lastcore data via `bootcore_record_file`. On core startup the `bootcore_load_file` function will attempt to load that file via the `user_io_load_or_mount` function. Currently only files (ROMS) are tracked, the lastgame functionality doesn't make any attempt to mount disk images.

## Misc
Changes to support/* and file_io.* are just function signature fixes required for const correctness.

### DIFF
--- a/MiSTer.ini
+++ b/MiSTer.ini
@@ -33,6 +33,8 @@ recents=0
 ; lastexactcore - Autoboot the last loaded exact core (corename_yyyymmdd.rbf autosaved in CONFIG/lastcore.dat) first found on the SD/USB
 ; corename - Autoboot first corename_*.rbf found on the SD/USB
 ; corename_yyyymmdd.rbf - Autoboot first corename_yyyymmdd.rbf found on the SD/USB
+; lastgame - Autoboot the load loaded core (same as lastcore) and load the most recently played game (ROM) also
+; lastexactgame - Same as lastgame, but use the lastexactcore behavior to find the core to use
 ;bootcore=lastcore    ; uncomment to autoboot a core, as the last loaded core.
 
 ; 10-30 timeout before autoboot, comment for autoboot without timeout.
--- a/bootcore.cpp
+++ b/bootcore.cpp
@@ -5,15 +5,97 @@
 #include "file_io.h"
 #include "cfg.h"
 #include "fpga_io.h"
+#include "hardware.h"
+#include "support/arcade/mra_loader.h"
+#include "bootcore.h"
+#include "user_io.h"
 #include <string.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <dirent.h>
 
+enum bootcoreType
+{
+	BOOTCORE_NONE,
+	BOOTCORE_LASTCORE,
+	BOOTCORE_LASTCORE_EXACT,
+	BOOTCORE_LASTGAME,
+	BOOTCORE_LASTGAME_EXACT,
+	BOOTCORE_CORENAME,
+	BOOTCORE_CORENAME_EXACT
+};
 
-extern int arcade_load(const char *xml);
-int16_t btimeout;
-char bootcoretype[64];
+static bool isExact(bootcoreType t)
+{
+	switch( t )
+	{
+		case BOOTCORE_LASTCORE_EXACT:
+		case BOOTCORE_LASTGAME_EXACT:
+		case BOOTCORE_CORENAME_EXACT:
+			return true;
+		default:
+			return false;
+	}
+}
+
+static bool isLastCore(bootcoreType t)
+{
+	switch( t )
+	{
+		case BOOTCORE_LASTCORE_EXACT:
+		case BOOTCORE_LASTGAME_EXACT:
+		case BOOTCORE_LASTGAME:
+		case BOOTCORE_LASTCORE:
+			return true;
+		default:
+			return false;
+	}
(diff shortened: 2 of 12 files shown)


sorgelig:
Speaking of lastcore feature - it appeared much before recent menu and currently it's mostly deprecated.

cadahl:
Cheap possibility: Maybe the current menu core could get a feature to auto-load the last core on boot? (It could perhaps distinguish how it was rebooted to get out of the loop.)

wickerwaka:
Yes, it is a pretty large change for a small feature. I generally just play one game at a time so I like turning on my MiSTer and being right there, as if I had left the cartridge in the console. If bootcore is a deprecated then I can appreciate not wanting to expand it's functionality.

I do think the menu and user_io changes are an improvement in general. The parsing is no longer duplicated in the menu code and in `parse_config`, the state relevant to the file selection is stored in a single place rather than being spread across several local and global statics. It could be used for some other feature, like mra-style files for console games, but I understand not wanted to take a risky code-reorganization change without a compelling feature to go along with it.

sorgelig:
Implemented feature requires further maintaining through further changes of API or code reorganization. Bootcore feature were broken several times without even noticing because it's very specific feature for only few people. Recent feature is more general and more versatile because:
1) it doesn't require to be turned on or off. It's always available and doesn't interfere other use scenarios. Several days i would like to use one core, then i've got tired and want to use other core. So i don't need to go to INI to change the settings. I just don't choose it from list.
2) besides the last core, i have last 16 cores to choose from. So it's again more pleased to current mood.
3) It doesn't breaks normal booting process, so it always works as expected unlike bootcore/lastcore which always breaks either one or other feature because trying to start predefined core ASAP. Something like Bluetooth/WiFi may not work for example. Also other pre-init features can be broken from this rush.

Basically all above can be told about lastgame feature. Recents menu is more generalized and more versatile for both lastcore and lastgame. These features are simply cluttering the code and adding doubling functionality.

wickerwaka:
@sorgelig I think it is worth taking another look at the parts of this change that aren't related to bootcore. When I wrote this I was thinking about potential future changes to rom loading, similar to what you are doing with mgl right now.

Putting the mgl logic directly into the menu code is exactly the kind of thing I was hoping to avoid. I don't think it is a positive direction as far as code maintainability is concerned.

sorgelig:
I'm not sure if i understood you.
In mgl i'm trying to simulate OSD action, so it won't interfere normal workflow.

wickerwaka:
Yeah, not an mgl fan myself and my opinions on these kinds of systems has changed over the last few months. However, you've built it now, so I'll share my opinions! :P

It's not good to be in a situation where you feel that the only way to reliably perform an action is by simulating interactions with the user interface. It means that too much logic and state has become intertwined with the presentation logic for the interface. `ioctl_index` and other pieces of information are calculated when the Load/Mount option is selected and then extension index, cheats and other things are handled based on the button press from the file selection screen. This code inside of the menu system is the single source of truth for how to load roms/images for cores, so if you want to programmatically load a rom your only option is to simulate menu interactions.

What I attempted to do as part of the bootcore change was to get all of the logic out of the menu system and drive it all from the confstr. The menu code still does the stuff that is relevant to user interactions, such as updating the recents list. Something very similar to the `user_io_load_or_mount` function from my PR would be able to do what the mgl code is currently doing.

PR #488, opened 2021-11-24
OSD restrictions

wickerwaka:
Limits the features available in the OSD based on the `restrictions` setting in the ini. Intended to support use cases like kiosks, demo systems, kid/family friendly setups, etc.

The restrictions are a comma separate list of strings that specify which aspects to disable:
* `settings`     - Audio/video filters, UART, save/reset settings, palettes, global key/button mappings, scripts, ini switching, storage switching
* `browsing`     - Browsing outside of a cores default directory when loading ROMs or images
* `options`      - Core-specific option bits
* `cores`        - Switching cores in the System menu and rebooting back to the main menu core
* `cheats`       - Cheats
* `dip_switches` - Arcade DIP Switches
* `volume`       - The core and global volume settings in the Misc. Options menu
* `mapping`      - Per-game/core button/key mapping
* `load`         - Loading roms and mounting images
* `all`          - Everything

When the `options` restriction is enabled option bits are enabled, however option toggles (such as resets) as still enabled. Option bits can be enabled by adding `Z` to the confstr for the bit, toggles can be disabled by adding `z`. For example, to allow the difficulty bits on the 7800 to be toggled even when the `options` restriction is enabled the confstr could be changed to: `"ZOC,Difficulty Right,Right,Left;"`

If the `restrictions_code` ini variable is set then there will also be an "Unlock" option available in the OSD which will disable all restrictions. The code is a pattern of button presses.

The majority of changes are in `menu.cpp`. In order to make it easier to manage the disabling of multiple options I added `MenuWriteEntry` and `OsdWriteEntry` functions. These are similar to `MenuWrite` and `OsdWrite` except they manage the setting of the `menumask` bits, inverted rendering when an item is highlighted and setting a new `enablemask` bitfield which indicates whether an entry is enabled or not. The `enablemask` variable can be checked when handling input to determine whether any action can be taken, so "is this menu item enabled" logic doesn't need to be replicated in both the menu drawing and menu input handling.

The cores with special handling in the mister executable (archie, minimig and ST) have not yet been updated to pay any attention to the restrictions. I'm happy to do the work but I didn't want to do any further UI changes until/if the current stuff is approved.

Two example videos of what arcade and console cores look like with all restrictions enabled.

[Arcade Example](https://drive.google.com/file/d/1fB37sMfGKOriioUfYSC-lFRAasPz75CQ/view?usp=sharing)

[Console Example](https://drive.google.com/file/d/1zZA6T-O8JdZDn67oGXbVgD-hmYqoHyTw/view?usp=sharing)


jotego:
Very interesting. Could you explain with an example what the use case for this is?

thehughhefner:
Interesting indeed. Use cases are kiosks or any situations where you don't want the user to mess up any settings. For example exhibitions or events. Hope this gets added

wickerwaka:
> Very interesting. Could you explain with an example what the use case for this is?

Yeah, kiosks and events are one use case, as @thehughhefner points out. My goal was to make it configurable so it can serve multiple purposes. At a tournament you may want to lock out everything, but still allow button remapping. For a setup in a common area you might want to just limit `settings,mappings` which would prevent anyone making permanent changes to the system, but still have full control to load cores and games and tweak options. If you want to setup a "virtual" Genesis then you could disable core switching. Etc.

sorgelig:
interesting, but i'm not ready to accept it. Pretty much specific use. It makes code more complicated and abide these rules... May be, some day...

wickerwaka:
@sorgelig please let me know if there are any simplifications I can make. The `browsing` feature, for instance, is pretty minor, but ends up touching every `ScanDirectory` call.

Initially I tried removing menu entries and whole pages, but I didn't like that it was hiding information and changing the left/right navigation. `menumask` management was still complex which is why I added the *Entry functions to take of that, from that point it was a lot easier to add the `enablemask` bitfield and keep everything visible.

sorgelig:
I will re-check it

PR #493, opened 2021-12-11
Shadow masks v2

wickerwaka:
Up to two sets of intensity values can now be set in the header along with the width and height

`4,4,106.25,62.5,100,25` specifies a 4x4 mask with two sets of intensities 106.25%/62.5% and 100%/25%

The high bit (bit 4) of the mask LUT values controls which intensity set is used.

Valid "on" intensity values: 100-143.75 in 6.25 increments. Valid "off" intensity values: 0-93.75 in 6.25 increments. Defaults are on=112.5 & off=62.5, if not specified.

Characters can be used instead of numbers to define the masks, making them more readable.
Red  = 4 or r
Green = 2 or g
Blue = 1 or b
Magenta = 5 or m
Black = 0 or k
White = 7 or w
Yellow = 6 or y
Cyan = 3 or c

Add 8 or use uppercase to use alternative intensity levels

```
4,2,1,8
8,4,2,1
```

Equivalent to:
 
```
r,g,b,K
K,r,g,b
```

This code will still work with old shadow mask cores, but they won't have access to any of the new features. https://github.com/wickerwaka/Template_MiSTer/commit/aefb09ed3bd6154052c81379fdcaede09f44c8c5 is required to use the new features.

Addresses https://github.com/MiSTer-devel/ShadowMasks_MiSTer/issues/1 and https://github.com/MiSTer-devel/SNES_MiSTer/issues/304

### DIFF
--- a/video.cpp
+++ b/video.cpp
@@ -10,6 +10,7 @@
 #include <sys/stat.h>
 #include <sys/types.h>
 #include <unistd.h>
+#include <math.h>
 
 #include "hardware.h"
 #include "user_io.h"
@@ -402,7 +403,9 @@ static bool has_shadow_mask = false;
 #define SM_FLAG(v) ( ( 0x0 << 13 ) | ( (v) & 0x7 ) )
 #define SM_VMAX(v) ( ( 0x1 << 13 ) | ( (v) & 0xf ) )
 #define SM_HMAX(v) ( ( 0x2 << 13 ) | ( (v) & 0xf ) )
-#define SM_LUT(o,v) ( ( 0x3 << 13 ) | ( ( (o) & 0x3f ) << 4 ) | ( (v) & 0x7 ) )
+#define SM_LUT(o,v) ( ( 0x3 << 13 ) | ( ( (o) & 0x3f ) << 4 ) | ( (v) & 0xf ) )
+#define SM_ON_INTENSITY(o,v) ( ( 0x4 << 13 ) | ( ( (o) & 0x1 ) << 4 ) | ( (v) & 0xf ) )
+#define SM_OFF_INTENSITY(o,v) ( ( 0x5 << 13 ) | ( ( (o) & 0x1 ) << 4 ) | ( (v) & 0xf ) )
 
 enum
 {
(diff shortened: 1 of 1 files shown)


sorgelig:
I want to get rid of this 2 stage intensity and have whole range per lut. I'm working on it.
There won't be much more resources but code and masks will be more simple

sorgelig:
i plan following format:
CXY, CXY, CXY.... 
C - color 0-7
X - intensity for "on" 0-F(100%-183%)
Y - intensity for "off" 0-F(0%-83%)

sorgelig:
Since one BRAM block is spent on lut array there is no problem to extend vmax/hmax to 16 for larger masks. Will it be useful?

wickerwaka:
> Since one BRAM block is spent on lut array there is no problem to extend vmax and hmax to 16 for larger masks without increasing resources usage. Will it be useful?

Yeah, people have asked for larger masks. I was treating block ram as if it was allocated per bit so I've been trying to be conservative, but if the entire resource is going to be used regardless then it makes sense to use it all!

tonurics:
Hello @sorgelig,

Your suggestions will definitely take things to the next level. And will give me a nice UI challenge for the Shadow Mask Editor.

I'm concerned about the "off" intensity. Values closer to 100% are going to be much more for both "on" and "off"; as 100% leaves the underlining channels unchanged, >100% brighter, <100% darker. Changes to "off" [darkness] also cause greater perceptual differences with RGB [you end up changing 2 of the color channels].

While 83% it is a great improvement over the 62.5% of the original release and a happy upgrade. It only gives us about half the brightness of the original release; whereas the betas were able to get us one-fourth brightness at 93.75%.

If it's possible to apply the math of "off" so that it's effectively a 17%-100% range, that would be perfect. The majority of our feedback from the original release was the masks were too dark [at the fixed 62.5%]. Having said that: I don't think amounts <40% or >140% will serve many, if any practical purposes.

sorgelig:
I was incorrect in calculation. Values are 0-93.75% and 100-193.75%. Steps are 6.25%.
100% for off will give no change on output. What is the use of it? Basically it will equal white color in mask.

tonurics:
Excellent. After I wrote my comment, I was thinking about your scale: I started wondering how you came up with steps of 5.5333% :)

While you're probably right about getting requests for it; much like there will probably be a request for "on" at 200%. I would still say that "off" at 100% would serve many more practical purposes than 0%; as it would allow for masks that only brighten. But ultimately: it's a very minor point, not worth debating. Having the 93.75% is a much bigger deal and resolves the concern I voiced earlier.

tonurics:
@sorgelig It looks like you edited your comment between my reading it and submitting a reply. So for the shake of completeness, here's side by side comparison:
sample
 
I think the difference between the two at 6.25% is negligible. So feel free disregard the change request and consider everything I brought up resolved. [I just shared the same image on Discord and told anyone there: if they feel otherwise, to reach out with feedback.]

sorgelig:
This is test version with pre-compiled `Main` and `NES`. Also included one shadowmask v2 file. Read the description in `VGA v2.txt` file. Please test it:
[shadowmask_v2.zip](https://github.com/MiSTer-devel/Main_MiSTer/files/7701691/shadowmask_v2.zip)

Note: supported both v1 and v2 masks files, so there in no need to re-work of old masks. I've changed enable bit in the core to disable masks in cores with older framework. Since only several cores were released with masks, i will re-compile them with new framework.

I was thinking if color names are really useful.. Especially with intensity per pixel it doesn't look like descriptive.

sorgelig:
btw, for light masks it's anyway better to have less than 100% for off to balance the overall brightness. For example for bright mask with 106.25% it's appropriate to use 93.75% for off.  This is another argument why 100% for "off" isn't needed.

wickerwaka:
I haven't had a chance to check out the core yet, but the new format looks good to me. I think there was value in the alphabetic colors in the previous format, but I don't think there is much point with this new format.

sorgelig:
@tonurics did you try?

tonurics:
Yes. But I've had other commitments that prevented me from doing thorough testing. I did create a few v2 masks at the extremes of each range to confirm they worked and made a quick pass of some of the v1 masks. I normally test using the SNES core [avoiding the NES's colour palette], but as far as I could tell everything appeared correct.

I've gone over the format changes in my mind's eye and have mostly been explaining what the changes will mean on Discord. The GUI Editor can be a double edged sword, as it allows people to create masks without needing to know the technical implementations. Having said that, I hope to have some free time soon to create a new GUI Editor for the v2 format. Once that is done, I think we'll see a lot more people creating and testing v2 masks.

In summarization: I've seen no issues, I like the format changes and agree with wickerwaka that the numeric colour identifiers are good enough [especially as we are back down to only 8].

sorgelig:
ok. then i will update the cores with v2

PR #497, opened 2021-12-27
Only enable shadow masks if the data is correctly read

wickerwaka:
If shadow mask files are moved or deleted, or when people enable shadow masks without having a file selected, they will get a black screen because the shadow mask enabled flag gets set without any valid data being sent. This change defers the setting of the flag until the file has been correctly read.


sorgelig:
I need to rework that part anyway for multires masks. I will add this check.

PR #501, opened 2021-12-30
Change Analog Stick Reporting range to -128 to 127

henryrizzi:
This change sets the range multiplier to 128 from 127, and also changes the min negative value limit to -128.

According to the PadTest 1.0 application from Shendo, the PlayStation stick has an analog range from -128 to 127. When testing in the current PSX core, I'm seeing a range of -127 to 127. This becomes immediately apparent when trying to run forward in Ape Escape as only the max analog value will cause the main character to run.

I've values from an original console (SCPH 5501) along with before and after from the PSX core.
SCPH 5501 w/ dualshock 2:
84B85794-CAC4-4440-BA6A-506D874EC408-COLLAGE
Before change:
C8668608-0DD8-4341-A18C-6607B6837702-COLLAGE
After:
E975BF1A-45AA-47FC-83F7-E05CCB84A4B6-COLLAGE

I don't believe this causes issues for other cores that use analog inputs, but I checked in the input tester and Vectrex core and the change does not seem to break any behavior.

### DIFF
--- a/input.cpp
+++ b/input.cpp
@@ -2799,10 +2799,10 @@ static void input_cb(struct input_event *ev, struct input_absinfo *absinfo, int 
 					else value = 0;
 				}
 
-				value = (value * 127) / hrange;
+				value = (value * 128) / hrange;
 
 				//final check to eliminate additive error
-				if (value < -127) value = -127;
+				if (value < -128) value = -128;
 				else if (value > 127) value = 127;
 
 				if (input[sub_dev].axis_pos[ev->code & 0xFF] == (int8_t)value) break;


RobertPeip:
If we are not 100% sure, we can also use the "is_psx()" function that already exists, e.g. here:

https://github.com/MiSTer-devel/Main_MiSTer/commit/3a1ec245de577e3b73e2150a1725d78068194672#diff-70b2ca68052525affed6efe3b203e780d3e59218b57184a75154c097a4ae2ac8R1624

sorgelig:
Many cores have inverted analog range. -128 cannot be negated (NEG -128 = -128). So to simplify it in verilog, Main supplies "valid" values which can be negated. 
In PSX core it's easy to subtract 1 if value is negative. It will loose -1 value but it's inside the drifting range.

henryrizzi:
I just added a commit which makes this change only apply to the PSX core by using the helper function from `user_io.cpp`. But not being able to easily invert from verilog does seem inconvenient.

PR #503, opened 2022-01-05
Fix Minimig kickstart path buffer overflow

matthargett:
While getting my local build environment set up, GCC 10 pointed me at this buffer overflow.
It stems from there being an "old" minimig config struct whose kickstart field still is 1024 bytes, but the new minimig config struct's kickstart field is 992 bytes. 
Since they're static arrays in structs anyway, just use `sizeof() -1` to make the code more resilient to future changes.


sorgelig:
It's not exactly overflow. It's POSSIBLE overflow if your path will exceed 992 bytes.

PR #513, opened 2022-01-18
Double Vrate CRT Monitors

neofreno:
Doubles the vertical refresh rate for crt monitors without changing the resolution.
This allows the original resolution unlike the forced_scandoubler option which doubles the vertical resolution.


sorgelig:
I didn't get what it really solves. Doubling vertical resolution isn't negative but rather positive side effect since you can apply either HQ2x or scanlines. Adds unneeded output latency. Although it may be very small but it's not worth it. Scaler also cannot properly sync (vsync_adjust) in such condition as it expects the same input and output refresh rate.

Basically VGA output is there for purity, to get an original video. Scandoubled video is STILL considered as an original video because it doesn't alter it besides just repeating of every line.

sorgelig:
forgot to tell, doubling refresh rate will add a negative side effect called "Stroboscopic movements". Every smooth move will be jerky.

neofreno:
Thank you for your quick response.
The idea is that CRT monitors can display the original resolution. As the monitors do not accept resolutions below 30khz we double the frequency to achieve this. When displaying the original resolution the scanlines are generated by the CRT itself and are not the effect of a filter. 
It is one more option given to the user so that he can choose a different way of viewing.
It is activated by a variable in the ini and does not conflict with any other option.
I attach two screens without any filter 2 with the scandouble and 2 with the modification, in them you can see the difference.
photo6003782814706809110
photo6003782814706809111
photo6003782814706809112
rcontent.com/35410501/150386163-b8567261-d0dd-45de-a708-70ee24ee86aa.jpg)

sorgelig:
Underlying adjustment complexity and stroboscopic moves doesn't make this option viable.
This video mode is not standard as well because it uses 100/120Hz and low vertical resolution. I believe not all monitors will be happy with it.

Btw, in scandoubled mode you didn't enable scanlines. If you will enable it, then it should be pretty much similar to your modification.

neofreno:
Hello.
Thank you for your attention.
The compatibility issues are correct, as not all monitors support 120hz.
The idea was that whoever has a compatible monitor should use the option, not everyone has a CRT monitor just as not everyone has a CRT TV or a PVM multiformat.

Regarding the scandouble activated in the core the difference is the brightness. I leave you a couple of images. The first one is with scandouble and the second one with the modification.

Greetings and thank you very much again.
554FAFF3-DA60-47EC-A992-0F443D3A4CF7
755F8AB7-1127-48CE-B4F7-20178DE97984

sorgelig:
The difference in these 2 pics is image compression of shooting camera. Being magnified to such value you can see difference in JPEG compression.

neofreno:
Both photos are with the same camera, the same angle, the same zoom and the same compression.
It is very difficult to demonstrate the difference as a photo can be affected by focus, light, zoom...

I just want to add one more option for those who want to use it, the alternative is to use scaling and add a custom video line for each core in Mister.ini, but with the amount of cores it is unfeasible.

It's a pity that you don't accept it.
Thank you very much for your attention.

sorgelig:
As i've told you above, this change switches scaler to the way it never supposed to work. I don't want to add such constraints to system. There are very large settings with only very limited use and obvious video side effects (awful stroboscopic moves like on very first 120Hz TVs). Adding such feature will require support it and taking in account in the future development which is not desirable.
Sorry, it changes a lot in code with doubtful advantages.

PR #520, opened 2022-02-02
Adaptive scanlines support

wickerwaka:
Send adaptive scanline data to cores that support it.

I know this code was only recently changed and now I'm changing it again. Sending three different possible forms of filter data was becoming cumbersome.

Added a `ScalerFilter` type instead of the `char` array to store the configuration data. The written format is unchanged so peoples settings won't get reset.

Reads a filter file into the `VideoFilter` struct. 16 phase filters are expanded to 64 phases in `read_video_filter`. `send_video_filters` then sends the phases depending on what version of the filters the core supports.

Required for https://github.com/MiSTer-devel/Template_MiSTer/pull/49


sorgelig:
btw, converting str[1024] to structure may break compatibility as compile would like to align the members to 4 byte boundary. While force byte alignment may work in some cases, it not always works too.

PR #524, opened 2022-02-04
GunCon 2: calibration menu/save/load

NolanNicholson:
This PR enables per-core calibration for the Namco GunCon 2 lightgun, using a similar pattern to the Wiimote.

I've used a new input quirk, QUIRK_GUNCON2. This quirk may be generalizable to other lightguns that don't have other bells and whistles (like the Wiimote does.)

### DIFF
--- a/input.cpp
+++ b/input.cpp
@@ -856,6 +856,7 @@ enum QUIRK
 	QUIRK_TOUCHGUN,
 	QUIRK_VCS,
 	QUIRK_JOYCON,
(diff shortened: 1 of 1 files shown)


sorgelig:
i will add more generalized quirk when more guns will be available and some generic handling will be extracted.

sorgelig:
honestly speaking, if i would be a fan of lightguns, then i would take wiimote internals and integrated into some gun, so it will get all benefits of wiimote and look like a gun.

PR #529, opened 2022-02-07
psx.cpp: fix memory leak

makigumo:


sorgelig:
i don't think it's correct. `new char[sz]` gives a single chunk of memory, not pointers to chunks.

sorgelig:
more over, delete[] is supposed to be used when array of objects is getting destroyed, so it's supposed to be some class with destructor, so in each element of array destructor is called. 
Here is simple array of chars which is a single chunk of virtual memory. so it's like malloc/free analogs. Nothing gets leaked.

makigumo:
AFAIK, when you allocate an array with new [], you have to use delete [].
Here (https://godbolt.org/z/dEcb3brfY) is an example, where the compiler complains about it.
```

<source>:5:1: warning: 'delete' applied to a pointer that was allocated with 'new[]'; did you mean 'delete[]'? [-Wmismatched-new-delete]
delete(toc);
^
      []
<source>:4:13: note: allocated with 'new[]' here
char *toc = new char[3];
            ^
1 warning generated.
Compiler returned: 0
```

sorgelig:
https://godbolt.org/z/rT3Wh9Pce
no errors.

wickerwaka:
Yes, `new[]` should be paired with `delete[]`. In this case it doesn't make any functional difference since it is a `char` type which has no destructor and nothing is overriding the new/delete operators (so `operator delete[]` just calls `operator delete` to deallocate). So it's not a memory leak, but it is a line of code that could eventually trigger a warning in newer compiler versions.

PR #533, opened 2022-02-10
Video Processing menu improvements

zakk4223:
Show filename instead of entire path.
On file picker lines, enable plus/minus to move through files in the same directory quickly.

### DIFF
--- a/menu.cpp
+++ b/menu.cpp
@@ -42,6 +42,9 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #include <sys/wait.h>
 #include <libgen.h>
 #include <bluetooth.h>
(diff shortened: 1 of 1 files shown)


sorgelig:
I didn't get the meaning of +/-.. doesn't pgup/pgdown or left/right work for you?
patch seems too complicated, i can make it simpler if it's just to display names without paths.

zakk4223:
maybe easier to describe via short video:

https://user-images.githubusercontent.com/22642/153356149-83f6eebf-1b7d-416f-8486-8549be41399c.mp4

sorgelig:
i see

sorgelig:
i've added this functionality.

PR #536, opened 2022-02-13
PSX: Original way to deal with mencards with file 'originalwaymemcards.txt' on PSX directory

wwark:
To permits user to use the original way of memory cards on PSX core.
Tested with and without file it is working as expected.
Thanks,

### DIFF
--- a/support/psx/psx.cpp
+++ b/support/psx/psx.cpp
@@ -252,7 +252,9 @@ void psx_mount_cd(int f_index, int s_index, const char *filename)
 			if (!loaded) Info("CD BIOS not found!", 4000);
 		}
 
-		if(*last_dir) psx_mount_save(last_dir);
+		sprintf(buf, "%s/originalwaymemcards.txt", HomeDir());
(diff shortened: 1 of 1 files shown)


sorgelig:
I don't see problem to place memory card into saves. It's also more convenient to have all saves in one place so later you may backup the whole saves. You may have original CD archive and don't bother to backup it, and later you realize you've copied one memory car there and saved progress and now it's lost. 
So, to unify modifiable and non-modifiable places it's better NOT to allow memory card from CD folder.

sorgelig:
Your commit won't make it work as before. Mounted memory card won't be remembered, so you have to mount it manually every time you load the core.
I will make other way to prevent memory card creation till game save anything. So it will prevent creation of many memory cards if you just look some games without playing.

sorgelig:
Persistent memory card is Implemented in latest changes. For reason described above, memory cards still have to be in saves/PSX folder.

wwark:
Thanks a lot @sorgelig for this change. it is working great !

Issue #537, opened 2022-02-13
First controller input is consequently ignored in OSD


littlegamer87:
As a minor issue, I would like to report that the MiSTer appears to ignore the first controller input when the device is powered on.
Controller used is a Switch Pro over Bluetooth (TP-Link UB400)

Steps to reproduce: Power on the MiSTer, allow time to load bluetooth and pair with the controller. The first input on the controller to control the OSD (dpad or buttons) is never processed. Waiting to press a button after pairing changes nothing. The second input press always works, and the ones after that too.

When loading a core, the issue reappears. EG after loading the SNES core (which requires me to operate the OSD after booting, where the first input is missed), the first input in the SNES OSD is missed as well. The GBA core does not open the OSD by default, what's interesting is that pressing the OSD button after loading the core (which I assigned to the home button on my controller), brings up the OSD without missing the input. After the OSD appears, the first input is missed again.

Tested on v220211, OS v211112

birdybro:
I think this is intentional. Whatever controller receives a button input first upon loading a core becomes "Player 1". Whatever controller receives a button press second is assigned to "Player 2", and so on...

littlegamer87:
Thanks for your feedback. I understand your reasoning but is it required to ignore the input for that?

paulb-nl:
I am not able to reproduce this with a wired controller. First input always works.

sorgelig:
I cannot reproduce it using Bluetooth controller (Apex 2).

birdybro:
Ah I must have misunderstood you. I tend to do that. I cannot reproduce that with a DualSense (DS5) or an 8bitDo SN30 Pro 2. I also paired a Switch Pro controller with the 8BitDo adapter, same result, it registers on first press. 

However, I **was** able to reproduce it when I used my Edimax BT500 bluetooth adapter to pair the Switch Pro controller.

First time I press down (nothing happens on screen):

```
057e:2009:00 P0 Input event: type=EV_ABS, Axis=17 [-1...1], Offset=1
Device A4:38:CC:6D:6D:23 assigned to player 1
FileOpenEx(open) File:/media/fat/config/kbd_057e_2009.map, error: No such file or directory.
PS2 code(make)(ext) for core: 114(0x72)
057e:2009:00 P1 Input event: type=EV_ABS, Axis=4 [-32767...32767], Offset=1510
PS2 code(break)(ext) for core: 114(0x72)
```

Second time:

```
[...]

PR #539, opened 2022-02-13
controller: add switch pro rev pid exemption

birdybro:
My switch pro controller has a different pid than the ones in the list that have accelerometer exemptions. This should resolve that. And could potentially alleviate --> https://github.com/MiSTer-devel/Main_MiSTer/issues/537

### DIFF
--- a/input.cpp
+++ b/input.cpp
@@ -3753,7 +3753,7 @@ int input_test(int getchar)
 							input[n].num = 2; // force mayflash mode 1/2 as second joystick.
 						}
 
-						if (input[n].vid == 0x057e && (input[n].pid == 0x0306 || input[n].pid == 0x0330))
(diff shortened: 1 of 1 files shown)


birdybro:
@sorgelig I have to look at this more closely, as it's now inverted the behavior of the Switch Pro controller with regard to up and down.

birdybro:
upon further investigation... The joycon logic in line 3488 is pointing to the pro controller's pid not the joycon pid.

Joycon - https://devicehunt.com/view/type/usb/vendor/057E/device/2006

Switch Pro - https://devicehunt.com/view/type/usb/vendor/057E/device/2009

birdybro:
Just confirmed. Left joycon is 057e:2006, right joycon is 057e:2007. Current code cited above is treating the Switch Pro controller as a joycon.

birdybro:
https://github.com/MiSTer-devel/Main_MiSTer/pull/541 - regression fixed with this PR

PR #541, opened 2022-02-13
Change switch pro accelerometer exemption to explicit

birdybro:
This resolves the regression in the previous commit by defining explicit behavior just for the Switch Pro controller. Sorry about the mistake earlier.


birdybro:
Bad assumptions on my part.

PR #543, opened 2022-02-15
Auto rotation and video info

wickerwaka:
Read the rotation flag when updating the video info. When the simulated display is rotated, when rotate the shadow masks and video filters also. Also calculate vscale using the width instead of the height so integer scaling is applied correctly.

Save the video info from the core and display it in the Information area (as long as there isn't a battery connected!)

Video Info in OSD

Split up code a little so retrieving the info, showing it in based on `video_info` and apply the video mode are separate functions.

Depends on https://github.com/MiSTer-devel/Template_MiSTer/pull/50 for the rotation flag.

### DIFF
--- a/input.cpp
+++ b/input.cpp
@@ -3753,7 +3753,7 @@ int input_test(int getchar)
 							input[n].num = 2; // force mayflash mode 1/2 as second joystick.
 						}
 
-						if (input[n].vid == 0x057e && (input[n].pid == 0x0306 || input[n].pid == 0x0330))
+						if (input[n].vid == 0x057e && (input[n].pid == 0x0306 || input[n].pid == 0x0330 || input[n].pid == 0x2009))
 						{
 							if (strcasestr(input[n].name, "Accelerometer"))
 							{
--- a/input.cpp
+++ b/input.cpp
@@ -3753,7 +3753,7 @@ int input_test(int getchar)
 							input[n].num = 2; // force mayflash mode 1/2 as second joystick.
 						}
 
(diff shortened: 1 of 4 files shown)


wickerwaka:
One thing I haven't tested but should: scandoubler fx scanlines. Those scanlines won't rotate, but the filters will, probably causing some undesirable results. Maybe the rotation flag should not be set if scandoubler scanlines are enabled?

sorgelig:
rotated video should not rotate shadow mask just by original logic. Original CRT monitors used vertically still were normal monitors with horizontal mask.

sorgelig:
Neither scanlines nor shadow mask should rotate. Vertical CRT display doesn't rotate hs/vs signals. Scanlines on vertical display should be vertical!

wickerwaka:
That's what this change does, but it's confusing to talk about! The rotation flag is set when we are *simulating* a rotated display using the `screen_rotate` module, in that case you do want the filters and masks to rotate.

sorgelig:
There was no intention to simulate rotated display because it gives dissonance looking on normal horizontal display with vertical scanlines.

wickerwaka:
I don't see what problems it creates but it does solve two problems:
* Fixes `vscale_mode` for these cores, currently people have to use calculated `vscale_border` values in order to get integer scaling
* No need for separate vertical and horizontal video presets

sorgelig:
agree for vscale_mode, but not vertical scanlines on horizontal display. Presets already can do it - so leave it there.

wickerwaka:
Updated it so it no longer does the rotation.

PR #544, opened 2022-02-16
mra_loader: show error message on screen

paulb-nl:
Only clear error_msg if MD5 is not empty or None
Show full path with not found error so it is clear which folder it is
looking in.


sorgelig:
Please re-do this PR with a new fork. I removed reverted commits but it seems github wants them to be added again..

paulb-nl:
Ok done.

PR #545, opened 2022-02-17
Correct PSX light gun x/y spans + improve CRT gun performance

NolanNicholson:
The PSX core has support for using an analog joystick or a light gun to emulate the Namco GunCon.

The framework reports analog joysticks and light guns using a range of +/-128. The original GunCon reports its horizontal coordinate in 8 MHz clocks since HSYNC, and its vertical coordinate in scanlines since VSYNC.

The GunCon range is not half or double the analog joystick range, so an integer multiply would be required to remap the coordinate inside the PSX core. To economize on FPGA resources for that core, that remapping is done here in the ARM instead, where it is cheaper.

### DIFF
--- a/input.cpp
+++ b/input.cpp
@@ -856,7 +856,7 @@ enum QUIRK
 	QUIRK_TOUCHGUN,
 	QUIRK_VCS,
 	QUIRK_JOYCON,
-	QUIRK_GUNCON2,
+	QUIRK_CRTGUN,
 };
 
 typedef struct
@@ -1221,7 +1221,7 @@ int input_has_lightgun()
 	{
 		if (input[i].quirk == QUIRK_WIIMOTE)  return 1;
 		if (input[i].quirk == QUIRK_TOUCHGUN) return 1;
-		if (input[i].quirk == QUIRK_GUNCON2) return 1;
+		if (input[i].quirk == QUIRK_CRTGUN) return 1;
 	}
 	return 0;
 }
@@ -3819,7 +3819,7 @@ int input_test(int getchar)
 						//Namco GunCon 2
 						if (input[n].vid == 0x0b9a && input[n].pid == 0x016a)
 						{
-							input[n].quirk = QUIRK_GUNCON2;
+							input[n].quirk = QUIRK_CRTGUN;
 							input[n].lightgun = 1;
 							input[n].guncal[0] = 25;
 							input[n].guncal[1] = 245;
@@ -4274,7 +4274,7 @@ int input_test(int getchar)
 									}
 								}
 
-								if (ev.type == EV_ABS && input[i].quirk == QUIRK_GUNCON2)
+								if (ev.type == EV_ABS && input[i].quirk == QUIRK_CRTGUN)
 								{
 									menu_lightgun_cb(i, ev.type, ev.code, ev.value);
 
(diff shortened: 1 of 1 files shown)


NolanNicholson:
Note: this change is not specific to the actual GunCon or GunCon 2 hardware - it should also fix the accuracy of other guns like the Wii Remote.

NolanNicholson:
Added another commit to input.cpp that improves the performance of the GunCon 2, particularly on NES games. I've also renamed QUIRK_GUNCON2 to QUIRK_CRTGUN, since other CRT guns should work with the same accommodations if they become available.

sorgelig:
Main already normalize analog joystick value according to min/max reported by driver. So you have to check driver code and fix it there if min/max values aren't real.

sorgelig:
> The GunCon range is not half or double the analog joystick range, so an integer multiply would be required to remap the coordinate inside the PSX core.

for gun mode there is no dead-zone in center, so it doesn't need to be equal for + and -. Gun just delivers current X/Y value within min/max value. You may set min=0 max=<your value> for both X and Y in driver. It will be of course converted to -128..+127 value and core must translate it to coordinates. This behavior cannot be changed as it affects other gun input devices.

NolanNicholson:
Affecting other gun input devices (Wiimote, etc.) is intentional here. Those will also not currently map correctly to the GunCon in the PSX core, unless the user calibrates their gun to have the calibrated min/max far beyond the actual edges of the screen.

I spoke with Robert about including the conversion to coordinates inside the PSX core, and he said there would not be room for it. The offset itself is already inside the core - it's modifying the "span" that's the problem.

sorgelig:
I will check it. I'm not a big fan of quirks without strong reason.

NolanNicholson:
I will educate myself a bit further and see what I can do to compile it with more general wi-fi support. Thanks!

NolanNicholson:
Apologies - I reopened this by mistake.

sorgelig:
i was talking with Robert. It can be simplified and made in FPGA.
Which PSX games use lightgun so i can test it?

Gazpr0m:
From some forum post:
Here's a list of all of the light gun game for the PS One. The type of gun used is in the brackets. 

Area 51 (Justifier)
Crypt Killer (Justifier)
Die Hard Trilogy (Justifier)
Die Hard Trilogy 2: Viva Las Vegas (Justifier/Guncon)
Elemental Gearbolt (Justifier/GunCon)
Ghoul Panic (GunCon)
Gunfighter: The Legend of Jesse James (GunCon)
Judge Dredd (lightgun?)
Lethal Enforcers 1&2 (Justifier)
Maximum Force (Justifier/GunCon)[1]
Moorhuhn series (GunCon)
Point Blank series (GunCon)
Project Horned Owl (Justifier)
Rescue Shot (GunCon)
Resident Evil: Survivor (GunCon)
Revolution X (Justifier)
Time Crisis (GunCon)
Time Crisis: Project Titan (GunCon)

NolanNicholson:
Thank you, Sorgelig. I will submit GunCon coordinate correction to the PSX core directly, and Robert and I will work to fit it within the core's resource constraints.

I will submit the CRTGUN improvements in a separate PR later.

PR #549, opened 2022-02-20
Add CHD image support for PSX

zakk4223:


bootsector:
OMG! 😱

thehughhefner:
Nice job Zakk!

makigumo:
Well done! Could you please use tabs for indentation like in the rest of the file?

support/psx/psx.cpp
```diff
@@ -107,6 +123,35 @@ static void unload_cue(toc_t *table)
...
+{
+
+  unload_chd(table);
+  chd_error err = mister_load_chd(filename, table);
+  if (err != CHDERR_NONE)
+  {
+    return 0;
+  }
+
+  for(int i = 0; i < table->last; i++)
```
tjanas, line 138: Does the compiler optimize out the i++ to make it ++i in machine code?

sorgelig: table->last is not the last after processing, it's total number of tracks. If you are talking about it.

PR #555, opened 2022-02-25
PSX: Try loading optional cd_bios.rom from both game dir and parent dir

Optiroc:
When loading a PSX game try loading `cd_bios.rom` from both game and parent directory before falling back to `boot.rom`. This is as stated in the README and doesn't force a multi-level hierarchy for enabling optional BIOSes.

> You can also place a cd_bios.rom in the same directory as the CD or 1 directory above, to have it uses together with that CD. This can be used for games that depend on the BIOS region(US,EU,JP).

### DIFF
--- a/support/psx/psx.cpp
+++ b/support/psx/psx.cpp
@@ -616,12 +616,24 @@ void psx_mount_cd(int f_index, int s_index, const char *filename)
 					{
 						int bios_loaded = 0;
 
-						strcpy(buf, last_dir);
+						// load cd_bios.rom from game directory
+						sprintf(buf, "%s/", last_dir);
 						p = strrchr(buf, '/');
 						if (p)
 						{
 							strcpy(p + 1, "cd_bios.rom");
 							bios_loaded = load_bios(buf);
+						}
+
+						// load cd_bios.rom from parent directory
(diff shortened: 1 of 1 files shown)


sorgelig:
Currently supported structure like this:
top level: EU,US,JP. Then inside each folder you place cd_bios.rom and games you want.
What your patch adds?

Optiroc:
It enables a bit more freedom in how you chose to structure the content.

You can have a flat hierarchy:
```
PSX/
  GameA/
    GameA.chd
  GameB/
    GameB.chd
  OddGame/
    cd_bios.rom
    OddGame.chd
```

And for hierarchies you can sort by other factors than region, for examples genres, while freely mixing regions or games with BIOS quirks:
```
PSX/
  RPG/
    AnRPG/
      AnRPG.chd
    WeirdRPG/
      cd_bios.rom
      WeirdRPG.chd
  Shoot em Up/
...etc...
```

Personally I prefer a flat hierarchies, and how the first example is structured works in the other CD cores. So I was surprised that it didn't in the PSX core since the README suggested otherwise.

sorgelig:
i think it's because you use chd which before wasn't supported. With chd you shave one level of folders.

PR #557, opened 2022-02-26
IDE: Semi-implement READ_VERIFY and suppress debug output

alexoughton:
This is a solution for using NTFS under Windows NT on ao486.

Without supporting ATA command 0x40 (READ_VERIFY), NTFS conversion in Windows NT setup (or using convert.exe) gets stuck in an infinite loop. This PR provides an OK response to the 0x40 command although it does not actually provide any verification of a good read.

Additionally there are other unsupported commands which are called by NT during the conversion process which are OK to ignore/fail. However, they occur so rapidly during the conversion loop that this causes a flood of printf debug messages which cause significant slowdown to IDE access. This PR therefore also moves "unknown command" to dbg_printf so they are suppressed by default.

### DIFF
--- a/ide.cpp
+++ b/ide.cpp
@@ -928,6 +928,12 @@ static int handle_hdd(ide_config *ide)
 		ide_set_regs(ide);
 		break;
 
+	case 0x40: // READ VERIFY
+		dbg_printf("Received read verify command. Not implemented but returning OK.\n");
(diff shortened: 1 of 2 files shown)


sorgelig:
what is the purpose to patch debug output in cdrom?
These prints were intentionally left, so it will be easy to see on user side if game misbehaves.
As long as it doesn't add problem in your specific case, it doesn't need to be changed.

PR #559, opened 2022-02-27
Moved the parsing of MGL files further up the process list

Flandango:
Moved the parsing of MGL files before the parsing of the config settings, right after the parsing of MRA files.
If the core loads any images or mounts any files during configuration parsing, mgl_done is set to 1 and not reset after parsing the mgl file.  Moving it up to just after the parsing of MRA files fixes this issue.


sorgelig:
i've already fixed this problem.

Flandango:
Sounds good and thanks for all your work!

PR #564, opened 2022-03-05
Auto-bootstrap GCC, and force plink to cache the host key

c0d3h4x0r:
Just a suggestion to make things a bit easier.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,6 +19,7 @@
 
 # ignore directories
 Debug
+gcc-*
 
 host
 *.lst
--- a/build.sh
+++ b/build.sh
@@ -14,7 +14,7 @@ set -o pipefail
 make
 
 set +e
-plink root@$HOST -pw 1 -batch 'killall MiSTer'
+echo y|plink root@$HOST -pw 1 'killall MiSTer'
 
 set -e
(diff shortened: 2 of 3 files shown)


sorgelig:
this is wrong. there can be different compatible versions of gcc. It's not corrct to forcing to specific gcc version/build as it will mess developing environment.

c0d3h4x0r:
Would this be okay as a default, only if `CC` is not already set?

sorgelig:
i use this script as part of build process. It's not supposed to download anything. It's not for normal user, it's for developer. Therefore such auto downloads aren't acceptable here. I prefer it fail rather than auto-modification in system.

c0d3h4x0r:
Yes, I know who this script is for and what it does.

Most developers like to save time and make things easier on themselves.  My goal here is to make building easier for the developer who is trying to build this repo for the first time, and for the developer who hasn't fetched and built it in a long time so their GCC toolchain is out-of-date and causing build errors.

Would you object to a new/separate script called something like `setup_build_toolchain.sh`?

sorgelig:
ok

c0d3h4x0r:
Done.

PR #567, opened 2022-03-11
Override configuration based on video mode

tsowell:
This adds support for overriding configuration variables for specific video modes.  The intended purpose is to support automatically switching to an output mode that is best-suited to the core's video mode.  It's especially useful for letting CRTs do their own scaling for "rectangular pixel" modes like 320x200.

Overrides are specified at the end of MiSTer.ini in sections with names in the format `video=WIDTHxHEIGHT[@VREFRESH]`.

For example:

```
[video=320x200]
video_mode=640,16,96,48,400,12,2,35,25175
vscale_mode=2

[video=720x400@70.1]
video_mode=720,15,108,51,400,11,2,32,28322
```

When a core changes video mode, MiSTer checks for a configuration section matching the new mode and swaps in the new configuration if there is a match.  Sections that match width, height, and vertical refresh rate take precedence.  Then if there is no match, it looks for a section matching width and height but with no vertical refresh rate specified.

By the way, I tried to take a light touch with the changes to cfg.cpp, and one drawback is that it stores a copy of the entire cfg struct for each video mode override section... which is a bit wasteful.  Let me know if you'd like me to take a different approach.

### DIFF
--- a/MiSTer.ini
+++ b/MiSTer.ini
@@ -195,3 +195,14 @@ bt_reset_before_pair=0
 ; This option cannot be used when defmra in CONFSTR is used (i.e. if arcade rbf is loaded directly not through MRA).
 ; This option is ignored for Menu core.
 ;waitmount=/media/usb0
+
+; Overrides for video mode
+; When the core's video mode matches the parameters in the section header, any options in the section override options from MiSTer and core sections.
+; Refresh rate in header is optional and, if present, must match exactly the output from video_info or the logs.  For example, if it says "60.0Hz", the header needs to be "@60.0" to match.
+; When the core changes video mode, MiSTer will first look for a matching WIDTHxHEIGHT@VREFRESH section.
+; If no match is found, it will fall back to a matching WIDTHxHEIGHT section with no refresh rate.
+; If there is still no match, MiSTer/core options will be used without overrides.
+; [video=640x400]
+; ...
+; [video=640x400@70.1]
+; ...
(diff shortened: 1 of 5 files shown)


tsowell:
Thanks for looking at this and responding.  What do you think about doing this by reloading the configuration file, like how the core sections work?  I think that would meet your requirements.  Do you see a problem with doing that every time the video mode changes?

sorgelig:
At first: if there is no resolution sections used then code should avoid to reload of INI, so it should work exactly as before changes, so users not using this feature won't be affected by any possible side effect. Reloading ini gives some latency in processing other FPGA requests and input devices. If core will change resolution together with some other events (like reading CD data) then it may impact the work. I cannot tell how it will affect the work in reality...

tsowell:
Thanks - I was worried about impacting users who are not using the feature, so that's a great idea.

I just pushed a version that works by reloading INI and only reloads if there were video sections in the initial load.  Where possible, I tried to make it consistent with the way the core sections work.  Please let me know if you have any other thoughts.

sorgelig:
why HAS_VIDEO_SECTIONS and USING_VIDEO_SECTION added as possible options in ini? According to code they are calculated variables.

tsowell:
Just to make use of the existing cfg interface, but I understand why that wouldn't be the right place for them.  Here's a commit that moves them into separate variables.

sorgelig:
i didn't mean to move them. Just don't expose them into INI file would be fine. Anyway..

PR #583, opened 2022-04-21
Add Fanatec wheels

coolbho3k:
The ranges are still not working but we can figure that out later:
- Wheel center is -64, goes from -127 (left) to 0 (right).
- Pedals start from 0 and go negative.

### DIFF
--- a/input.cpp
+++ b/input.cpp
@@ -4068,6 +4068,58 @@ static void setup_wheels()
 					break;
 				}
 			}
+			else if (input[i].vid == 0x0eb7) // fanatec
+			{
+				struct ff_effect fef;
+
+				fef.type = FF_SPRING;
+				fef.id = -1;
+
+				fef.u.condition[0].right_saturation = 0xFFFFUL * cfg.wheel_force / 100;
+				fef.u.condition[0].left_saturation = 0xFFFFUL * cfg.wheel_force / 100;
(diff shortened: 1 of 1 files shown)


input.cpp
```diff
@@ -4068,6 +4068,57 @@ static void setup_wheels()
 					break;
 				}
 			}
+			else if (input[i].vid == 0x0eb7) // fanatec
+			{
+				struct ff_effect fef;
+
+				fef.type = FF_SPRING;
```
coolbho3k, line 4075: No support for autocenter effect in that driver

input.cpp
```diff
@@ -4068,6 +4068,57 @@ static void setup_wheels()
...
+					play_ev.type = EV_FF;
+					play_ev.code = fef.id;
+					play_ev.value = 1;
+					write(pool[i].fd, (const void *)&play_ev, sizeof(play_ev));
+				}
+
+				if (cfg.wheel_range && input[i].sysfs[0])
+				{
+					sprintf(path, "/sys%s", input[i].sysfs);
+					strcat(path, "/device/range");
```
coolbho3k, line 4099: Slightly different path here

sorgelig:
I think can add additional parameter "wh_type" which will change pedals range handling.

PR #587, opened 2022-04-25
Offer force reloading of volume configs

wizzomafizzo:
This patch adds a "vol_load" cmd to the /dev/MiSTer_cmd interface so the Volume.dat file can be read on request

As far as I've been able to tell, MiSTer offers no way to adjust system volume from linux besides writing directly to memory or simulating key presses of the volume media keys. It's possible to set a value in Volume.dat, but this value is not read enough to be of use right now (I understand that this is not the intended purpose)

This feature is useful to me so that the volume of background music can be adjusted in the menu. I believe it would also be of use for the Super Attract Mode script. There may be other use cases for applications like mistercon, webmenu and bcm

### DIFF
--- a/input.cpp
+++ b/input.cpp
@@ -27,6 +27,7 @@
 #include "video.h"
 #include "joymapping.h"
 #include "support.h"
+#include "audio.h"
 
 #define NUMDEV 30
 #define NUMPLAYERS 6
@@ -5053,6 +5054,7 @@ int input_test(int getchar)
 					{
 						user_io_screenshot_cmd(cmd);
 					}
+					else if (!strncmp(cmd, "vol_load", 8)) load_volume();
 				}
 			}
 


sorgelig:
Patch is not required. You just inject keyboard presses of media control keys mute/volUp/volDown

birdybro:
maybe something like this can be done @wizzomafizzo ?

http://kbd-project.org/docs/howto-linux-keyboard-and-console/kbd-4.html

wizzomafizzo:
> maybe something like this can be done @wizzomafizzo ?

I can try it again but I think this will only let me send control characters to the active shell. Mister menu won't see them

I appreciate that we don't want to bloat the code with unnecessary features and I'm trying really hard to work with what I'm given on the image. But I spent a few hours on this so far and I can't find a way to do it in a script without pulling in an extra binary just to send key presses

wizzomafizzo:
A note on the proposed workaround as well to inject keypresses. It appears to take up to 2 seconds to initialise a uinput device and send a keypress. This doesn't allow a cohesive experience for the user since it's a noticeable delay

I'm happy to be corrected, but it seems that the only way for me to make this solution work well is for me to run a daemon process at all times of a virtual keyboard just so I can press media keys fast enough. I also can't set the volume to a specific value, I need to read the contents of Volume.dat and send multiple keypresses

Yes, nobody will die if mister never supports this feature, but doesn't that setup sound a bit silly? Looking into the future as well, we could end up in a situation where users have multiple unmaintained binaries lying around their mister for injecting keypresses

birdybro:
What's your use-case that you have in mind? Is it for someone running Super Attract Mode unattended that can't use a controller to turn down the volume but also doesn't have volume control on their television/cabinet/etc...?

wizzomafizzo:
For my own project, I have increasing reports now of users who run background music and it's playing too loud. They adjust the volume, launch a core, now the core is too quiet. They turn it back up. They exit the core, now the music is blaring again

A workaround for them is to convert all their music files to be quieter, but it can be a lot of work and now I need to maintain documentation for how to convert 4 different file formats on 3 platforms. I also don't think there is a way to make vgm files play quieter. mpg123 has a flag to adjust volume but it's marked as deprecated. ogg123, aplay and vgmplay don't offer this feature at all, so it would be inconsistent

My idea was to offer to the user a config option to change the system volume while menu is open, and then when a core is launched, set the volume back to its original state

I can't speak for the other projects, that was more a "hey, I'm not the only person trying to mess with the volume"

wizzomafizzo:
I imagine that's why it chooses to mute the cores in the first place then? In their case, I think this method of writing to Volume.dat already works well, because they are most interested in the volume of the core itself and that launch is when the value is read and updated. It doesn't work so well for me, because I can't guarantee the timing of inotify is perfect, and I want the user to be able to adjust the volume to their liking without multiple reboots

If my current patch is not appropriate I would still like to work something out. Admittedly it's a bit of a hacky solution, but I thought you might be more receptive to a patch if it was very small. In my opinion, basic volume control seems a pretty critical feature to offer to the linux system

birdybro:
No worries at all, I wasn't trying to come across as non-receptive, trying to understand and come up with a suggestion for a workaround in the meantime while this gets evaluated by the person in charge :)

sorgelig:
> I wish there was a "Unity Gain" achieved with each core design when adjusting the AUDIO_L/R by each dev

MiSTer provides volume attenuation per core available for user. So you can normalize volume per core

wizzomafizzo:
@sorgelig what can I do to make this patch acceptable for you? I really want to make it work

You have offered a workaround, but I do not think the workaround is acceptable for the reasons listed. I have done my due diligence to find an alternative. If there's no way in hell you are adding this feature, then please tell me that, but I am pleading with you I do not want to simulate keypresses

wizzomafizzo:
> i will think how to make it better

Thank you

mrchrisster:
Currently looking for a way to do the same thing for SAM. It would be great if volume could be controlled dynamically from the shell

thehughhefner:
@sorgelig sir did you manage to think about it?

Paradox:
Further, core volumes cannot be fully muted, or, turned down enough to be inaudible. Would be nice if one, or both, could be added, so core volume control mimics global volume control.

mrchrisster:
Having instant control over core/global volume would be awesome!

sorgelig:
Implemented volume command.
format: volume mute/unmute/0-7

wizzomafizzo:
Thank you sorg! This is great

PR #629, opened 2022-05-31
Input socket for external controller visualization

jaysonlarose:
This is a modification for MiSTer I've been working on which allows you to connect to a TCP socket and receive input events from the Linux input system. This is made mainly for creating controller visualization, like what many streamers use (video demo [here](https://www.youtube.com/watch?v=eq-ZMLZzzjs)), but there are other applications for getting access to the raw input events, as well. For example, I assign an unused button on my controller to serve as a mute/unmute button by hooking into my desktop PC's Jack audio system.

I've written a companion program that hooks into this modification: <https://github.com/jaysonlarose/mister_viz>

I wrote it to be minimally invasive with the existing codebase. It defaults to being disabled unless specifically enabled via MiSTer.ini, so it should only use up a few additional CPU cycles per input by default. One thing I did do was moving the definition of typedef struct devInput into input.h to make it available to my code, although at the moment all I use from it is the vendorID and productID section. I can just send the vendorID and productID directly in the call to `input_socket_send()` if that's a dealbreaker.

I'd love to see this incorporated into the main codebase, Please let me know what you think.


sorgelig:
This is out of MiSTer scope. If you want to add remote control over TCP, then you can make a side project through uinput. So some resident app will work with TCP and then simulate input event over uinput.

sorgelig:
It's still has no relation to MiSTer purpose. There is no plans to make a streamer device from MiSTer. Also feature you've made can be done differently. For example MiSTer sends joystick and keyboard events to core and code in that place simply can push/signal some events issued to the core. It will eliminate required polling. So any resident app may way for such event and then visualize it.
It doesn't mean it will be integrated into main code, but just an example of different and probably more efficient/thinner implementation.

jaysonlarose:
Bummer. I was hoping I wasn't going to have to keep a fork up to date. Thanks for your time.

PR #636, opened 2022-06-07
Simplify vscale_mode

wickerwaka:
Removed loop from vscale calculations, calculate max horizontal and
vertical scale and use the minimum.

vscale_mode has been changed to make it more reliable.

vscale_mode=4 - Generate modes that match the cores aspect ratio.
Generally better, will work with more cores and produce most optimal
resolutions, but will not be compatible with displays that always
stretch the image to fill.

vscale_mode=5 - Generate modes that match the original display aspect
ratio, if the core needs a wider aspect then the resolution is not
changed. Compatible with a wider range of displays.

Updated INI with a clearer description.

Removed mode 6 since it was not being used and adjusting display width
doesn't enable integer horizontal scaling.

### DIFF
--- a/MiSTer.ini
+++ b/MiSTer.ini
@@ -10,9 +10,8 @@ vscale_mode=0          ; 0 - scale to fit the screen height.
                        ; 1 - use integer scale only.
                        ; 2 - use 0.5 steps of scale.
                        ; 3 - use 0.25 steps of scale.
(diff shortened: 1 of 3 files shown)


sorgelig:
You are not right about mode 6. It has special purpose - it move non-integer scaling to display. Displays usually have more advanced scaling algo.

PR #650, opened 2022-06-26
Send freesync/VRR enable frames if the display supports it

zakk4223:
If the display advertises Freesync/Vesa VRR support via the edid, send the correct HDMI config frames to the display to enable it.
Many consumer TVs will handle non-standard refresh rates properly once VRR is enabled.
Users must set vrr_mode=1 to enable, the default is to send no VRR config frames.
If a display advertises both Freesync and Vesa VRR, Freesync is preferred.

This change also strobes the proper register on the ADV7513 that causes it to re-read the display edid. This is only done once per execution of main, so you must load/reload a core to get MiSTer to notice a change in display.

### DIFF
--- a/MiSTer.ini
+++ b/MiSTer.ini
@@ -27,6 +27,22 @@ osd_timeout=30         ; 5-3600 timeout (in seconds) for OSD to disappear in Men
 osd_rotate=0           ; Display OSD menu rotated,  0 - no rotation, 1 - rotate right (+90°), 2 - rotate left (-90°)                  
 vga_sog=0              ; 1 - enable sync on green (needs analog I/O board v6.0 or newer).
 
+
+; Variable Refresh Rate control
+; 0 - Do not enable VRR (send no VRR control frames)
+; 1 - Auto Detect VRR from display EDID. 
+; 2 - Force Enable Freesync
+; 3 - Force Enable Vesa HDMI Forum VRR
+vrr_mode=0
+; Freesync min/max framerate parameter defaults to the min/max capability reported by the display
+; Freesync minimum framerate. 
+vrr_freesync_min_framerate=0
+; Freesync maximum framerate
+vrr_freesync_max_framerate=0
+; VESA VRR base framerate. Normally set to the current video mode's output framerate
+vrr_vesa_framerate=0
+
+
 ; 1 - enables the recent file loaded/mounted.
 ; WARNING: This option will enable write to SD card on every load/mount which may wear the SD card after many writes to the same place
 ;          There is also higher chance to corrupt the File System if MiSTer will be reset or powered off while writing.
--- a/cfg.cpp
+++ b/cfg.cpp
@@ -96,6 +96,10 @@ static const ini_var_t ini_vars[] =
 	{ "WHEEL_FORCE", (void*)(&(cfg.wheel_force)), UINT8, 0, 100 },
 	{ "WHEEL_RANGE", (void*)(&(cfg.wheel_range)), UINT16, 0, 1000 },
 	{ "HDMI_GAME_MODE", (void *)(&(cfg.hdmi_game_mode)), UINT8, 0, 1},
+	{ "VRR_MODE", (void *)(&(cfg.vrr_mode)), UINT8, 0, 3},
+	{ "VRR_FREESYNC_MIN_FRAMERATE", (void *)(&(cfg.vrr_freesync_min_framerate)), UINT8, 0, 255},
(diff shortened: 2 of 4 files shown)


elexor:
great stuff! 
Is this true vrr support? because I can still see the pixel clock is being changed to adjust the refreshrate.

video mode switching and blankouts still occur with this. 

vrr should be fixed scanrate and fixed pixelclock with variable vertical blanking as the mechanism to adjust refreshrate.


the VBI should grow and shrink in realtime to adjust the refreshrate on a vrr display. this will ensure no display blankouts. variable pixel clock is not a feature vrr uses.

the fixed base refreshrate should be higher then any core refreshrate will use, then the vbi can dynamically adjust to match the cores refreshrate.

birdybro:
This only tells the tv to enable VRR it doesn't do anything different with blanking intervals.

elexor:
> This only tells the tv to enable VRR it doesn't do anything different with blanking intervals. Some TV's appear to be more resilient to non-standard refresh rates when VRR is enabled in terms of sync stability. They also appear to skip/duplicate less frames according to Ace when in this mode.
> 
> Sync drops will still happen if you switch the video_mode, yes. It's not the most advanced HDMI chip on there and this (among other things) kinda pushes it beyond what it was made for:
> 
> https://www.analog.com/media/en/technical-documentation/data-sheets/adv7513.pdf
> 
> It's awesome it still works well regardless.

Thanks for the reply, understood.

Also just occurred to me that vrr vbi adjustment alone wouldn't give you enough granularity to hit precise fractional refresh rate values that these cores require.

elexor:
looks like sorg is adding variable vbi to cores already? nice

birdybro:
Yup, it's happening!

elexor:
I noticed he's using vsync adjust 0 when vrr is active does that mean it's adding a frame of lag?

sorgelig:
I don't have true VRR monitor. I'm testing currently on iPad3 LCD with board supporting VRR at some extent. This board doesn't care of HDMI config written by @zakk4223, but it still uses some data read from EDID. Original place of HDMI VRR init was incorrect, so i had to move it to video setup place. I'm not sure if it affects VRR monitors. So it needs further test and tweaks in Main.

atrac17:
Here's a third for VRR and vsync_adjust=2 setting as well.

sorgelig:
i've already pushed the modification, so when vsync_adjust=2 is used then VRR logic won't be used (only VRR config will be set).

sorgelig:
> The code as it currently stands disables freesync/vrr on most displays. set_vrr_mode() either needs to be called after hdmi_config(), or hdmi_config() needs to not modify register 0x40. That register controls the sending of the SPD/Spare hdmi packets, and when set to zero they are no longer sent which most displays will disable vrr when that happens.

better if you will modify this part and verify how it works. I don't have VRR monitor to test it.

zakk4223:
I've been doing some testing with various devices and some cores built with the scaler changes:

The current 'adjust pixel clock' method is unworkable with many modern displays. The advertised max VRR refresh rate pushes the pixel clock well beyond the mister's capabilities.

The 'reduce vblank' method works very well. However I have an HDMI switcher that I've verified works with other freesync devices (including working flawlessly with framerate changes) that refused to pass through any video when VRR capable cores are active. 
If I modify the scaler to ignore any 1 line VBI changes it works. I assume there is some constant shifting of the sync that prevents it from locking onto it.

It still drops signal briefly on large framerate changes (like Genesis EU<->US). Looking at the linux AMD drivers there's effort there to prevent large changes in VBI. Changes are ramped up/down at about 1hz/frame (based on 60hz I think). I will try a crude ramping implementation to see if that helps.

sorgelig:
> The current 'adjust pixel clock' method is unworkable with many modern displays. The advertised max VRR refresh rate pushes the pixel clock well beyond the mister's capabilities.

it needs a manual max refresh rate setup. As a workaround max refresh can be limited in Main by ~63Hz unless higher value is manually set.
Reduced vblank often won't give much higher refresh than base refresh because higher resolutions usually have very narrow vblank, so this version isn't so useful. It will mostly require a manual video mode setup with max refresh. That's why adjusting pixel clock is more viable version.

As for ramping VBlank - it will require quite huge code which may affect scaler performance, so overclocked resolutions such as 2560x1440 or 2048x1536 won't work anymore. Or core will fail to build without slacks.

sorgelig:
So, it seems Freesync unlike the meaning is not really free sync.. Another advertisement lie...

PR #655, opened 2022-07-09
fix incorrect value for ioctl_index for more than 4 file extensions

eriks5:
`ioctl_index` is 16 bits, but when loading or mounting images from the menu (`F` and `S` conf_str options), `ioctl_index` is a *char* causing all but the lower two bits of the selected extension index to be lost. 

This pull request makes the `F` and `S` options use the all 16 bits, making it possible to distinguish more than 4 file extensions in the core.

### DIFF
--- a/menu.cpp
+++ b/menu.cpp
@@ -2216,7 +2216,7 @@ void HandleUI(void)
 			if (selPath[0])
 			{
 
-				char idx = user_io_ext_idx(selPath, fs_pFileExt) << 6 | ioctl_index;
+				uint16_t idx = user_io_ext_idx(selPath, fs_pFileExt) << 6 | ioctl_index;
 				if (addon[0] == 'f' && addon[1] != '1') process_addon(addon, idx);
 
 				if (fs_Options & SCANO_NEOGEO)
--- a/menu.cpp
+++ b/menu.cpp
@@ -2287,7 +2287,7 @@ void HandleUI(void)
 			}
 			else
 			{
-				user_io_set_index(user_io_ext_idx(selPath, fs_pFileExt) << 6 | (menusub + 1));
+				user_io_set_aindex(user_io_ext_idx(selPath, fs_pFileExt) << 6 | (menusub + 1));
 				user_io_file_mount(selPath, ioctl_index);
 			}
 
--- a/user_io.cpp
+++ b/user_io.cpp
@@ -2364,7 +2364,7 @@ int user_io_file_tx_a(const char* name, uint16_t index)
 	return 1;
 }
 
-int user_io_file_tx(const char* name, unsigned char index, char opensave, char mute, char composite, uint32_t load_addr)
+int user_io_file_tx(const char* name, uint16_t index, char opensave, char mute, char composite, uint32_t load_addr)
 {
 	fileTYPE f = {};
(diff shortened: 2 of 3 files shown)


sorgelig:
There are many places where it's limited to 4 extensions, so your change may not backward compatible.
4 extensions are enough according to already existing cores. There is also wildcards and at the end if really more than 4 extensions are needed then another F/S option can be used. 
Screen is also limits display of 4 extensions.
So it's better to keep this way.

eriks5:
Any chance to discuss this?

I considered backwards compatibility. But since as you say currently no core needs more than 4 extensions backwards compatibility didn't seem an issue.

The C128 core I'm working on needs 2 more disk images for the `S0` and `S1` disk mount options, as the C128 needs to support the D71 and G71 disk formats along with the existing D64, G64, T64 and D81 formats the C64 core already supports. Like the C64 core, it needs to know if it is a 3.5" disk to activate the 1581 (D81 extension) or a 5.25" disk for the 1541/1571 drive (anything else).

I'm sure there's a workaround possible in the core but using the extra available bits in `iotcl_index`  seemed the better approach to me. The [hps_io](https://github.com/MiSTer-devel/Main_MiSTer/wiki/SYS---HPS-IO#rom-and-file-loading-nvram-saving) documentation doesn't specify that it's limited to 4 extensions. That actually states what this pull request will provide:
> ioctl_index - [15:6] which file type [5:0] file number

As a workaround I already tried using `ioctl_file_ext` to detect a D81 file type, but that variable is not set for the `S` option, only the `F` option, so that didn't work out.

birdybro:
Correct there is another way to do this, check out ao486, you can mount 4 IDE devices in addition to two floppy images simultaneously. `S0` and `S1` are used for floppy images (.img, .ima. .vfd), `S2` and `S3` are used for VHD's (IDE 0-0 and IDE 0-1), and `S4` and `S5` are used for either VHDs or for CD images (.chd, .cue, .iso).

https://github.com/MiSTer-devel/Main_MiSTer/blob/master/support/x86/x86.cpp

It's because a support file was written here to handle the exceptions. The floppy exceptions are handled here as well. If you are planning on forking the core into MiSTer-Devel I'm sure something could be worked out to add an equivalent support file. Maybe the core could be forked into the org first if it's already mostly up and running, then this support be added?

eriks5:
Thanks for thinking along! I already have a workaround in place. Since I only need to know if one of the six file extensions possible is selected, I moved that to a position that has a unique value in `ioctl_index[7:6]`.

The reason I don't want to add additional menu items or more complexity in Main_MiSTer is because I want to keep this core as close to the C64 it is forked off as possible. Because it gives a better user experience if works similar to the C64, and also less chance of merge conflicts when pulling in future updates to that core.

birdybro:
Makes sense, also the less reliance on Main support files the more portable the core is, which is always a good idea moving forward. :)

sorgelig:
Besides the index, hps_io transmits file extension. You can check Amstrad core how to use it.
After all, C64 uses special handling for mounted images. Cores with special handling in hps_io may have their own customization. So if you will re-use C64 disk code, then hps can send specific index bits to reflect a specific disk type.

PR #666, opened 2022-07-18
Upstream video optimization

wickerwaka:
Reduce cost of setting video mode information when using
vsync_adjust=1/2 or vscale_mode=4/5 from 90ms to 5ms (worst case, most
instances are less than 1ms).

Split video initialization into video_init and video_set_mode.
video_init is called once, video_set_mode is called whenever the mode
changes.

Split hdmi_config into hdmi_config_init and hdmi_config_set_mode. Same
as video_, hdmi_config_init does the bulk of the initialization,
hdmi_config_set_mode is just for parameters that can change based on the
mode.

Load video filter data in loadScalerCfg and persist it.

Calculate a digest for scaler data and use that to determine whether new
data needs to be sent.

Only send gamma information if the filename has changed.

Offload fb module parameter writing to a separate thread via the new
offload system.

Reduce the amount of work being done in set_vrr_mode when vrr is
disabled.

Added a spike profiler to help track down these kinds of issues in the future. Disabled by default.


video.cpp
```diff
@@ -382,29 +406,50 @@ static bool read_video_filter(int type, VideoFilter *out)
...
+		// Make a default NN filter in case of error
+		out->is_adaptive = false;
+		FilterPhase nn_phases[2] =
+		{
+			{ .t = { 0, 256, 0, 0 } },
+			{ .t = { 0, 0, 256, 0 } }
+		};
+		scale_phases(out->phases, nn_phases, 2);
+		valid = false;
+	}
```
boogermann, line 437: this part of the code is unreachable.
Is that supposed to kick in when is_adaptive and valid is false?
I would suggest the following refactoring:
```c
bool valid;
if(is_adaptive) {
	out->is_adaptive = true;
	valid = scale_phases(out->phases, phases, count / 2);
	valid = valid && scale_phases(out->adaptive_phases, phases + (count / 2), count / 2);
} else {
	out->is_adaptive = false;
	valid = scale_phases(out->phases, phases, count == 32 ? 16 : count);
	if(!valid){
		// Make a default NN filter in case of error
		FilterPhase nn_phases[2] = {
			{.t = {0, 256, 0, 0}},
			{.t = {0, 0, 256, 0}}
		};
		scale_phases(out->phases, nn_phases, 2);
	}
}
```

boogermann (commented):
Make a default NN filter in case of error is unreachable

video.cpp
wickerwaka: Thanks, I'll fix it up. It is meant to be a fallback for when parsing fails for any reason.

PR #677, opened 2022-08-11
the clean target should not remove gcc files

kdridi:


sorgelig:
i didn't get it. it seems something related to your specific config.

kdridi:
In the build process, before using make, you should setup the dev env by lanching this command "source setup_default_toolchain.sh".

This command fetch the compilation toolchain and install it.

The "make clean" target erase some of the file of the installed toolchain, that's why i did some modification in the Makefile to prevent this deletions by excluding the toolchain folder using the path of $(BASE)-gcc

sorgelig:
your PR breaks "make clean". It stops removing object (.o) files. 
So I've rolled it back.

Issue #669, opened 2022-07-30
[Bug] HDMI output flashing since 20220517 release


dshadoff:
After updating for the first time since May 11 2022, my HDMI output does not work correctly; it displays for a few seconds, then goes blank for a few seconds, then displays for a few seconds.

Testing the various Main_MiSTer releases shows that releases up to May 11, 2022 have no issues; this issue starts with the May 17, 2022 release.

My HDMI output goes from the MiSTer into an AverMedia EzRecorder 130 capture device (turned off, so should be passthru), then into a 1280x720 HDMI monitor.

If I remove the capture device (connecting MiSTer directly to the monitor), the flashing no longer takes place - but I cannot record video while playing.

I am attaching my boot logs, in case they are helpful.

[test.zip](https://github.com/MiSTer-devel/Main_MiSTer/files/9227523/test.zip)

dshadoff:
After rebuilding code at several commit-points between May 11 and May 17, I have identified the commit which introduced the issue.
"Release 20220511" rebuilds cleanly and does not have the issue
However, the very next commit:
"video: configure HDMI from HPS" (from later on May 11) is the first build which has the issue.

birdybro:
Have you tried adding `hdmi_game_mode=1` to your MiSTer.ini to see if there is any change? Does it go away when you change to `vsync_adjust=0`? (not saying you should permanently change to that, just curious).

dshadoff:
Neither of those help.

sorgelig:
This issue can be fixed only by the one who has this problem. This is not the issue which can be reproduced by others...

dshadoff:
Working with Martin, this has been narrowed down to the setting of register 0x59 in video.cpp .
(More information when available).

atrac17:
> Working with Martin, this has been narrowed down to the setting of register 0x59 in video.cpp . (More information when available).

Great, this has plagued me for weeks. Good work guys.

wickerwaka:
> > Working with Martin, this has been narrowed down to the setting of register 0x59 in video.cpp . (More information when available).
> 
> Great, this has plagued me for weeks. Good work guys.

Same capture card or a different setup?

atrac17:
Different setup

dshadoff:
Fixed in 08/19 release

PR #678, opened 2022-08-15
Don't mess with YCC quantization bits

wickerwaka:
Some capture cards don't see to like it. Standard says it should match the RGB quantization, but also says sinks should ignore it.


dshadoff:
This fixes my problem in issue #669

PR #679, opened 2022-08-16
Add pcxt project

spark2k06:
Initially, with serdrive as HDD/FDD image server, only active when starting the PCXT core, never on other cores.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,6 +24,7 @@ INCLUDE += -I./lib/libchdr/include
 INCLUDE += -I./lib/flac/include
 INCLUDE += -I./lib/flac/src/include
 INCLUDE += -I./lib/bluetooth
+INCLUDE += -I./lib/serial_server/library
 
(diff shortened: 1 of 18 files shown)


spark2k06:
Thank you very much for accepting the pull request. I have prepared the final repository that takes advantage of these changes:

https://github.com/spark2k06/PCXT_MiSTer

How would be the procedure to move it to the official MiSTer repository? It would be nice to sync the official Main update with this core.

sorgelig:
I've invited you into MiSTer-devel group. After accepting, you can create PCXT_MiSTer repository and it will become official.
I suggest to push changes directly to MiSTer-devel/PCXT_MiSTer repository instead of PR from your repo, so it will have a better commits history.

spark2k06:
All right, then I'll accept the invitation. Once you have made the Main changes public I will create the project and announce it in misterpga's Changelog thread, as this core depends on the Main.

PR #680, opened 2022-08-16
Add support for sdl gamecontrollerdb to setup default mappings for controllers

zakk4223:
When a controller has no configured mapping, try to find it in an sdl gamecontrollerdb.txt style entry. 
As a start, we can distribute the existing gamecontrollerdb.txt from https://github.com/gabomdq/SDL_GameControllerDB

### DIFF
--- a/gamecontroller_db.cpp
+++ b/gamecontroller_db.cpp
@@ -1,0 +1,354 @@
+#include "gamecontroller_db.h"
+#include <stdio.h>
+#include <string.h>
+#include <stdlib.h>
+#include "input.h"
+#include "file_io.h"
+#include "user_io.h"
+#include "profiling.h"
+
+
+
+
+static const char *sdlname_to_mister_idx[] = {
+
+	"dpright",
+	"dpleft",
+	"dpdown",
+	"dpup",
+	"a",
+	"b",
+	"x",
+	"y",
+	"leftshoulder",
+	"rightshoulder",
+	"back",
+	"start",
+	"---",
+	"---",
+	"---",
+	"---",
+	"---",
+	"---",
+	"---",
+	"---",
+	"---", //20
+	"guide", //21
+	"---",
+	"---",
+	"leftx",
+	"lefty",
+	"rightx",
+	"righty",
+};
+
+
+typedef struct {
+
+	char guid[256];
+	//char name[256]; Do we really care?
+	char mapping_str[1024];
+	bool map_parsed;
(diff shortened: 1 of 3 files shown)


sorgelig:
Do i understand right you reserve 16MB+ of memory for this maping?? 
Also loading/re-loading the whole database with every controller plug/unplug doesn't look good.
While it gives some subtle convenience (but user will always find controller not in the database which will defeat the whole purpose) it looks quite heavy operation. We aren't on i7 8-core CPU.

zakk4223:
It doesn't reload the whole database on every controller plug/unplug. It only does that in input_test() when state==0 and that's only possible once per execution of main binary. The only thing done on controller plugin is to lookup the controller guid in the already loaded database.

You are right about the memory. I initially deferred parsing of the map strings until they were needed (when a matching controller is attached) but that's too much memory overhead to save some initial parsing work. I can reduce each entry to ~140 bytes I think (maybe less...), and start with a smaller database size and only grow if for some reason there are more entries in the file.

sorgelig:
I suggest to supply controller ID to parser, so it won't keep whole controller database and won't require to keep useless data in memory. It also should cache config for already parsed ID, so re-plug (or plug another controller with same ID) won't need to parse database again. So it will imply that user have only one unknown controller (several controllers of same model will be considered as one controller).

sorgelig:
with cashed controller ID you will need to keep up to NUMPLAYERS maps in memory (in worst case when all controllers are different).

zakk4223:
Updated commit searches the db files for the controller as-needed instead of loading all the entries upfront. Much smaller memory footprint now ;)

gamecontroller_db.cpp
```diff
@@ -10,16 +11,16 @@
...
 	"dpright",
 	"dpleft",
 	"dpdown",
 	"dpup",
-	"a",
 	"b",
-	"x",
+	"a",
 	"y",
+	"x",
```
boogermann, line 23: won't this change have the buttons swapped since you're using the array to map?
`SYS_BTN_A` is `[4]`, `SYS_BTN_B` is `[5]`, `SYS_BTN_X` is `[6]` and `SYS_BTN_Y` is `[7]`

zakk4223: Yes, and that is intended. SDL does a/b/x/y in the xbox layout style (a is south, b is east) . Mister's default/"main" uses the SNES/nintendo virtual gamepad style where a is east, b is south.

PR #684, opened 2022-08-22
Support for custom BIOS in PCXT core

spark2k06:
The optional custom ROM loading needs to be independent of model-specific ROM loading during core start-up. For this purpose, a status bit selectable from the OSD menu shall be used.

image

### DIFF
--- a/menu.cpp
+++ b/menu.cpp
@@ -2182,7 +2182,7 @@ void HandleUI(void)
 
 									if (is_pce() && !bit) pcecd_reset();
 									if (is_saturn() && !bit) saturn_reset();
-									if (is_pcxt() && !bit) pcxt_init();
+									if (is_pcxt() && !bit) pcxt_init(!(bool)user_io_status_get("[23]"));
 
 									user_io_status_set(opt, 1, ex);
 									user_io_status_set(opt, 0, ex);
--- a/support/pcxt/pcxt.cpp
+++ b/support/pcxt/pcxt.cpp
@@ -40,12 +40,16 @@ typedef struct
 
 static pcxt_config config;
 
-void pcxt_init()
+void pcxt_init(bool rom_by_model)
 {	
 	user_io_status_set("[0]", 1);
+	if (rom_by_model) pcxt_model_rom();
+}
+
+void pcxt_model_rom()
+{
 	const char* home = HomeDir();
 	static char mainpath[512];
-
 	int status = user_io_status_get("[3]");
 	if (status)
 	{
@@ -54,8 +58,7 @@ void pcxt_init()
 	else
 	{
 		sprintf(mainpath, "%s/pcxt.rom", home);
-	}	
-	
+	}
 	user_io_file_tx(mainpath);
 }
 
--- a/support/pcxt/pcxt.h
+++ b/support/pcxt/pcxt.h
@@ -3,7 +3,8 @@
 
 #include "../../file_io.h"
 
-void pcxt_init(void);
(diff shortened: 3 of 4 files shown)


sorgelig:
I don't think development is going to a right way. You need to minimize customization in Main. Especially don't relay on specific bits in status. MiSTer framework supports ROM loading. You may use FC option to load and remember the ROM. It will be automatically loaded on core load.

spark2k06:
Well, let me explain the need and then tell me what would be the best solution. Initially, as you can see in the picture, I chose to use the model option to choose between two machines, IBM PCXT and Tandy 1000. 

This option has two functions, on the one hand, in the Main you select a rom according to the model, which must be "pcxt.rom" or "tandy.rom", which facilitates the task of ROM selection for a basic user, once he has placed these roms in the directory. The loading of this ROM is carried out, both in the first load and in each reset carried out by the user, always taking into account the selected model.

On the other hand, depending on the selected model, the core uses this bit to enable Tandy/CGA graphics or only CGA, as well as the type of keyboard compatible with the BIOS corresponding to the model.

This new proposal aims to be able to modify the BIOS during boot, and when rebooting, and only at this moment, the selected BIOS is loaded... the use in principle that would be given would be temporary and with a scope of development or testing of new BIOS that could then be used as main, or diagnostic ROMs, etc... that is to say, it would be a use in principle for more advanced users. The problem is that during the reset, the selected model should not interfere in this case... that's why I have added the option "Ignore Model on Reset", so the ROM according to the selected model would only apply to the initial boot.

And now, according to your proposal, although it might make it a little less intuitive to remove the model selection, I have the following doubts. How could I make the core act as Tandy or as PCXT depending on the selected ROM? It occurs to me that the name could start with "pcxt_" and "tandy_"... but, is this text accessible from the core?

birdybro:
I believe this may work: You could inform the user to rename the boot0.rom (synonym for boot.rom) for the PCXT bios and boot1.rom for the Tandy bios. Then upon status[3] (your system toggle), do another ioctl_download and reset with the correct boot*X*.rom. Most people will use just one bios for each type, and power users are used to renaming boot rom files, at least on consoles, when testing things.

From the development docs (not sure if this is totally accurate however):

https://mister-devel.github.io/MkDocs_MiSTer/developer/hps_io/#rom-and-file-loading-nvram-saving

`Note: boot.rom is sent via ioctl_index == 0, boot1.rom is sent with [5:0] set to 0, and [15:6] set to 1. boot2.rom.. etc`

spark2k06:
If an advanced user uses the same names then it can be left as it is right now. Instead of boot0.rom and boot1.rom it is pcxt.com and tandy.rom, the advanced user only has to replace the contents of these... but in the long run, it is a more time consuming and cumbersome process than selecting each ROM from the OSD menu.

sorgelig:
boot0.rom .. boot3.rom are kind of obsolete if core allows to load a custom roms. FC option should be used instead, so selected ROM will be remembered and automatically reloaded on core load. However if core uses fixed ROM which can be simply placed and forgot, then boot0.rom .. boot3.rom could be used to avoid selection in OSD.

As for PCXT you can use option like FC0 to load BIOS for IBM mode, and FC1 to load BIOS for Tandy. Core need to reserve the memory for both BIOSes and switch between then when specific model is selected. This is how it's done for other computer cores.

spark2k06:
Now I understand the mechanism better, thank you. I will try to make a good solution based on your proposal. Later on I will make another pull request doing a code cleanup, because that would no longer be necessary under this approach.

PR #686, opened 2022-08-25
Cleanup initialization PCXT

spark2k06:
It is no longer necessary to resort to loading predefined ROMs, but now FC0 and FC1 are used to load ROMs.


spark2k06:
Can you do a revert of "FDD support, based on AO486"? I'm still testing it from my fork, in fact, I thought that in the pull request I made to you there was only the "Cleanup initialization PCXT" commit, which is the one I'm interested in updating for the moment in the official Main

Thanks!

PR #691, opened 2022-09-07
Add support for [arcade_vertical] section

wickerwaka:
A lot of people maintain large lists of per-game ini sections so they can configure video modes, scaling options and some other things for arcade games that have a vertical orientation. There is a `<rotation>` tag in the mra files that specifies the orientation of each game. This change adds an `[arcade_vertical]` ini section for games that have their rotation tag set to `vertical`.

Not all games have the rotation tag set correctly. There hasn't really been much reason to do it, but maybe this change will encourage people to use it more consistently.

### DIFF
--- a/cfg.cpp
+++ b/cfg.cpp
@@ -12,6 +12,7 @@
 #include "file_io.h"
 #include "user_io.h"
 #include "video.h"
+#include "support/arcade/mra_loader.h"
 
 cfg_t cfg;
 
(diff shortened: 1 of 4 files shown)


sorgelig:
I don't get it. What it will give? In current version you can adjust video mode per core or group of cores. In practice each user basically choose which rotation is used per core and set specific video mode (if required). Other rotation is basically almost never used.

wickerwaka:
Setting a specific video mode is the main use case, setting default filters would be another. For myself, I don't physically rotate my screen when playing vertical arcade cores, so I want to use a mode with the highest possible vertical resolution. So I can do
```
[arcade_vertical]
video_mode=1536,2048,60
```

sorgelig:
I prefer not to multiply variables.. Usually each core is used only one way be specific user. So it's very little use such separation, but it will require to multiply possible cases in the code making it more and more complex because one or two people will use it.

birdybro:
Just to chime in, this will also help me and a few others that have brought it up in discord because `vscale_mode=4` combined with `video_mode=13` isn't as compatible with vertical cores on some displays the same way Waka uses them, so I have to have an `MiSTer_alt_1.ini` with `vscale_mode=5` that I switch to for those games. If I use `vscale_mode=4` and `video_mode=13` then it is horizontally stretched because the LG display gets confused, but I prefer `vscale_mode=4` for all other cores since I want to always maintain the core's aspect ratio while filling vertically.

thehughhefner:
I also want to chime in and say that I’ve tested this feature and it is working really well. With the change wickerwaka has added, I don’t need to switch between ini files. I hope that this can be merged

Toryalai1:
I have been testing this with multiple vertical arcade games and it is working really good. Now I don’t need to swap ini files. I hope this get merged

Issue #693, opened 2022-09-10
OSD message from command line


pocomane:
I got a request for mbc, asking to [show some message on screen by the command line](https://github.com/pocomane/MiSTer_Batch_Control/issues/28). Is there some way to do this? Like a command to send to /dev/MiSTer_cmd, or a fake mra/mgl that have the only effect of showing a message?

If there are no way, are there any chance to add it to Main_MiSTer ?

sorgelig:
Currently there is no way to display custom message. 
I will think about it. It's not easy to implement because messages cannot be queued, so other message from core (or opposite) can hide previous message making a race condition.
Generally speaking OSD message is exceptional way to signal the user.

pocomane:
Just for completeness, if this feature were to be implemented, other ones like #634 could be delegated to external tools.

Issue #451, opened 2021-09-03
Feature request: ability to change the background per core


gitdann:
right now we can change the background of the default mister by putting the jpg in the root of the sd card, if its possible could we have the ability to change the background image by dropping the jpg in the root of each core? that way as we scroll cores the background would change.

thank you.

sorgelig:
It's time consuming operation. It will freeze the system while processing the picture.

PR #704, opened 2022-10-26
[Menu Feature] Add command `load_bg` to change the MiSTer menu background on-demand.

Natrox:
Hello folks!

I have worked on a little feature to change the background of the menu on-demand. In essence, for the end-user, they can send off the following command:
`echo load_bg <image> >/dev/MiSTer_cmd`

The new code will take care of asynchronous loading of the new background image, swapping it in (while respecting screen dim), and making sure the core itself is not stalled. It has been tested extensively for stability and I have made sure it does not block input or anything else. It does not touch any file data - thus it does not mess with the existing wallpaper image.

If the wallpaper is not active (e.g. user has noise as wallpaper), it will not override the user's choice.

**Why?**
It's a big ticket request, a lot of users would like to change their wallpaper on-the-fly. This provides that option as well as much more. Using this code, we can give users the ability to change the wallpaper of their menu. However, the scripting team can do a lot more with this - some examples:
- Show random screenshots of games played on a timer (see video below).
- Add iconography to the wallpaper regarding controller battery status, connected controllers, etc.
- Create basic animated wallpapers (low FPS but fun).
- Show information about whether there are any updates available.
- Use a wallpaper for each different core.

The possibilities are quite endless!

I have a small proof-of-concept video showing how it can be used, as well as some code how to do it (which I can pull up):
[](https://www.youtube.com/watch?v=bHshkmax_ts)

**Issues affected**
This provides functionality to address (from what I can find):
#451 
#693 (through image rendering)

**What's next?**
If this functionality can make it into the main distribution, we can start doing work on scripting to provide all kinds of functionality to the user. I am working with @wizzomafizzo on functionality making use of the feature. Of course, other users/scripters may do their own thing, which I think is exciting.

Thank you for all your hard work so far, and thank you for your consideration! If there are any issues with the code as-is, please let me know and I'll adjust it. I can also provide a binary for testing purposes if anyone would like it (or you can compile my branch).


video.cpp
```diff
@@ -3193,6 +3227,30 @@ void video_cmd(char *cmd)
...
+		static char fname[1024];
+		sprintf(fname, "%s", bg);
+
+		// We need to do the load async because we're currently
+		// on the input thread - if we don't use offload we will
+		// stall the core's input processing
+		offload_add_work([=]
+		{
+			load_bg_specific(fname);
+			video_menu_bg(user_io_status_get("[3:1]")); // force redraw
```
wickerwaka, line 3251: It is not safe to call video_menu_bg in a separate thread. video_menu_bg makes user io spi calls, if two of those happened from different threads at the same time they would collide with each other and have unpredictable results. I haven't looked at what other global state video_menu_bg might interact with.

Natrox: Fair point, however I could not find any other way to force a call of video_menu_bg. The chance of this happening is low but not impossible - would adding a simple guard suffice (e.g. with a bool that is set upon function entry)? I can amend it to guard against this.

birdybro:
Suggestion: Convert this to a draft PR for now, work out the specific issues raised to make it ready to merge, once it's ready to merge squash your commits into a single commit to clean up the history when merged.

video.cpp
Natrox: Alright, as discussed, I'm going to make sure any SPI stuff is moved outside of the offload - change incoming soon. Afterwards, for maintenance sake I think `video_menu_bg` is a good candidate for refactoring, I can spearhead that effort if you agree.

Natrox:
> Suggestion: Convert this to a draft PR for now, work out the specific issues raised to make it ready to merge, once it's ready to merge squash your commits into a single commit to clean up the history if/when merged.

Thank you. Agreed, I'll convert it for the time being and work on the issues raised - good one on the squashing I keep forgetting haha.

sorgelig:
Generally speaking it's not a good idea to re-load background. It's quite slow. It will be sluggish if background gets reloaded while user interacting menu. It's also not well tested against memory leak.
Just understand it's not RPi. You won't achieve the same as there!

Natrox:
Cleanup of commit history, and new version. As discussed, I have made sure that no SPI calls are made in the offload thread. Everything still runs without hitching, and functionally nothing has changed.

I have also slightly bolstered thread safety and fixed a small bug regarding image reloads while the screen is idle; first PR would not respect a screen dim. 

A quick summary of what this change does and does not do:
- It exposes a new command, `load_bg`, to change the wallpaper.
- Running this command causes an image load to happen on the existing offload thread - meaning there is no input blockage.
- The feature is purely optional and does not affect any existing code paths (beyond changes that poll for swapping).
- Importantly, the image loaded through this command is not cached - therefore, memory leaks do not occur. I have left this running overnight with the wallpaper switching to a random one every few seconds with no issue.
- Everything is thread-safe. I've slightly changed the `offload.h/cpp` (see [diff of ofload.cpp](https://github.com/Natrox/Main_MiSTer/commit/97d2d00864445cd9dd1957ac19bafcfe1df63986#diff-766e75ac687d1773fe9bcebfd8b40b921a82ce1968e689ef7922292c50b5bdae))  and given it a parameter to "bail out" if the queue is full. This prevents any spamming as well as potential hangs.


Ultimately, the decision to accept or reject this PR is up to you of course. I worked on this change as it's something that is often requested, and because it was said to be impossible without stalling the input thread. I have engineered all this to both add a cool new feature that scripts can use to do all kinds of stuff - as well as to give myself a challenge.

It is not a Raspberry Pi, but if you coax it a little it can still do a lot of things in a user-friendly way. :relaxed:

As always, I am open to any feedback or concerns and all that. If this does not make it in, I will maintain it in a fork for the die-hard stakeholders.

Natrox:
Closing stale pull request.
Archived in personal branch `dynamic-background`.

PR #705, opened 2022-10-30
Allow setting controller assignment

fitzsim:
For MiSTercade at least, it can be desirable to disable dynamic
controller assignment, and instead set controllers to particular
devices.

For example, on a cabinet with left and right controls, even if the
second player mashes some buttons first, the operator may still want
the left controls to be for player 1, and the right controls for
player 2, as they would have been for the original game.

This patch allows setting up to four controllers to specific devices.

It adds the device index to the controller mapping dialog box.  The
user can first enable dynamic assignment, then experiment with which
controllers correspond to which devices.  For example, they can push a
button on controller P𝑛, note the device index, D𝑚, from the mapping
dialog, then set:

controller_𝑛_device=𝑚

in **mister.ini**.

### DIFF
--- a/MiSTer.ini
+++ b/MiSTer.ini
@@ -121,6 +121,14 @@ refresh_max=0
 ; 0 - disable
 controller_info=6
 
+; Use these settings to override dynamic controller assignment.  The
+; device index is that reported in the controller button map as
+; D<index>.
+; controller_1_device=0
+; controller_2_device=1
+; controller_3_device=2
+; controller_4_device=3
+
 ; JammaSD/J-PAC/I-PAC keys to joysticks translation
 ; You have to provide correct VID and PID of your input device
 ; Examples: Legacy J-PAC with Mini-USB or USB capable I-PAC with PS/2 connectors VID=0xD209/PID=0x0301
--- a/cfg.cpp
+++ b/cfg.cpp
@@ -67,6 +67,10 @@ static const ini_var_t ini_vars[] =
 	{ "GAMEPAD_DEFAULTS", (void*)(&(cfg.gamepad_defaults)), UINT8, 0, 1 },
 	{ "RECENTS", (void*)(&(cfg.recents)), UINT8, 0, 1 },
 	{ "CONTROLLER_INFO", (void*)(&(cfg.controller_info)), UINT8, 0, 10 },
+	{ "CONTROLLER_1_DEVICE", (void*)(&(cfg.controller_device[0])), UINT8, 0, 255 },
+	{ "CONTROLLER_2_DEVICE", (void*)(&(cfg.controller_device[1])), UINT8, 0, 255 },
+	{ "CONTROLLER_3_DEVICE", (void*)(&(cfg.controller_device[2])), UINT8, 0, 255 },
+	{ "CONTROLLER_4_DEVICE", (void*)(&(cfg.controller_device[3])), UINT8, 0, 255 },
 	{ "REFRESH_MIN", (void*)(&(cfg.refresh_min)), FLOAT, 0, 150 },
 	{ "REFRESH_MAX", (void*)(&(cfg.refresh_max)), FLOAT, 0, 150 },
 	{ "JAMMA_VID", (void*)(&(cfg.jamma_vid)), UINT16, 0, 0xFFFF },
@@ -416,6 +420,10 @@ void cfg_parse()
 	cfg.bootscreen = 1;
 	cfg.fb_terminal = 1;
 	cfg.controller_info = 6;
+	for (int i = 0; i < 4; i++)
+	{
+		cfg.controller_device[i] = 255;
+	}
 	cfg.browse_expand = 1;
 	cfg.logo = 1;
 	cfg.rumble = 1;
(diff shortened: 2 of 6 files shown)


sorgelig:
This patch assumes USB enumeration is always the same which is not always true.

wickerwaka:
I think in the case of MiSTercade the enumeration order can be relied on, but it's not a very general solution. What can be relied upon is the physical path of the device, which represents its location within the USB topology. This is the `Phys` info in `/proc/bus/input/devices` or the result from the `EVIOCGPHYS` ioctl. This would allow you to specify that you always want the device that is connected to a specific USB port to be assigned to a specific player.

However, the format is a little unwieldy: `usb-ffb40000.usb-1.6/input0` for instance. Not sure how that could be expressed in a way that would be a little bit more user-friendly for placing in the ini. Maybe that doesn't matter since it is definitely a "power user" option for more permanent setups.

I would use this myself for a case I'm building that is going to just have two USB ports on the front, so it's not just an arcade thing.

fitzsim:
For my MiSTercade setup the enumeration seems reliable (i.e., the
patch has been working for me for about a month).  Does Linux always
assign the same device indices if the same set of USB devices are
connected?  I guess there could be race conditions during USB
device probing.

I agree though that it should be made reliable.  I'll think about
where the path could be displayed in the UI.  Maybe another
configuration option to include the path in the controller info dialog
box?

sorgelig:
Some arcade cabinets use keyboard as controller. Like JammaSD/J-PAC/I-PAC
They are always assigned as player1/2 regardless which controller is pressed first.
So i suggest to stick to such controllers if you care about correct player assignment.

sorgelig:
Yes, USB path is reliable way to distinguish. I will check what can i do there and may be make some easy way for user to assign it.

sorgelig:
implemented.

fitzsim:
The changes work well on my setup, thank you!

I wasn't sure where the USB debug log was, so I ran this via SSH:

`killall MiSTer; /media/fat/MiSTer`

which produced the following terminal output:

```
opened 0( 0): /dev/input/event1 (16d0:10be) 0 "usb-ffb40000.usb-1.4/input1" "Mister Addons MiSTer JAMMA Board"
opened 1( 1): /dev/input/event0 (16d0:10be) 0 "usb-ffb40000.usb-1.4/input0" "Mister Addons MiSTer JAMMA Board"
```

These `MiSTer.ini` settings worked:

```
player_1_controller=usb-ffb40000.usb-1.4/input0
player_2_controller=usb-ffb40000.usb-1.4/input1
```

Note that the `input0` and `input1` are needed to differentiate the two
inputs, contrary to the comment's recommendation to omit the `inputX`
part.

The MiSTercade entries in `/proc/bus/input/devices` are:

```
I: Bus=0003 Vendor=16d0 Product=10be Version=0101
N: Name="Mister Addons MiSTer JAMMA Board"
P: Phys=usb-ffb40000.usb-1.4/input0
S: Sysfs=/devices/platform/soc/ffb40000.usb/usb1/1-1/1-1.4/1-1.4:1.0/0003:16D0:10BE.0001/input/input0
U: Uniq=
H: Handlers=js0 event0 
B: PROP=0
B: EV=1b
B: KEY=fff 0 0 0 0 0 0 0 0 0
B: ABS=30027
B: MSC=10

I: Bus=0003 Vendor=16d0 Product=10be Version=0101
N: Name="Mister Addons MiSTer JAMMA Board"
P: Phys=usb-ffb40000.usb-1.4/input1
S: Sysfs=/devices/platform/soc/ffb40000.usb/usb1/1-1/1-1.4/1-1.4:1.1/0003:16D0:10BE.0002/input/input1
U: Uniq=
H: Handlers=js1 event1 
B: PROP=0
B: EV=1b
B: KEY=fff 0 0 0 0 0 0 0 0 0
B: ABS=30027
B: MSC=10
```

That's why `input0` and `input1` are reliable for the MiSTercade -- they
represent interface 0 and interface 1 of the USB endpoint, which are
always indexed in the same order by the driver.

MiSTercade may be a special case; I'm not sure if any of the above is
worth summarizing in the `MiSTer.ini` comments.

sorgelig:
I believe there are many special cases depending on used controllers. If you won't plug any additional input devices, then probably inputX will stay the same. In your case probably sysfs would be more reliable to distinguish. Anyway, will see how it will work.
There is no point to mention quirks of specific devices in INI. Better to put it into Wiki.

sorgelig:
> I wasn't sure where the USB debug log was

it's mini-USB connector near Ethernet. It's USB-serial port. But you can use SSH if you start MiSTer binary from there (as you already done).

sorgelig:
i've just extended ID search for sysfs as well. So now you can use this:
```
player_1_controller=16D0:10BE.0001
player_2_controller=16D0:10BE.0002
```
it's stronger bind than eventX.

fitzsim:
Even better, thanks.  I updated the MiSTer binary and mister.ini on my MiSTercade and it's working as expected.

misteraddons:
Just popping in to say thank you to everyone who made this happen :)

PR #707, opened 2022-11-12
Improved handling of Win+F12

spark2k06:
This key is not essential for PCXT and Tandy era software, however some games and demos may intercept it and switch screens or exit to MS/Dos... which makes taking a screenshot not possible, among other things.

### DIFF
--- a/user_io.cpp
+++ b/user_io.cpp
@@ -70,6 +70,7 @@ static char keyboard_leds = 0;
 static bool caps_status = 0;
 static bool num_status = 0;
 static bool scrl_status = 0;
+static bool winkey_pressed = 0;
 
 static uint16_t sdram_cfg = 0;
 
@@ -3395,6 +3396,19 @@ void user_io_poll()
 
 static void send_keycode(unsigned short key, int press)
 {
+	if (is_pcxt())
+	{
+		//WIN+... we override this hotkey in the core.
+		if (key == 125 || key == 126)
+		{
+			winkey_pressed = press;			
+			return;
+		}
+		if (winkey_pressed)
+		{
+			return;
+		}
+	}
 	if (is_minimig())
 	{
 		if (press > 1) return;
--- a/user_io.cpp
+++ b/user_io.cpp
@@ -3806,10 +3806,14 @@ void user_io_kbd(uint16_t key, int press)
 				if (is_menu() && !video_fb_state()) printf("PS2 code(break)%s for core: %d(0x%X)\n", (code & EXT) ? "(ext)" : "", code & 255, code & 255);
 
 				if (key == KEY_MENU) key = KEY_F12;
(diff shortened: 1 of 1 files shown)


spark2k06:
In addition, and for all cores, we only release Win+F12... the same demos and games are also sensitive to key release during OSD interaction.

sorgelig:
but it's already handled here: https://github.com/MiSTer-devel/Main_MiSTer/blob/79137718883252972998db14330861f266dc944c/user_io.cpp#L3804

spark2k06:
Normally it shouldn't have any effect, but there are some game and demo... I'll give you an example, the game Action Fighter, with a splash screen... that is sensitive even to the first E0 scancode of the key Win, and already make the game start... this is a problem.

Not only on PCXT, on ao486 it also happens with some games or demos, possibly few, but it happens. During the menu, not a single scancode should be sent, not a single key release...ideally.

sorgelig:
It's not a good idea to block Win key from cores because it can be still used in core as a key. But if you think core really doesn't need Win key at all, then better to block it inside the core instead.
I found that key de-presses are still forwarded to core while OSD is active. It's done to eliminate a sticky keys. Usually in emulated system, depress has no effect if key was released before, but for ao486 and PCXT it still has effect. I will fix this. 
It is possible that some specific game/app may mis-react to Win key (if you won't block it in core) - it can be left as is. As a quirk. OSD still can be opened by hardware OSD on MiSTer or by controller MENU key.

sorgelig:
correction: i see, there will be a problem to block key if core uses serial PS2 connection.. 
Lemme think about it.

sorgelig:
Now it shouldn't pass keys when OSD is opened.
Still thinking about Win key. I think it's better to leave it as is. May be couple games/app will react on it - shouldn't be a big loss i think.

sorgelig:
hmm.. now Win/Alt are sticking...

spark2k06:
It's really rare that keys other than Win or F12 are pressed when the OSD is going to open, hence my solution for all cores.

On the other hand, I understand that my specific solution to disable the Win key only makes sense in PCXT, in ao486 I know that it may be necessary, that's why I have taken it into account in the solution to this.  Here the problem remains, but only in ao486.

spark2k06:
My opinion about the suppression of the Win key in PCXT, is that it should be in the Main... a real PCXT should never receive scancodes from the keyboards of its time.

spark2k06:
I've downloaded and compiled the sources with the latest changes you've made. Particularly with PCXT at least, they don't solve the problem.

sorgelig:
You have 2 commits. First one can be applied (disable win key for pcxt), while second commit now is conflicting. I close this PR. Please open another PR with win disable only.

spark2k06:
https://github.com/MiSTer-devel/Main_MiSTer/pull/708

PR #715, opened 2022-12-18
Feature: HDMI color controls via tranceiver CSC

Natrox:
This change adds color controls for HDMI output to the settings ini.
The implementation makes use of the ADV7513's color space converter (CSC) to facilitate basic color adjustments. It allows for global modulation of output without touching FPGA fabric.

**Impacted functionality**
This change moves CSC-related HDMI functionality to its own function. Due to the nature of the change, this necessitated a rewrite of the existing color space functionality (HDMI-limited, YPbPr). These matrices have been converted from 12-bit integer to `float`. Output will remain identical after merging in this PR as neutral default values have been set up in `cfg.cpp`. Careful review of pre- and post-change output has been performed to ensure there is no errant change.

**List of ini options**
The default `MiSTer.ini` has been amended to include the related options as well as an explanation with examples. The list of options is as follows:
```
cc_brightness = [0-100] (default: 50)
cc_contrast = [0-100] (default: 50)
cc_saturation = [0-100] (default: 100)
cc_hue = [0-360] (default: 0)
cc_gain_offset = [-2 .. 2 ] x 6 (in form of "Rg, Ro, Gg, Go, Bg, Bo") (default: 1, 0, 1, 0, 1, 0)
```
All defaults are implicit, so users do not need to use these options if they do not wish to.

**Why?**
The primary reason for having these controls is for a forthcoming HDR PR. I have a branch ready with a rewrite of wickerwaka's HDR work. Not all displays play nice with the HDR color space, so these controls will allow users to tweak their output to improve their experience. The HDR implementation is similar to the RetroTink 5x.

Beyond that, it provides users with limited device color controls (e.g. iPad display, some CRTs) the ability to tweak the colors. It may also be desirable to use this as an alternate way to colorize grayscale cores.

**Further info**
The gain/offset features are effectively contrast/brightness controls, except per-channel instead of full-spectrum. They are intended for more advanced users, providing ways to change color balance and alter the white-point. Two examples are included in the ini for users to play with.

--

Thank you for all your hard work!

### DIFF
--- a/MiSTer.ini
+++ b/MiSTer.ini
@@ -124,16 +124,16 @@ refresh_max=0
 ; Each of 6 (mandatory) values in gain/offset can be set to -2 - 2.
 ; These 6 values represent gain and offset in order: Rg,Ro,Gg,Go,Bg,Bo
 ; Example 1: Inverted colors, hue shifted 180 degrees:
-; cc_hue= 180
-;	cc_gain_offset= -1, 1, -1, 1, -1, 1
+; video_hue= 180
+; video_gain_offset= -1, 1, -1, 1, -1, 1
 ; Example 2: Slightly desaturated, warm display
-; cc_saturation= 80
-; cc_gain_offset= 1.5, -0.1, 1.3, -0.15, 0.9, 0.05
-cc_brightness=50
-cc_contrast=50
-cc_saturation=100
-cc_hue=0
-cc_gain_offset=1,0,1,0,1,0
+; video_saturation= 80
+; video_gain_offset= 1.5, -0.1, 1.3, -0.15, 0.9, 0.05
+video_brightness=50
+video_contrast=50
+video_saturation=100
+video_hue=0
+video_gain_offset=1,0,1,0,1,0
 
 ; 1-10 (seconds) to display controller's button map upon first time key press
 ; 0 - disable
--- a/MiSTer.ini
+++ b/MiSTer.ini
@@ -117,6 +117,24 @@ refresh_max=0
 ;video_mode_ntsc=0
 ;video_mode_pal=7
 
+; Provided below are parameters for HDMI color controls.
+; The defaults that are set below will result in an unaltered image.
+; Brightness, contrast and saturation all can be set to 0 - 100.
+; Hue can be set to 0 - 360, observing the HSL color representation.
+; Each of 6 (mandatory) values in gain/offset can be set to -2 - 2.
+; These 6 values represent gain and offset in order: Rg,Ro,Gg,Go,Bg,Bo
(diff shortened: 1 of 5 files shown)


sorgelig:
I still have no idea what is this for. You mentioned iPad display. I'm using iPad display - this is my main MiSTer setup and i have no problem with it at all. I don't understand why brightness/contrast/saturation/hue need to be controlled by MiSTer. TVs and Monitors have such control already. 
I'm quite against making code over-complicated with features never used in last 5 years of MiSTer. And after many monitors and TVs i've used i never had such need. 
Video code is already so complex, so i need some time to find what i need.

sorgelig:
i suggest to change `cc_` to `video_` in ini parameters.

Natrox:
Hey Sorg, thanks for the quick response.

Thanks for the suggestion, I will make that change shortly!

As I have stated, the change is mostly in service of HDR. I would like to put in the PR for HDR as well, but without color controls it does not provide a good output on all displays. In my personal case, I needed to up the contrast, brightness and red gain for a nice picture.

Although HDR is not a necessary feature either, it has been requested often. I have personally been running with it for about a month and it has been great. It's a great way to get additional brightness and contrast out of OLED TVs for instance. Due to the nature of using HDR with games that were not built for it, the TV controls won't suffice - HDR relies heavily on the input signal being 'correct'.

It is somewhat hard to explain, so I have attached a photo compilation.
[Link to compilation image...](https://i.imgur.com/5Xh8Oyz.jpg)

1: No HDR
2: HDR BT2020
3: HDR BT2020 with +10 brightness/contrast, 1.1 red gain

Comparing the non-HDR and HDR photos, you can see the boost in contrast and brightness. It is most notable from the skies in the FF7 photo. However, by default you may notice red appear quite dull, and the image in general is a bit darker. Using the controls, I am able to correct that. It is simply not viable to do this in the TVs picture options without raising black levels as well. Contrast is already at maximum by default (on LGs any way).

The difference can be slight at times (Metroid shows fairly little change). It's highly variable based on game and display. My display does not need much work to look 'right' because it is already quite good - other people's displays may need a lot more tweaking.

.. Beyond HDR, it is hard to justify the change. I mentioned iPad displays as an offhand example, but there may be other cases where one would like to use these controls. There was at least one user on the Discord server whose CRT requires inverted colors for input, so they would be quite happy I suppose. :)

If you prefer, I can close this PR and open a new one with HDR rolled into it. Although it is less granular to do it like that, it saves us some time and also provides most people with an immediate use case.

I do agree that `video.cpp` is quite complex. If you have any suggestions on code (any code, not just mine) that can do with some clean-up or rework, I'd be happy to help with that as well!

PR #722, opened 2023-01-02
(Feature) HLG HDR mode with hdr=3 (Request by Mathieulh)

Natrox:
Hello again, and thank you for merging the previous metadata change.
There is one more change I would like to add, it concerns `hdr=3`.

It is a small change, but when `hdr` is set to 3, it will now use HLG HDR mode.
HLG is a mode designed to be backwards compatible with SDR content.
Mathieulh on Discord mentioned it and tested it out on his own display, it was a big improvement over the other modes.

Since HLG requires no color space conversion, it made sense to use the existing `hdr=3` option. I have amended the INI to make mention of it.

---

Thanks!


asturur:
@Natrox yesterda i tried all the profiles on my OLED tv ( Panasonic ).
Mode BT2020 definitely pull out a lot of luminosity but even reducing contrast is still super contrasted
HLG is not that brillnant but colors look definitely normal.

They are both good and improve scanlines a bit.

PR #730, opened 2023-01-14
Re-ordering of HDR options to make HLG default

Natrox:
Hey all,

This is a small change that aims to simplify the HDR option somewhat.
I have removed BT2020 HDR10 mode and changed HLG to be the default (for `hdr=1`).

I have also made some small changes to the config (uint16 -> uint8 for some options), and I have tweaked the comments in MiSTer.ini to read a bit easier. For the HDR option, I have added explicit recommendation to go with HLG mode.

**Functionality impacted**
HDR with BT2020 has been removed, people who were using it (by setting `hdr=1`) will be on HLG after this update. For those that cannot use HLG due to incompatibility, DCI P3 mode has been retained (`hdr=2`).

**Further considerations**
There has been some talk of getting rid of all HDR options beyond HLG, and getting rid of color controls. I think there is value in keeping these around, but we should monitor the community and see whether they have true value or just end up causing confusion. To that end, I decided the "default" HDR mode should be HLG - in case we decide to remove `hdr=2` and color controls.

It should be safe to make `hdr=1` default in the .ini, since displays that do not support it will display SDR as usual. I have not made this change as I think we should await feedback from users.

---

Thank you again for your hard work, and I wish you a good 2023!

### DIFF
--- a/MiSTer.ini
+++ b/MiSTer.ini
@@ -20,9 +20,8 @@ menu_pal=0             ; 1 - PAL mode for menu core
 hdmi_limited=0         ; 1 - use limited (16..235) color range over HDMI
                        ; 2 - use limited (16..255) color range over HDMI, for VGA converters.
 direct_video=0         ; 1 - enable core video timing over HDMI, use only with VGA converters.
-hdr=0                  ; 1 - enable HDR using the BT2020 color space (faux-HDR, use color controls to tweak).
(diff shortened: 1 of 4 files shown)


sorgelig:
I don't use HDR, but i'm fully agree to simplify it which will also simplify the code.
So if it's fine then make HDR simply on/off if it will be ok.

Issue #734, opened 2023-01-27
[BUG] Direct Video YPbPr Pink Tint After MiSTer_20221124


joshuaryoung:
Previous to and including MiSTer_20221124, my direct video ypbpr to crt setup was displaying a perfect picture. On any build after that the picture has a pink tint and everything looks blow out (see included pic). This does not happen on the VGA output of the analog i/o board however.

Setup: Direct Video -> Moread HDMI to VGA converter (green line and C-Sync lines are combined with resistor and diode) -> Sony Trinitron CRT

MiSTer: 128mb SDRAM module, Digital I/O Board from MiSTer Addons

![Untitled](https://user-images.githubusercontent.com/10542661/215188923-7f248794-0e29-4d17-98e5-d7a3cdfeff34.jpg)

MiSTer.ini
```
[MiSTer]
hdr=0
vga_sog=1
hdmi_game_mode=0
vrr_mode=0
controller_info=3
key_menu_as_rgui=0     ; set to 1 to make the MENU key map to RGUI in Minimig (e.g. for Right Amiga)
forced_scandoubler=0   ; set to 1 to run scandoubler on VGA output always (depends on core).
ypbpr=1                ; set to 1 for YPbPr on VGA output.
composite_sync=1       ; set to 1 for composite sync on HSync signal of VGA output.
vga_scaler=0           ; set to 1 to connect VGA to scaler output.
hdmi_audio_96k=0       ; set to 1 for 96khz/16bit HDMI audio (48khz/16bit otherwise)
keyrah_mode=0x18d80002 ; VIDPID of keyrah for special code translation (0x23418037 for Arduino Micro)
; volumectl=1          ; enable audio volume control by multimedia keys
vscale_mode=1          ; 0 - scale to fit the screen height.
                       ; 1 - use integer scale only.
                       ; 2 - use 0.5 steps of scale.
                       ; 3 - use 0.25 steps of scale.
vscale_border=0        ; set vertical border for TVs cutting the upper/bottom parts of screen (1-399)
bootscreen=1          ; uncomment to disable boot screen of some cores like Minimig.
;mouse_throttle=10     ; 1-100 mouse speed divider. Useful for very sensitive mice
rbf_hide_datecode=0    ; 1 - hides datecodes from rbf file names. Press F2 for quick temporary toggle
menu_pal=0             ; 1 - PAL mode for menu core
hdmi_limited=0         ; 1 - use limited (16..235) color range over HDMI
[...]

PR #735, opened 2023-01-27
(Bugfix) Fix color controls issue affecting YPbPr + direct_video

Natrox:
Hey Sorg,

This one is a bugfix. A problem was reported in #734 regarding YPbPr+direct_video.
There was a pink tint that was introduced by my color controls PR.

Although in principle the color controls can be used with YPbPr+dv, in practice there will be few people using YPbPr+dv and even fewer interested in color controls.

I have isolated YPbPr from the color controls code. It uses the previously working logic and style, but still in the CSC function.

I've tested this with the user who made the report in the MiSTer Discord.

Apologies for the bug, and thank you.

PS:
The diff looks pretty messed up, I changed indentation on a large block of code.
For better reading, please diff with options to ignore indentation.

### DIFF
--- a/video.cpp
+++ b/video.cpp
@@ -1066,12 +1066,6 @@ static void hdmi_config_set_csc()
 	// default color conversion matrices
 	// for the original hexadecimal versions please refer
 	// to the ADV7513 programming guide section 4.3.7
-	float ypbpr_coeffs[] = {
-		0.42944335937f, 1.64038085938f, 1.93017578125f, 0.49389648437f,
-		0.25683593750f, 0.50415039062f, 0.09790039062f, 0.06250f,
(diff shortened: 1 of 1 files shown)


sorgelig:
1) Please re-do PR. adjust git settings to ignore white space changes. Also use the same indentation as original code (tabs before first symbol in line, spaces between symbols).
2) Original YPbPr trick was to pass YPbPr through HDMI-VGA converter as-is, so YPbPr will appear on RGB output. If you change HDMI mode to YPbPr then converter will see it and convert it back to RGB and there won't be YPbPr output after converter.

Natrox:
Hey Sorg, thank you for the merge. Apologies for the formatting errors. It looks like I accidentally let VSCode reformat a selection for me. I will check my settings.

Issue #764, opened 2023-04-02
Mouse input clamping appears to be incorrect leading to overflows when moving quickly


garyjsweet:
I have an ultimarc u-trak trackball along with a homemade rotary encoder based spinner. I've been noticing odd behaviour
when the trackball or spinner are moved very quickly. The cursor appears to move backwards briefly, or stalls for a short period
of time. Slow to medium movements worked just fine.

After some debugging I found that the mouse deltas are being incorrectly clamped to -255 to 255 inside user_io_mouse().

Despite the fact that byte-0 of the mouse data contains effective sign bits for the x & y directions (in bits 4 & 5 respectively) the
actual 8-bit delta value is also signed - at least on all the devices I've tested (trackball, spinner, real-mice).

When I clamp from -127 to 127 instead the weird overflow behaviour with fast movements goes away; this makes game play much more predictable in games like missile command for example.

I have a PR ready which I'll push shortly.

sorgelig:
closing this issue as there is discussion in PR already.

PR #765, opened 2023-04-02
Fix mouse delta clamping to prevent overflow when moving quickly

garyjsweet:
I have an ultimarc u-trak trackball along with a homemade
rotary encoder based spinner. I've been noticing odd behaviour
when the trackball or spinner are moved very quickly. The cursor
appears to move backwards briefly, or stalls for a short period
of time. Slow to medium movements worked just fine.

After some debugging I found that the mouse deltas were being
incrrectly clamped to -255 to 255 inside user_io_mouse().

Despite the fact that byte-0 of the mouse data contains effective
sign bits for the x & y directions (in bits 4 & 5 respectively) the
actual 8-bit delta value is also signed - at least on all the devices
I've tested (trackball, spinner, real-mice).

When I clamp from -127 to 127 instead the weird overflow behaviour
with fast movements goes away; this makes game play much more
predictable in games like missile command for example.

I also cleaned up some code in input_test() which was using the
byte-0 sign bits but, in fact, just needed to cast the bytes into
signed versions.

Fixes : https://github.com/MiSTer-devel/Main_MiSTer/issues/764


input.cpp
```diff
@@ -5121,10 +5121,10 @@ int input_test(int getchar)
 								continue;
 							}
 
-							int xval, yval, zval;
-							xval = ((data[0] & 0x10) ? -256 : 0) | data[1];
-							yval = ((data[0] & 0x20) ? -256 : 0) | data[2];
-							zval = ((data[3] & 0x80) ? -256 : 0) | data[3];
+							// The bytes in the mouse delta data are signed
```
sorgelig, line 5124: This is not correct. offsets are 9 bits, not 8 bits. If mouse will give large offset then this code won't work. For example +128 is a legit value while your "fix" will mistreat it as -128.

sorgelig:
Which core you are talking about? Probably there is a bug there.

garyjsweet:
> Which core you are talking about? Probably there is a bug there.

The odd behaviour happens with at least missile command and the input tester utility.

sorgelig:
As an alternative and easier to handle in cores it's possible to clamp offsets by 8 bits, but reading from linux input event originally is correct, your code isn't.

garyjsweet:
> The odd behaviour happens with at least missile command and the input tester utility.
If you move the mouse to the right edge of the screen in the input tester and then move it quickly right, you will see the cursor jump back to the left for example.

sorgelig:
try in ao486 core

garyjsweet:
> As an alternative for easier handling in cores it's possible to clamp offsets by 8 bits, but reading from linux input event originally is correct, your code isn't.

I dumped the raw input from /dev/input/mouseX and I didn't see anything that didn't just look like an 8-bit signed delta.

sorgelig:
i think, you can read this article for example:
https://wiki.osdev.org/PS/2_Mouse

It specifically mentioned such incorrect reading.

sorgelig:
it more looks like bug in core(s) which you are trying to mask by limiting offsets to 8bits.

garyjsweet:
> it more looks like bug in core(s) which you are trying to mask by limiting offsets to 8bits.

That's certainly possible. I had kind of assumed that the "input tester" was correct, but maybe not.

garyjsweet:
> i think, you can read this article for example: https://wiki.osdev.org/PS/2_Mouse
> 
> It specifically mentioned your incorrect reading.

Thanks for the link, that should help.

sorgelig:
With real mouse it's hard to produce large offsets, so this bug wasn't visible before. Other devices mimicking mouse may produce larger offsets. 
You've mentioned DIY device - so you need to check if it follows correct PS/2 mouse protocol as well and doesn't clamp offsets to 8bit signed values.

garyjsweet:
> With real mouse it's hard to produce large offsets, so this bug wasn't visible before. Other devices mimicking mouse may produce larger offsets. You've mentioned DIY device - so you need to check if it follows correct PS/2 mouse protocol as well and doesn't clamp offsets to 8bit signed values.

The device presents as a standard USB HID mouse. The HID structure is using a standard signed 8-bit delta. It doesn't take a 9-bit field,

garyjsweet:
> try in ao486 core

The "fixed" code appears to work just fine on the ao486 core.

sorgelig:
I was talking about "non-fixed" code in ao486. 

> The device presents as a standard USB HID mouse. The HID structure is using a standard signed 8-bit delta. It doesn't take a 9-bit field,

Ok. USB HID has different stricture which is converted by linux driver to PS/2, so it doesn't matter how many bits are mentioned in USB descriptor - it will be converted to 9bit signed value for PS/2 structure.

garyjsweet:
>  check if it follows correct PS/2 mouse protocol as well and doesn't clamp offsets to 8bit signed values.

Am I correct in assuming that the cores should be expecting to receive PS/2 style mouse data then (i.e. 9-bit mouse deltas)?
I don't imagine many people are using a real PS/2 mouse (is this even possible with MiSTer?).

It's possible that no-one ever sees 9-bit data during testing given that USB is likely the predominant mouse connection which might explain why cores could be buggy in that regard.

input.cpp
garyjsweet: It's worth pointing out that this wasn't actually part of the "fix", it just seemed like a simplification (but if values are truly 9-bits, I agree it's wrong). The fix for my problem came from the other change of clamping deltas to -127 to 127 (but as stated, that could be buggy cores).

sorgelig:
> It's possible that no-one ever sees 9-bit data during testing given that USB is likely the predominant mouse connection which might explain why cores could be buggy in that regard.

Yes. This is what i've wrote above. probably no one saw >127 values and assume it's 8bit signed value.

> Am I correct in assuming that the cores should be expecting to receive PS/2 style mouse data then (i.e. 9-bit mouse deltas)? I don't imagine many people are using a real PS/2 mouse (is this even possible with MiSTer?).

It doesn't matter if it's USB or real PS/2 device (which you cannot plug directly to MiSTer, of course). It follows PS/2 structure definition. And definition clearly says: 9bit X/Y/Z offsets. 
You may open issue in input tester, so author will check it. It's more correct to fix the bug than hiding it.

garyjsweet:
> You may open issue in input tester, so author will check it. It's more correct to fix the bug than hiding it.

Will do. Thanks for your help.

garyjsweet:
Raised issue https://github.com/MiSTer-devel/InputTest_MiSTer/issues/14 against the input tester core.

sorgelig:
Ok. after more study it turned out neither my original code correct nor yours :) Neither the code in missile command.
I need to double confirm it. I will check.

sorgelig:
hmm.. nope.. my code is correct. I see correct small offsets while the link in wiki.osdev i gave above describes it incorrectly.

birdybro:
> > Which core you are talking about? Probably there is a bug there.
> 
> The odd behaviour happens with at least missile command and the input tester utility.

Both cores use input code made by @JimmyStones so maybe he has some "input" (heh) on this.

PR #774, opened 2023-04-30
This change implements dynamic wallpaper in the Menu.

soulctcher:
Enable game_wallpapers=1 in the configuration file. Once enabled, if there is a .png or .jpg in the wallpapers folder for whatever is selected in the file selection menu, the background will load from that image file. Each time the selection changes, the background will change with it, if there is a file with that name. If there is not a file that matches, you can have a default.png or default.jpg in the wallpapers folder and it will load that image instead. If that fails, you will default to the checkerboard.


poopybunghole:
Clever idea hope it gets implemented into Mister

asturur:
Cool idea, but maybe a small delay before reading / setting, i wouldn't want so much reading when i m quickly browsing the menu, just food for thought

soulctcher:
> Cool idea, but maybe a small delay before reading / setting, i wouldn't want so much reading when i m quickly browsing the menu, just food for thought

Idea implemented.  There's now an additional configuration variable that you can set in milliseconds between 0-3000 for that delay.  I've found that about 100ms is the sweet spot for allowing the super-speedy scrolling, but a near-instant update once stopping, but you can push it up higher if you want.

I also added a filter to pull characters that would be invalid in file paths out, which fixes entries like "Neo Geo MVS/AES".

soulctcher:
If you want it to not redraw the background on that first menu item change which has a longer delay than the rest, you'll have to up the delay to about 500ms, but that does mean that it takes a little longer to draw it once you stop on a menu item.

calle81:
Great idea! Makes Mister more fun to tinker with and great when kids use the device and easily can identify their favorite games.

sorgelig:
MiSTer has no GPU and caching for huge graphics. Cycling wallpapers will give input lag in menu core as time required to load the picture is significant. This is why it's not implemented. Not because i can't.

asturur:
> > Cool idea, but maybe a small delay before reading / setting, i wouldn't want so much reading when i m quickly browsing the menu, just food for thought
> 
> Idea implemented. There's now an additional configuration variable that you can set in milliseconds between 0-3000 for that delay. I've found that about 100ms is the sweet spot for allowing the super-speedy scrolling, but a near-instant update once stopping, but you can push it up higher if you want.
> 
> I also added a filter to pull characters that would be invalid in file paths out, which fixes entries like "Neo Geo MVS/AES".

Would you share a build for curiosity?

soulctcher:
> MiSTer has no GPU and caching for huge graphics. Cycling wallpapers will give input lag in menu core as time required to load the picture is significant. This is why it's not implemented. Not because i can't.

Thank you kindly for the response, sorgelig.  No significant input lag in the menu core is demonstrated with the implementation unless you leave it on 0ms immediate loading, though even that is completely tolerable.  Input lag is completely eliminated while navigating with the 100-500ms delay configured.

This is implemented by re-using your existing call that places the initial wallpaper into the framebuffer.  While an "expensive" operation, it's only expensive momentarily while the menu does nothing else.  Since the main menu wallpaper isn't used during core operations, it has no impact on any core operations.  The only thing most people are doing inside the main menu is navigating.  Additionally, since I have implemented this as an option and not a mandatory setting, it would be trivial to simply allow for it to be configured for those that wish to use it, while those who do not want to could leave the option out or disabled.

I kindly ask that you load and evaluate the actual implementation as it is a very practical way to provide some basic additional customization to your wonderful project in a simple fashion.  However, in the end, it is your project and you have your own goals/rules/needs/wants for the project.  If you decide not to implement this change, I do respect that and I thank you for your time.

soulctcher:
> Would you share a build for curiosity?

Certainly.  Give me just a bit and I'll toss a fresh MiSTer binary on my repository for testing purposes.  Please keep in mind that you'll need to enable at least game_wallpapers=1 and optionally game_wallpapers_delay=<a value in ms between 0-3000> in your ini file.

soulctcher:
For those interested in a quick test:
[https://github.com/soulctcher/Main_MiSTer/releases](https://github.com/soulctcher/Main_MiSTer/releases)

sorgelig:
> I kindly ask that you load and evaluate the actual implementation as it is a very practical way to provide some basic additional customization to your wonderful project in a simple fashion. However, in the end, it is your project and you have your own goals/rules/needs/wants for the project.

It's not because of my wish. I've clearly stated it's very time and resource consuming operation - too heavy for MiSTer's CPU. When new wallpaper will start to get loaded and decompressed, you may press the button at the same time. System won't react on it and will wait quite a lot to be noticeable. It's not about input lag of some milliseconds. It's about unresponsive system for short period. I've implemented original wallpaper loading, so of course i know how it works and what to expect from it.
Menu core can process more tasks while idling. For example some people (like me) may copy large data over network, or do some linux servicing directly on MiSTer.. All these tasks will be affected by wallpaper loading. At last, why not to turn MiSTer off if you don't use it long time - save the power :)

asturur:
I would like to voice an opinion, is a matter of wish not of technical issues.
You like to do network file copies or servicing, while someone maybe would say no, turn off the network that you are disturbing how my wallpaper load. Some people like wallpapers, other like responsiveness.

The 'it cant be done' is more a 'that is not the direction i want the project to go' and that is totally fine as long as we are clear with each other, otherwise the developer could start digging resources to improbe the feature and then hitting another refusal later.

The feature seems implemented as an option, so who doesn't like it or think is too stressing for the system could simply not enable it and not bother find a wallpaper for each menu item. The mister is also dual core if i m not wrong so eventually the graphic loading could be implemented in a way it can be interrupted on input or execute with the lowest priority possible.

video.cpp
```diff
@@ -3353,7 +3416,7 @@ void video_menu_bg(int n, int idle)
 			switch (n)
 			{
 			case 1:
-				if (!menubg) menubg = load_bg();
+				if (!menubg || cfg.game_wallpapers) menubg = load_bg();  // The OR statement allows reloading of the background image if the option is enabled.
```
asturur, line 3419: not sure what this does, but you don't want to reload the background if you didn't find a name for the entry if the old one is still displayed. I don't have a full grasp of the code and just from the diff i don't get it

asturur: my question is, if i m displaying default.png and i have to load again default.png, what happens? is it smart enough to not doing anything?

KremlingKuthroat19:
Great concept! Quick question: will this work with MGL files?

video.cpp
soulctcher: Ah... That is a good question and has value in reducing the operations if possible.  I'll look at that next to ensure that gets optimized.

soulctcher:
> Great concept! Quick question: will this work with MGL files?

Yes.  In fact in my video posted on Twitter, my main menu is a combination of .mgl files and .mra soft links.  It also works on directory names.

soulctcher:
> I would like to voice an opinion, is a matter of wish not of technical issues. You like to do network file copies or servicing, while someone maybe would say no, turn off the network that you are disturbing how my wallpaper load. Some people like wallpapers, other like responsiveness.
> 
> The 'it cant be done' is more a 'that is not the direction i want the project to go'

This is how I also look at it. It's an option that could be disabled. I also have some opportunities to improve the implementation already. Why not allow the option to exist?  I suppose at a minimum, I can maintain a fork, but that would be a tragic split.

birdybro:
> Why not allow the option to exist?

I believe sorgelig already explained why.

soulctcher:
> In addition to the hardware not having a gpu or cache so this will likely result in graphical glitches upon transition

Like any other new feature, there will always be a potential for issues, however this implementation uses existing code to fill the frame buffer that is already in use and proven. So hopefully few, if any, issues.

> and input lag

I've already demonstrated the minimal impact the draw operation has with delay at 0ms, but this can be mitigated even further by using the configuration option that I've implemented for delay.  Also, let's keep in mind that we're talking about the main menu only, not in-core OSD. So there is no impact to input performance during core operations.

> (which will lead to an increased frequency of issues raised by users on github, the forums, and discord, which we all have to support), it will also be extra code added into an already pretty large codebase that adds to the maintenance burden. Each bit of code added and the performance impact it has on the platform has to be carefully considered.

I appreciate all of what is being said, but the same could be said about literally any and every change to the project. In reality, this is a simplistic change with mostly already implemented and proven code.  If it's to be rejected, then so be it, but I think you'll find there's a lot of people who would be fine with this very minor quality of life improvement.

soulctcher:
> Precisely. Sometimes you gotta argue the cost/benefit. It's healthy to do that and good for any project.

I have no quarrel with that.  There have been calls for a nicer interface for a while and it's very understandable why it hasn't occurred in the way that most people want it to.  But here we have a compromise that is very reasonable.  Please know that there is an audience for this change.

We cannot make MiSTer into EmulationStation, and with good reason.  _But_, updating the framebuffer that carries the wallpaper?  That we can do, and rather simply.  And I will reduce the operation from happening in a further instance when I get a moment to sit down and go over the scenario @asturur mentioned above, which should help a bit more with a minimal code change.  In fact, I believe I've even found a reduction in some of what I wrote originally which will probably result in fewer lines even after adding that.

I'm not looking to create a slippery slope of added features or to drastically change the direction of the project here, and I want to be very clear that I respect everything that everyone (you, @sorgelig, and others involved) has done.

asturur:
This code as it is is still a nice to have even if it doesn't get merged, fortunately main doesn't get updated that often to be a burden. There are a bunch of repos with autoforks that get auto updated and auto built ( in primis mister-db9 in which you can use the snac pad to control the menus and with, and is around and alive since long )  you can check if there is a place there for the feature too, or you can check what actions they use to keep the build alive

soulctcher:
Much of the code I wrote has been moved out of video.cpp and into menu.cpp allowing for further things to occur more naturally. This also gets out of most of what was happening in video.cpp most of the time.  The ini variables have been renamed to dynamic_wallpaper and dynamic_wallpaper_delay to match what I was calling it originally.  This latest update cleaned up a few things, added softer file matching so you can have naming that'll not require all the additional parentheses/brackets.  For example, you can have 1942.png and it'll find that just fine for 1942 (Revision B).  It also no longer attempts to redraw the background if the next file name matches the previous one.  (Thanks @asturur for the suggestion.)

birdybro:
If you hold the button down to scroll through multiple MRA files, does instantly stopping actually instantly stop scrolling, or are there queued inputs still processing to wait for the images to load one after the other?

birdybro:
To improve the speed of scrolling upon holding the button down (which based on the video appears slower than without this feature), maybe it should be adjusted so no image is loaded at all if the held input's frequency of bouncing is under a certain threshold. It didn't look like this is already implemented based on the section of the video example where the button is held down, it looks like it constantly tries to load each wallpaper.

soulctcher:
> i checked the code quickly, i would expect more condition that reset the timer to 0 but i see only one. Is the timer reset each menu item? i can't see that, but it could be that i m not familiar with the codebase enough

If the menu selection changes, the wallpaper_timer is set to a new value (time since boot + the value of dynamic_wallpaper_delay), so the only time you need to reset the timer back to zero is when you've drawn the background after expiration.

Although, the thought did make me think of a race condition that I want to test, just to see how the current code behaves between entering a core that loads super fast and having a very long (3000ms) delay value set.  I _think_ the framebuffer will get flushed properly and the menu code will get halted, if I remember correctly, but I'm still learning the code.

soulctcher:
> Why not just make the delay to attempt to load a new wallpaper the default?

I can and will do that, but I still prefer to keep the option available if someone wants to override it with their own value.  Thanks for the feedback; the input is welcome.

sorgelig:
At first, Menu is made as a startup core which user briefly see before selecting the core. So there is no point to make a slideshow from it or similar. Probably, someone can make a core specifically to be a slideshow which can be started after timeout. So there is a way to customize without bringing it to Main.
I'm definitely not going to add feature which will potentially interfere the work. Random user will enable it without understanding the consequences and then later complain about them. And it's not a secret some users downloaded "pre-configured" packs where it can be enabled. 
Current concept is: "when user do nothing, MiSTer do nothing". And i don't want to change this concept.

soulctcher:
> Current concept is: "When user does nothing, MiSTer does nothing". And i don't want to change this concept.

I'm not sure how exactly you think what I'm doing here interferes with this concept in any way; it doesn't.  If I do nothing on MiSTer, it does nothing.  There is no "slideshow" here.  You must actively change the menu for it to reload the framebuffer with the new image.

But as I've said previously, it's your project, so if you don't want to merge these changes, that's your choice.

sromeroi:
Is this then not being implemented in the main mister menu?

Not only it can provide nice backgrounds for "folders" (Arcade, Consoles, etc)... It might also become a simple way to have game-screenshots when browsing the games (maybe using a different folder like "screenshots" or similar, instead of "wallpapers").

sromeroi:
> > Is this then not being implemented in the main mister menu?
> 
> Correct. Read the conversation.

I read it before posting. It was a rhetorical question.

birdybro:
If every feature that was interesting to some people were introduced the codebase in Main would be gargantuan by now, read the closed pull requests in the past and the closed issues with feature requests.

asturur:
Every developer should be able to own the and drive the project as he sees it may fit.
Main is not updated every week, and forks exists for this exact reason, you can still have the wallpapers with your own build.

The code is here, is not going to disappear,  the author can also link a binary built
maybe.

Let's cheers we have this rather than being sad is not merged

birdybro:
Exactly, great way to look at it. That's the spirit of open source, if someone wants to customize it themselves they are free to.

soulctcher:
100% agree.  The beauty of open source is that you can go grab a binary right now from my releases page:
[https://github.com/soulctcher/Main_MiSTer/releases](https://github.com/soulctcher/Main_MiSTer/releases)

I plan on continuing to make improvements to the code to tidy it up, hopefully, improve efficiency, etc.

thisisacleverusername:
Possible solution to make everyone happy: 

Don't make the background change automatic; make it such that you *can* press a key to change the background to match the selected item. 

-This will make menu scrolling absolutely 100% unaffected, with zero new lag or interruptions. 
-This will also meet the concept of "When user does nothing, MiSTer does nothing".
-This will also set expectations if, say, an FTP transfer is processing in the background - the user has stopped and requested an image change, and their exceptions will be to wait for it. 
-This will make sure new and unsuspecting users don't accidentally turn on a feature with permanent repercussions they don't understand - if they never hit the "change background" button, they'll never see anything change, and if they do, they'll recognize the cause and effect. 

Thoughts?

Natrox:
I know I am weeks late but I need to address something.

I want to also make you aware that I have already implemented dynamic loading in an older PR: https://github.com/MiSTer-devel/Main_MiSTer/pull/704

In this PR I have changed the code to allow for seamless background loading of new wallpapers, without impacting user input or stalling the program. The PR was closed because I recognized it was not going to be accepted ever.

However, if you intend to keep a fork going, I strongly recommend you build the feature off my previous work, instead of using a re-implementation. You can use the `load_bg` command at any time, external to the binary or internally. 

If you need assistance, let me know.

asturur:
I love this pr and the only issue i have is indeed blocking input when the timing is unlucky. I would love for someone to merge the 2 works and and have a fork with the auto patch ( as mister db9 fork does adding s bunch of patches to all cores )

I could also try to maintain the fork for the ci part , i m open to use my time for collaboration sake

Daddu3:
Very eager ATM to try this, but I can't seem to make this work for some reason. Wondering what I'm doing wrong. 

I had downloaded the latest Mister-file in the releases, and my ini has following settings:


```
dynamic_wallpaper=1      ; Enables wallpapers  in the configuration file. 
                         ; Once enabled, if there is a .png or .jpg in the wallpapers folder for whatever is selected in the file selection menu, the background will load from that image file.
						  
dynamic_wallpaper_delay= 
                         ;Add a delay that helps with menu navigation speed between 0-3000 milliseconds.
```

Also I wrote my pngs like this:

image

hittm4n:
I realise this is a year on. I stumbled on this little project as i'm looking for a way for my MiSTers background to change every 5 minutes or so. I have 40 beautiful wallpapers and i only get to view one of them. I sometimes load my MiSTer just to have some music playing (Lukhash and MBR mostly) and something retro showing on my TV. But i am stuck on that one initial loaded image. BORING! This can't be hard to implement right? A new background every x minutes? I've read everything here and i just wanted to add my voice to the "i would love this in MiSter" choir. Background music, done. Loading random wallpapers on menucore load, done. Load an image for each menu core selected.... please!!! And if not that, just allow the background to change every now and then... thanks soulcatchr, i see how hard you tried to get this merged :)

manotroll:
The Mega Drive is capable of loading the screens, but an FPGA could not

add the feature and allow the user to activate it or not

in this case, the "caba" appears on the Mega Drive
https://www.youtube.com/watch?v=8l1l1yclVDo

birdybro:
>but an FPGA could not

Please read the thread because it's not about whether or not MiSTer FPGA can do this.

Natrox:
> > but an FPGA could not
> 
> Please read the thread because it's not about whether or not MiSTer FPGA can do this.

I concur. As for the dynamic wallpaper functionality, I will merge upstream tomorrow and can provide binaries for testing. 

Unfortunately this PR is probably not going to be accepted, so you'd have to DIY a bit if you'd like the functionality. 

Tag me if you'd like to be informed when the branch is ready.

PR #796, opened 2023-08-11
Feature/network

willoucom:
I wasn't happy with how it works now, mounting a network drive in /media/fat/cifs doesn't seem logical and simple enough for users.

I propose to add a games search path in /media/network which can be used with cifs and nfs.

I kept the original /media/fat/cifs folder to avoid creating problems for current users of the folder.

I also changed the comment block to reflect how the method works (in the correct order).

### DIFF
--- a/file_io.cpp
+++ b/file_io.cpp
@@ -943,6 +943,8 @@ int findPrefixDir(char *dir, size_t dir_len)
 	// /media/fat
 	// /media/usb<0..5>
 	// /media/usb<0..5>/games
+	// /media/network
+	// /media/network/games
 	// /media/fat/cifs
 	// /media/fat/cifs/games
 	// /media/fat/games/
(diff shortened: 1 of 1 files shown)


birdybro:
This seems redundant to me. What problem are you attempting to solve specifically?

willoucom:
I think the current tree is confusing:
- SD in /media/fat 
- USB in /media/usb<0..5>
- NAS in /media/fat/cifs 

I think a different approach may be better:
- SD in /media/fat
- USB in /media/usb<0..5>
- NAS in /media/network

It doesn't solve anything, but it makes it easier to understand (and less ticklish for old-school beaded Linux veterans like me)

I understand your desire not to encourage novice users to use the network share, and that the current cifs mount status is a small hack. However, I think that even a hack can be improved without it formalizing a functioning that could be problematic in the future.

Actually I would like to get rid of /media/fat/cifs but I can't realistically introduce a change that could break even minor usage. I think it is possible to eliminate this problematic operation with a change management plan in several stages in order to reach a stable state in a few months.
In essence, this amendment would be the first step.
The next step would be to either work on changes in the network stack to make it stable and eliminate latency, or introduce a local caching mechanism. But it is probably premature to address these subjects now.

sorgelig:
Probably /media/network is more common than cifs. It can be accepted.
But we aren't talking about "more stable" later. Network cannot be stable by its nature. So it's not a trial feature. It probably will be this way always. Network cannot provide the same latency as local drives.

birdybro:
Good point about the latency as well. It's important for people to remember that network latency (especially over WiFi) can cause issues in cd-based cores occasionally (which can result in false positive issues).

PR #811, opened 2023-09-04
Add a second JAMMA quirk to support 4-player JAMMA-VERSUS

javier-rodas:
Add a second JAMMA quirk to support 4-player JAMMA-VERSUS with two JammaSD/J-PAC (or similar USB JAMMA) controller interfaces.


atrac17:
Lovely stuff. Thanks for the merge!

PR #813, opened 2023-09-07
Toolchain improvements

c0d3h4x0r:
get_toolchain.sh:
 - renamed from setup_default_toolchain.sh
 - no longer needs to be 'sourced'
 - no longer needs to set environment variables for build to work
 - added helpful comments
 - more robust handling of interrupted/failed download or extraction
 - always place extracted GNU ARM toolchain under dir named 'gcc-arm' (version independent name so Makefile and VS project file can safely reference it)
 - added apt-get installation of pinned versions of 'make' and 'plink' dependencies

Makefile:
 - don't repeat lengthy sed statement
 - always use max parallelism
 - update BASE to no longer require things be on the PATH
 - add phony 'toolchain' target so users can run 'make toolchain'

MiSTer.vcxproj:
 - update include paths to reference new in-project gcc-arm dir
 - replace old toolchain setup script with new one

fpga_io.cpp:
 - fix for 'impossible constraints' (in/out register conflicts) exposed by trying the GCC ARM 10.5 compiler


sorgelig:
Sorry, i have to roll back the changes. Compiling in wsl was already slow, but with these changes it becomes painfully slow! Even small change in one takes ages to compile. May be it's because my toolchain is on SSD while project is on HDD, so it's faster. 
Also i found that build log in VS doesn't display actual compilation progress but spits whole list of files and then wait for long time without any changes. 
There are also some unneeded restrictions for versions of make and putty. These versions aren't available in wsl and i'm not going to install them from 3rd party repos and risking break compatibility with other tools in wsl. Current versions of make and putty are working fine.

c0d3h4x0r:
I am working on Windows 11 with WSL2, and I only use SSDs.  I am building inside Ubuntu inside of WSL2.  

Without the `MAKEFLAGS` addition, building is _extremely_ slow -- like, 20 minutes or more! -- because `make` only uses one CPU core at a time.  With these changes, my build is very fast instead, because all of my available CPU cores get used.

Without these changes, the first thing the build does is generate all the `.d` files, but it shows no output while doing it.  When only using one CPU core, the generation of `.d` files takes many minutes, and the build doesn't appear to be doing anything during that time.  With these changes, the build shows while it is generating `.d` files, as it should.

Your computer probably has an abnormal hardware bottleneck, maybe because your project is on extremely slow HDD.  Is your HDD externally attached via USB?

Those version of `make` and `putty` are available in WSL.  I am using WSL2 (Ubuntu 22.04).  Perhaps you are using an older version of Ubuntu?

Explicitly pinning all build toolchain dependencies is always the best idea, because it makes your builds immune to version changes and lets other people know exactly what they need to get a successful build.  Without explicit pinning, users get random/uknown versions of tools used by your build, and when a new version releases containing a breaking change, suddenly nobody can build.

c0d3h4x0r:
Or maybe you are still using the old WSL v1?  WSL2 is much higher performance.

sorgelig:
My HDD are on SATA. It's relatively new 10TB HDD. 
It doesn't take 20mins to build as you describe. It takes less than 5 mins i think if compiled freshly, and pretty quick if i change something in file and compile again. Dependency build output can be added. It does once till the next cleanup of project.

Anyway, glad to hear you have a very modern and fast hardware, but i don't want to force myself to upgrade. I'm pretty much happy with my current setup and not willing to make such radical changes in build script.

sorgelig:
I've pushed some of your changes giving benefits in my environment, including multi-thread compilation.
I think the main problem is moving compiler into project folder.

c0d3h4x0r:
Interesting. That shouldn't impact build speed at all.  So maybe the real problem is antivirus scanning or OS search indexing of the newly-extracted compiler files?  Have you excluded your project folder from Windows Search Indexer and from whatever antivirus software you may be running?

c0d3h4x0r:
Side note: I don't think `wsl --update` impacts any already-installed Linux distro under WSL. It just updates WSL itself to make the v2 engine available for future installs. There's some other command you must run to change an already-installed distro to use the WSL v2 engine.

c0d3h4x0r:
If I update the script to first che k for Ubuntu 22.04 before doing `apt-get install` of those specific versions of `make` and `putty`, will that be more acceptable?

sorgelig:
Right, i don't want to update Ubuntu distro as i have many settings there. I think install make and putty without specific versions should be sufficient. I think Ubuntu 20 is enough for compatibility.

> will that be more acceptable?

Not sure what do you mean. Keeping toolchain in project folder isn't a good option. I think script should install toolchain into default place.

c0d3h4x0r:
I mean like this: https://github.com/c0d3h4x0r/Main_MiSTer/commit/14c46cbcb53f5447c4c0f22638894d39ae072a24

c0d3h4x0r:
I am still very confused why you have a big performance problem with gcc inside the project folder, but I don't.  Are you using some add-on plugin with Visual Studio that causes it to index all the gcc files?  I need to understand the difference before I can accept that it's a bad idea.

sorgelig:
> I mean like this: [c0d3h4x0r@3706c3f](https://github.com/c0d3h4x0r/Main_MiSTer/commit/3706c3fc94e240f4a853e7984af77efe8d73b1db)

In general it's ok, but i don't understand your obsession of specific make/putty versions. I think they are fine as along as they not extremely old. I believe any versions from Ubuntu 20 and up should work. MiSTer doesn't use very specific features of either utility.

> I am still very confused why you have a big performance problem with gcc inside the project folder, but I don't. Are you using some add-on plugin with Visual Studio that causes it to index all the gcc files? I need to understand the difference before I can accept that it's a bad idea.

I don't use any plugins in VS. Actually it's very slow even if i use make in command prompt of wsl. Probably it's because of Windows 11 itself. I start to notice that recent Win11 updates are slower with HDD. Even after migrating from Win10 to first revision of Win11 i've noticed how slower it is.

c0d3h4x0r:
I'm a build engineer for a living, so I care about these things :)  Unpinned dependencies are always a source of trouble.

I've never noticed speed differences with HDD access between Windows 10 and Windows 11.  I have two different desktop PCs running Windows 11, and both have large HDDs for storing Steam-installed games, and I have no performance problems with the HDDs on either machine.  It really sounds like there's something wrong with your HDD.  Have you run a SMART check or chkdsk or anything on it?  Another thing that can happen is lots of files (like thousands) can accumulate in the %TEMP% folder, which NTFS doesn't handle well and it slows down the entire system, so if your %TEMP% folder is on HDD, you might want to try manually clearing it out.

PR #828, opened 2023-09-24
N64 joystick emulation

markun:


jsmolina:
@markun  it might fix this issue? https://github.com/MiSTer-devel/Main_MiSTer/issues/829

PR #834, opened 2023-10-09
Scale N64 analog optimally for different gate shapes

Lemonici:
The original code for N64 analog scaling used the cardinal directions to get a maximum observed value. This works for controllers with circular gates but since the N64 had an octagonal gate with diagonals that had more range than the cardinals it means that controllers that try to imitate that will have a reduced range before hitting a large outer deadzone. To address this, both the cardinal and general maximum distances need to be stored and when scaling, the less aggressive one should be used.

Unfortunately, general distance requires square roots. To avoid a `sqrt()` every time the input loop ran in the N64 core, I have the code only track the sum of squares and update the distance only when a new maximum is observed. This does require including "math.h" in the main binary but I couldn't find an elegant way to avoid it.


Lemonici:
(Shoutout to Mark_Eire from the Discord for extensive testing)

PR #838, opened 2023-10-15
Added controller handling for new N64 options

trbocode:
Some games use two controllers, or hold analog stick in the right hand


sorgelig:
please move it into n64 support code.

trbocode:
I tried to cache the controller variables, but didn't find a contained way to do it. What would be the best way?

sorgelig:
I think you need to talk to @RobertPeip to coordinate it between HPS and core.

trbocode:
I think this should be good enough - No 3 calls to sscanf anymore

PR #841, opened 2023-10-21
Update names.txt: added Sega Saturn

mrfpga:


MJY71:
It's already there. Kitrinx added that and PSX under the 'Computer Cores' section for some reason quite a while back.

sorgelig:
I've corrected it.

PR #849, opened 2023-11-08
Fix: when extensions are less than 3 characters send NUL

hansl:
Currently if the extension is less than 3 characters the last characters of the filename will be sent to the FPGA. These are not consistent and would not help the core establish what kind of file it is.

There are few ROMs with less than 3 characters, but they do exist (e.g. .ws for WonderSwan). It is unclear reading the code how those are handled outside of this specific function.

## Security considerations
This does not fix a buffer overflow as the code to get there would fail if the complete filename would be less than 4 characters (e.g. the root filesystem is read only).

## Testing
This has been tested with the happy path; the code at least works with current files.

### DIFF
--- a/user_io.cpp
+++ b/user_io.cpp
@@ -2510,7 +2510,8 @@ int user_io_file_tx(const char* name, unsigned char index, char opensave, char m
 	int len = strlen(f.name);
 	char *p = strrchr(f.name, '.');
 	if (p == 0) {
-	    p = f.name + len - 4;
+            // In case a '.' is not found, send all `NUL` characters.
+	    p = f.name + len;
 	}
 	user_io_file_info(p);
 
--- a/user_io.cpp
+++ b/user_io.cpp
@@ -1961,8 +1961,12 @@ void user_io_file_info(const char *ext)
 {
 	EnableFpga();
 	spi8(FIO_FILE_INFO);
-	spi_w(toupper(ext[0]) << 8 | toupper(ext[1]));
-	spi_w(toupper(ext[2]) << 8 | toupper(ext[3]));
+	char c1 = *ext ? toupper(*ext++) : 0;
+	char c2 = *ext ? toupper(*ext++) : 0;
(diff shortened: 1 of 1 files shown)


user_io.cpp
```diff
@@ -2504,7 +2508,10 @@ int user_io_file_tx(const char* name, unsigned char index, char opensave, char m
 	user_io_set_index(index);
 
 	int len = strlen(f.name);
-	char *p = f.name + len - 4;
+	char *p = strrchr(f.name, '.');
+	if (p == 0) {
+	    p = f.name + len - 4;
```
wickerwaka, line 2513: Should this be `p = f.name + len` ?

hansl: That could work. I thought reverting to the old behaviour would be enough.

There's also an edge case where a directory has `.` in it, but the file itself has no extension. I'm not handling that case at all.

If you feel that should be done and send NUL x 4 to the core (which is what mine does at the moment), I'm okay with that change.

hansl: I did this, I guess it makes sense. 👍

sorgelig:
While some use cases may exhibit problem, but you forget real usage.  Cores request specific extensions. Core knows how many letters are in requested file extension. And there are no cores requesting files with no extensions. Also, i don't remember core besides Amstrad using this info.

sorgelig:
update: i had to revert this PR. It has potential read behind the buffer. It may crash if file has no extension or extension with 1 letter.

Original code may send part of name but it won't crash and there is no real case with file path less than 4 letters.

hansl:
> It has potential read behind the buffer. It may crash if file has no extension or extension with 1 letter.

It only pushes the pointer if the character isn't `\0`. It won't read past the buffer since it won't move `ext` past it.

sorgelig:
yeah. i overlooked that part. 
Submit PR again, i will merge it.

sorgelig:
i was able to recover the commit

PR #853, opened 2023-11-19
N64 improvements.

yxkalle:
Better Auto detection of N64 games (region, security chips, save type). Better info message after loading a game. Clean up. Added references.


sorgelig:
It seems something wrong with PR.

yxkalle:
Seems like it. Fixed now.

PR #855, opened 2023-11-25
Added Dockerfile and .devcontainer

niemasd:
To make it a bit easier for developers, I added a `Dockerfile` and `.devcontainer` so that VS Code (or other environments that support Docker images) can load the development environment with all prerequisites with the click of a button, regardless of platform.


sorgelig:
I wonder why ppl like to go a hard ways :)
Windows 10/11 offers WSL which is tightly integrated. Why not use it?

niemasd:
I personally use WSL as well, but I think a lot of folks struggle with setting up dependencies. In this case, it wasn't too crazy of dependencies, but especially when some dependencies need to be compiled from source and have their own dependencies (e.g. special libraries), I think sometimes folks like a one-click solution.

In my case, I like`Dockerfile`s because they tell me exactly what commands I need to run to get it working in my WSL environment without any fuss (so for me, it's more of a nice piece of documentation that happens to also have functional uses)

sorgelig:
visual studio project is already configured to run compiler in wsl.

birdybro:
Many different ways to hunt a deer I guess. :)

PR #857, opened 2023-11-30
Added dead zone configuration for analog joysticks

yxkalle:
The user can define an analog joystick dead zone by a special device identifier and a radius (0 - 64 units).
Up to eight uniquely identifiable devices can be added to MiSTer.ini.

The identifier is constructed very much like the one used for "player_N_controller=. . .", 
to be as flexible as possible.


rsn8887:
Sorry for chiming in, but I messed quite a bit with deadzones a couple years ago. I found this reference useful on how to properly implement a radial deadzone: https://web.archive.org/web/20130517043043/http://www.third-helix.com/2013/04/doing-thumbstick-dead-zones-right/

From the math, it seems this code is doing something more than a radial deadzone?

yxkalle:
The gate is approximated by measuring the maximal distance away from origo that the joystick has for the more cardinal and diagonal angles. I'm working with polar coordinates most of the time. It is given that the bigger of a dead zone you define, the more distortion you will get. That is because the radial coordinate get more and more squished the smaller the distance from the edge of the dead zone to the edge of the gate gets. Angular coordinates are always preserved, though.

Here's a checkered image that has being mapped through the algorithm. 
Circular gate with a deadzone with a radius of 32 units inside a gate with a radius of 128:
image
Squircular cate:
image
Square gate:
image
Huge 64 unit gate:
image
A small 8 unit gate:
image

sorgelig:
> To be honest I was a bit unsure where to put this .ini parsing block, but it seems to work as intended. It only runs once per device, just after controller button assignment is shown on screen (if controller_info is set, that is).

ok. i see now.

PR #870, opened 2024-01-22
OSD + F7 = Joystick Mapping Dialog

misteraddons:
Add a keyboard hotkey to trigger the joystick mapping dialog. Useful for tournaments where novice users navigating the menu accidentally change core settings. 

Open to a better way of accomplishing this, or a different hotkey if F7 when the OSD is open is unacceptable.

The goal is to create a macro pad with a few keys: "OK", "Back", "Map Joystick", etc.

Thanks!

### DIFF
--- a/cheats.cpp
+++ b/cheats.cpp
@@ -72,7 +72,7 @@ static int find_by_crc(uint32_t romcrc)
 {
 	if (!romcrc) return 0;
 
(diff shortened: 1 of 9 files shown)


sorgelig:
MENU_JOYSYSMAP is system wide joystick settings which should be done in menu core, not in other cores.
Such shortcut will only confuse users as it will change completely different settings.
And by the way long press of USER button on MiSTer in menu core opens MENU_JOYSYSMAP.

PR #876, opened 2024-02-25
Tapto support

spark2k06:
TapTo is an open source system for launching games and custom actions using physical objects. More info:

https://github.com/wizzomafizzo/tapto

With this pull request, we support this development, which I believe is being very well received. The implementations are:

* Reader detection in the menu via the TapTo API service.
* x86 support through CMOS memory, using the TapTo API service.

TapTo Info


sorgelig:
What it gives to MiSTer?
Looks like quite weird and niche feature. I didn't get it...

birdybro:
https://www.youtube.com/watch?v=YwQviLwWHYM - Video from Lu for context as to how it works.

It is very very niche. I personally don't understand wanting to buy, program, and label NFC cards just to avoid using the menu to load a game. If it's for loading your favorites we already have the MGL format which Wizzo has written a useful script to generate these automatically for favorites.

spark2k06:
For me, what this project offers is the feeling that you have something physical about the game launching, a component that has been lost in recent times.

Because yes, there is no better experience than having, collecting, seeing and touching game covers before launching them.

I know not everyone will agree with this idea, but I love it. If it doesn't fit with sorgelig's idea and finally doesn't make it to the official Main, it doesn't matter, we'll keep an alternative fork for the use and enjoyment of those users who like it, which there are quite a few 😉

spark2k06:
Only for favourites, of course.

RetroShopPT:
I think it's very interesting for arcade games. We use NFC with Jammix Cabinets where the user chooses the game using the card. Very interesting and practical for our retrogaming shows.

spark2k06:
> Main has command interface where you can send command to launch specific MGL. Thus instead of modifying the Main, you can run some resident app which will send command to Main when NFC tag is detected. Thus TapTo can be developed (and more important - will be supported) independently. As for launch games inside x86 core - it's not a good idea. DOS has several game launchers where each game can have its own config. It can be easily modified at any time. It's much more flexible and handy than NFC tag. Especially you don't want thousands of NFC tag for each game.

The problem is that for x86 kernels, such as PCXT and ao486, this would only be valid for floppy disks. It's better to start from a VHD where you have all of them, and the interface to the NFC via CMOS memory seemed convenient, but I understand that you don't share it.

sorgelig:
ao486 has special command interface to load CD/HDD/DSK and configure hardware. Thus i would understand if you would offer a launcher using these features, so everyone could use it. But asking people to buy a special hardware for it doesn't look handy. It means user without additional hardware won't be able to launch games. It doesn't look right.

spark2k06:
> ao486 has special command interface to load CD/HDD/DSK and configure hardware. Thus i would understand if you would offer a launcher using these features, so everyone could use it. But asking people to buy a special hardware for it doesn't look handy. It means user without additional hardware won't be able to launch games. It doesn't look right.

Users without an NFC reader can still enjoy other ways of launching games. This hardware is optional for those who like the idea, like many other official MiSTer addons.

birdybro:
> The problem is that for x86 cores, such as PCXT and ao486, this would only be valid for floppy disks. It's better to start from a VHD where you have all of them, and the interface to the NFC via CMOS memory seemed convenient, but I understand that you don't share it.
> 
> https://youtu.be/A_z9OEE0csg

Just coming from my perspective as someone who has used ao486 quite a bit, I think the more idiomatic solution would be to make individual vhd's that have a very minimal MSDOS that autoexec.bat's into the game, like the x68000 sets.

spark2k06:
> > The problem is that for x86 cores, such as PCXT and ao486, this would only be valid for floppy disks. It's better to start from a VHD where you have all of them, and the interface to the NFC via CMOS memory seemed convenient, but I understand that you don't share it.
> 
> Just coming from my perspective as someone who has used ao486 quite a bit, I think the more idiomatic solution would be to make individual vhd's that have a very minimal MSDOS that autoexec.bat's into the game, like the x68000 sets.

That is another feasible way to achieve it. However, in ao486 we currently have another problem with selecting a VHD from the menu or from TapTo, and that is that it mounts in IDE 1:1

spark2k06:
I reorder changes in files for ease of maintenance and move the development into a fork of the main repository, all in the new-features branch which I will keep as the default branch. The master branch will be a synchronised copy of master from the main repository in order to automate the update process.

In addition to tapto support, there is currently another development to set up a custom loading screen, both in background and foreground mode. I have not made a pull request for this feature because I understand that it may be considered unfeasible for various reasons, but I am interested in having it for personal use... it is also possible that other users may be interested in it:

https://github.com/spark2k06/Main_MiSTer

PR #887, opened 2024-03-20
cfg attribute in reset tag for MGL file

spark2k06:
This attribute allows the use of a custom configuration setting in a core running via MGL

### DIFF
--- a/cfg.h
+++ b/cfg.h
@@ -98,6 +98,7 @@ typedef struct {
 	char osd_lock[25];
 	uint16_t osd_lock_time;
 	char debug;
+	char pathstatus[1024] = {0};
 } cfg_t;
 
 extern cfg_t cfg;
--- a/cfg.h
+++ b/cfg.h
@@ -97,10 +97,10 @@ typedef struct {
 	uint32_t controller_unique_mapping[256];
 	char osd_lock[25];
 	uint16_t osd_lock_time;
-	char debug;
-	char pathstatus[1024] = {0};
+	char debug;	
 } cfg_t;
 
+extern char cfg_pathstatus[1024];
 extern cfg_t cfg;
 
 //// functions ////
--- a/cfg.cpp
+++ b/cfg.cpp
@@ -16,6 +16,7 @@
 #include "support/arcade/mra_loader.h"
 
 cfg_t cfg;
+char cfg_pathstatus[1024] = {0};
 static FILE *orig_stdout = NULL;
 static FILE *dev_null = NULL;
 
--- a/support/arcade/mra_loader.cpp
+++ b/support/arcade/mra_loader.cpp
@@ -16,6 +16,7 @@
 
 #include "buffer.h"
 #include "mra_loader.h"
+#include "cfg.h"
 
 #define kBigTextSize 1024
 struct arc_struct {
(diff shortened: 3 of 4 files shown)


limi:
This would definitely make the goal of “consolizing” the `ao486` experience — e.g. what [0mhz.net](https://0mhz.net) is doing — a lot easier when custom configurations are needed per-game.

spark2k06:
Example:

```ini
<mistergamedescription>
    <rbf>_Computer/PCXT</rbf>
    <file delay="1" type="s" index="2" path="/media/fat/games/PCXT/HDD/PCXT_CGA_Tandy_2.3.1.vhd" cfg="/media/fat/games/PCXT/MGL/config/AlleyCat.cfg"/>
    <reset delay="1"/>
</mistergamedescription>
```

birdybro:
This will also be useful for some console cores like PSX where you may want different video crop options for different games.

cfg.h
```diff
@@ -98,6 +98,7 @@ typedef struct {
 	char osd_lock[25];
 	uint16_t osd_lock_time;
 	char debug;
+	char pathstatus[1024] = {0};
```
wickerwaka, line 101: This is not data coming from the INI file, it should not be in the `cfg_t` structure.

zakk4223:
Are you not using setname because you want to still load the user's video filters/input mappings etc? 
Maybe it makes more sense to slightly change the behavior of setname so that when it loads config files it first tries the name based on the setname tag, and if that doesn't exist it falls back to <core_name_from_rbf> based name. Any settings that are saved would just write to the setname based name.

This way users can setup global settings like video filters and input mappings that will apply to all mgls loaded, and if you setup all the mgls with setname fields, if a user needs to remap controls/change video filters/etc just for that game, it'll only apply to that mgl/game.

cfg.h
spark2k06: Fixed in this commit:

https://github.com/MiSTer-devel/Main_MiSTer/pull/887/commits/7fbc796aafba5ae27a225b5736a92c96544e73fd

spark2k06:
> Are you not using setname because you want to still load the user's video filters/input mappings etc? Maybe it makes more sense to slightly change the behavior of setname so that when it loads config files it first tries the name based on the setname tag, and if that doesn't exist it falls back to <core_name_from_rbf> based name. Any settings that are saved would just write to the setname based name.
> 
> This way users can setup global settings like video filters and input mappings that will apply to all mgls loaded, and if you setup all the mgls with setname fields, if a user needs to remap controls/change video filters/etc just for that game, it'll only apply to that mgl/game.

I hadn't thought about it, I had initially focused on the configuration of the core status bits.

spark2k06:
However, I also think that the filters, at least for me, see them more as something personal to the core itself than to the game.

spark2k06:
To be more flexible, I am now experimenting with this other possibility:

https://github.com/spark2k06/Main_MiSTer/commit/4249abf29a0d3c656c2ebb5628fe31089bdaa570
https://github.com/spark2k06/Main_MiSTer/commit/006fdc94ce5610201a63ff032d4570290ba3429d

It allows us to have different types of customised configuration:

* main
* audio
* video
* x86sys

In a customised path and base filename.

So we will be able to do something like this:

```ini
<mistergamedescription>
    <rbf>_Computer/PCXT</rbf>
    <cfg path="/media/fat/games/PCXT/MGL/config" filename="AlleyCat" main="" video=""/>
    <file delay="1" type="s" index="2" path="/media/fat/games/PCXT/HDD/PCXT_CGA_Tandy_2.3.1.vhd"/>
    <reset delay="1"/>
</mistergamedescription>
```
In this case, we would be indicating that, in the path and base name indicated (AlleyCat), the configuration must be saved and loaded for main and video configurations. In case one of these attributes is not indicated, the core default will be used.

spark2k06:
I'm realising that this has a few implications, and I'm going to make sure it works well in all areas, both saving and loading configurations. I'm closing PR and I'll create it again with a single commit and test it well. Sorry for the inconvenience.

birdybro:
Alternatively in the future you can squash commits and do a force push to your fork and it will show only one commit in the history at the end.

PR #889, opened 2024-03-24
cfg tag for MGL file

spark2k06:
It allows us to have a custom configuration based on the path to the game folder.

Example:

```ini
<mistergamedescription>
    <rbf>_Computer/PCXT</rbf>
    <cfg path="/media/fat/games/PCXT/MGL/config/AlleyCat"/>
    <file delay="1" type="s" index="2" path="/media/fat/games/PCXT/HDD/PCXT_CGA_Tandy_2.3.1.vhd"/>
    <reset delay="1"/>
</mistergamedescription>
```

It is only necessary to indicate the customised path. If it does not exist, it will be created automatically. If there are no configuration files, the values will be taken from those used by the core. When saving the configuration, they will be saved in the custom path. From then on, the values will be taken from here, and it is compatible with any kind of configuration, not only the main core configuration (cur_status).

### DIFF
--- a/cheats.cpp
+++ b/cheats.cpp
@@ -72,7 +72,7 @@ static int find_by_crc(uint32_t romcrc)
 {
 	if (!romcrc) return 0;
 
-	sprintf(cheat_zip, "%s/cheats/%s", getRootDir(), CoreName);
+	sprintf(cheat_zip, "%s/cheats/%s", getRootDir(), CoreName2);
 	DIR *d = opendir(cheat_zip);
 	if (!d)
 	{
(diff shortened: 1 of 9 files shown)


zakk4223:
I think this is going to make it difficult to distribute globally usable  MGLs with such a feature. You can't assume something like /media/fat/games/PCXT/.... is necessarily writable on a given user's Mister. Or that they haven't fully switched to network storage and would prefer all 'games/CORE' stuff go there. At a minimum it needs to obey the 'game dir' search path (usb, cifs etc). 

Problem with that is users who have stuff like that setup sometimes have a lot of it set to read only. And nothing really writes there (other than writable vhds). Having config files outside of /media/fat/config is going to be..unexpected.

I think if setname triggered the same type of fallback (between core_name and orig_name) with the addition that it also did fallback for HomeDir() that would also work? Config files would stay in /media/fat/config, users would see a visual indication of what's going on (sidebar name) and the HomeDir() fallback would mean bios files etc could be loaded from original core directory, without users having to create symlinks.

Problem is the implementation is much less clean, because all the filenames have the core name embedded in them :/

(disclaimer: obviously I'm not Sorg so feel free to ignore me)

sorgelig:
I agree solution isn't clean.
I think it would be better and sufficient to use some flag for `setname` tag when HomeDir is still from original core name while all config files are from setname.
I will think about implementation.

PR #900, opened 2024-05-15
Add for ZIP archives use your own <zip> suffix

TpaBkaY:


sorgelig:
That's quite big change in sorting. I'm not sure if i want to see zip files between folder names.
Folders can be used as a category of ROMs while zip may represent some subset of ROMs. 
Also zip extension is visible, so additional <zip> will be redundant and will cut the space.

TpaBkaY:
At the moment folders are showed between ZIP archives and has the same \<DIR\> suffix. And it is very difficult to understand where the folder is and where it is not. This was the main reason for creating this patch set.

This sorting change in patch is necessary so that folders are always at the top of the list.

PR #907, opened 2024-06-29
Change in handling overlapping button presses (fixes issue #824)

### LINKED ISSUE #824: Bug: Odd input behaviour when assigning multiple inputs to a single button
I have noticed some odd behaviour with inputs on MiSTer when assigning multiple inputs to a single button. Consider these two scenarios:

R-Type
A = shot, B = fire pod, C = A (set C to autofire = 32ms). The goal of this configuration is to allow for autofire of the shot (hold C) and allow for shot to be charged when holding A. Keep C held down then switch to holding A to charge beam. Notice that A does not start charging the beam. The only way this will work is if you deliberately introduce a slight delay when physically switching between C and A, or A and C. At first I though this was perhaps an issue with autofire. But, I just picked up on something similar in Rastan Saga, that does not involve autofire (see below).

Rastan Saga
A = slash, B = jump, C = UP + B. Use C to jump onto a rope (to do a bigger jump) while up is pressed. Notice that even though up was help the whole time, your character does not continue to move up the rope. Much like the issue I describe with R-Type, you can't move up the rope unless you release all buttons / directions then press up again. It's only C that causes this, if you do a big jump normally (hold UP + B) you continue to travel up the rope as normal.

I tested these exact scenarios on MAME, and was not able to reproduce these issues. I am using an arcade stick that has a Brook Universal board in it, and used the same stick on both MiSTer and MAME.

The ability to combine inputs is extremely useful, so hope that it can be fixed to align with how it works in MAME.

deepthaw:
Fixes https://github.com/MiSTer-devel/Main_MiSTer/issues/824

This PR would change the handling of overlapping presses of inputs which are alternate inputs to the same button. Presently, if one of those inputs is being held, pressing the alternate input will cause a new press to be registered and the button will no longer be considered held - even though the physical input still is.

This differs from how physical hardware as well as many emulators such as MAME handle the situation. In physical hardware, if two buttons were wired up to the same signal, holding one would close the circuit which means the other would have no impact on the signal. Presently, if two inputs (FireA and FireB) are both mapped to act as fire, pressing FireB while FireA is being held will register as a new button press, and the fact that FireA is being held is "lost."

This behavior is most noticeable in games (typically shooters like R-Type) in which a single fire button acts differently between being held (typically charges a special shot) or being tapped (acts as a rapid fire.) Some arcade cabinets would have a second physical fire button with an autofire circuit added as a player convenience, a feature that has carried into many home ports and emulators.

Typically a user holds in the autofire button, then presses and holds the fire button when they wish to begin charging a special shot. There is no need to release the autofire button, as the fire button being pressed keeps the circuit closed. Once they release the fire button, the charged shot is performed and it immediately returns to autofiring (if the autofire button is still being held.) There is no need to delay between inputs and the transitions happen immediately. This convention has been codified in later CAVE shooters and others where the cabinet would come with a separate autofire button which interacted with the fire button in this way.

I have implemented the fix in the following way: the masks that track joystick and button (and autofire) states has been increased from 32 to 64 bits. The upper 32 bits now store the state of the alternate button mappings, while the lower 32 bits store the state of the primary button mappings. This assignment occurs in input_cb(). The two bitmasks are ORed together in user_io_digital_joystick() before being sent to the FPGA.

I have tested it both with software (MiSTer) and physical autofire, as well as had several members of the arcade shooter (shmup) community test it and it seems to be performing as intended. If any code still writes directly to the bitmask, it will write to the lower 32-bits, which is the primary mappings and should continue to function as before.

### DIFF
--- a/input.cpp
+++ b/input.cpp
@@ -1643,8 +1643,8 @@ static int keyrah_trans(int key, int press)
 static void input_cb(struct input_event *ev, struct input_absinfo *absinfo, int dev);
 
 static int kbd_toggle = 0;
-static uint32_t joy[NUMPLAYERS] = {};
-static uint32_t autofire[NUMPLAYERS] = {};
+static uint64_t joy[NUMPLAYERS] = {};		// 0-31 primary mappings, 32-64 alternate
+static uint64_t autofire[NUMPLAYERS] = {};	// 0-31 primary mappings, 32-64 alternate
 static uint32_t autofirecodes[NUMPLAYERS][BTN_NUM] = {};
 static int af_delay[NUMPLAYERS] = {};
 
@@ -1840,12 +1840,12 @@ static void joy_apply_deadzone(int* x, int* y, const devInput* dev, const int st
 }
 
 static uint32_t osdbtn = 0;
-static void joy_digital(int jnum, uint32_t mask, uint32_t code, char press, int bnum, int dont_save = 0)
+static void joy_digital(int jnum, uint64_t mask, uint32_t code, char press, int bnum, int dont_save = 0)
 {
 	static char str[128];
-	static uint32_t lastcode[NUMPLAYERS], lastmask[NUMPLAYERS];
+	static uint32_t lastcode[NUMPLAYERS];
+	static uint64_t lastmask[NUMPLAYERS];
 	int num = jnum - 1;
-
 	if (num < NUMPLAYERS)
 	{
 		if (jnum)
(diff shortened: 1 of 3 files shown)


deepthaw:
I'm bad at github - this should reference https://github.com/MiSTer-devel/Main_MiSTer/issues/824

deepthaw:
TLDR;

assume we're playing R-Type and A and C are both mapped to fire. C has turbo enabled.

current behavior: While holding C, autofire goes as expected. Pressing and holding A does nothing because it's "held" status is immediately nullified by C getting pressed again via autofire. Holding A by itself causes a charge shoot to begin. Pressing C while A is being held causes the the charge to immediately be cancelled, and autofire resumes.

new behavior: While holding C, autofire goes as expected. Pressing and holding A causes the autofire to be suppressed, and the special shot begins charging. Once A is released, the charged shot is released and autofire resumes immediately. Pressing C while A is being held does nothing, and the charge shot continues to charge.

PR #914, opened 2024-08-27
chore(ci): add simple build workflow

dnwe:
Add a simple GitHub Action workflow which builds (and caches) the .devcontainer docker image and then runs a `make` of the MiSTer binary within it before uploading it as an artifact on the workflow result.

This provides a simple compilation test on pull requests and also allows the author or others to grab a built binary for testing purposes

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,8 +1,8 @@
 FROM ubuntu:22.04
+
 RUN apt-get update && apt-get upgrade -y && \
     apt-get install -y make wget xz-utils && \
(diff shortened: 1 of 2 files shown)


sorgelig:
It's not a good idea to make a binary with every change in this repo.
There is unstable/nightly builds which is fine for those who want to test latest changes.

dnwe:
To be clear, the binary is only attached to the workflow result, it isn’t added to the repo or anything. It is mostly useful as confirmation that when someone has raised a PR their changes do compile. However, a side benefit is (for example) if someone submits a pull request to add support for a new steering wheel in input.cpp there’s also a binary available, build in a clean env, that can be linked to owner(s) of that wheel to test and confirm the changes are good

PR #923, opened 2024-10-08
Fixed potential buffer overruns when adding too many .ini array elements.

steverobb:
There's no guard for writing off the end of int arrays in .ini files if too many elements are added.  This change ignores any extra elements, similar to what already happens with string arrays, and correctly sets the bounds of the `NO_MERGE_VIDPID` and `CONTROLLER_UNIQUE_MAPPING` arrays.


rsn8887:
This resulted in VID/PID pairs not being accepted unless the value was <256. If possible, I'd say please test your code next time.

PR #936, opened 2024-11-29
Build output goes into the bin folder.

steverobb:
This change puts the build output into a bin folder to make the workspace cleaner.


steverobb:
Any movement on this?

sorgelig:
I was thinking about this.
Ok

PR #939, opened 2024-12-13
Philips CD-i

Slamy:
This change adds support for the Philips CD-i core.
After a discussion with Kitrinx it seems to be desired to include changes like these.
Having multiple Main MiSTer binaries with the current "ini solution" is not so desired.
The changes are yet incomplete as the Table of Contents isn't evaluated yet. This is something I will add later on.

There are changes in this which are debatable.
I've shortened the "Saving..." time to 500ms. I can remove that if you want.

The CD-i support code is just a copy of the PSX code as both act very similiar on the FPGA side.

### DIFF
--- a/menu.cpp
+++ b/menu.cpp
@@ -7305,7 +7305,7 @@ int menu_allow_cfg_switch()
 
 void menu_process_save()
 {
-	menu_save_timer = GetTimer(1000);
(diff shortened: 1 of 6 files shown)


sorgelig:
Not sure about shortened time. Different people may have different reaction. I just wanted to make sure it's not missed.

PR #943, opened 2024-12-22
change zip sorting again

okdana:
i've just updated after several months and found that the 'zip folder' ux is suddenly frustrating. that's a consequence of #900, which changed the way zips are sorted so that they appear *in between* directories and other files

i don't really understand the purpose of that. @TpaBkaY said it's to better distinguish them from directories -- that much is obvious -- but why? afaik the mister ui treats zips as directories in almost every other respect, so what is the benefit to separating them out like this?

i don't mean to start a fight over it but i would actually propose that it work like this instead (see diff)

however if having the zips sorted differently has some benefit i'm not considering, maybe it could be an ini setting? it could even toggle the 'zips as directories' mechanic off entirely (sort them with regular files, auto-select the file inside instead of descending if it's the only one in the archive). would that be desirable?

### DIFF
--- a/file_io.cpp
+++ b/file_io.cpp
@@ -1224,12 +1224,6 @@ struct DirentComp
 		if ((de1.de.d_type == DT_DIR) && (de2.de.d_type != DT_DIR)) return true;
 		if ((de1.de.d_type != DT_DIR) && (de2.de.d_type == DT_DIR)) return false;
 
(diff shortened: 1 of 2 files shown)


birdybro:
I don't know of any other UX that treats zip files as directories in terms of sorting. Folders being at the top of the list always is consistent with basically every single UX for file navigation I've ever seen, and as such is probably the most intuitive design.

birdybro:
Technically windows does. You just double click a zip and it goes into it like a folder.

PR #947, opened 2025-01-06
Fix broken VID/PID and custom controller mapping parsing

rsn8887:
This reverts commit 59916ebc52aeb81b3b6fef93122ae8ecf82a0b62. Max is not an array size, it is the maximum allowed entry value.


rsn8887:
This reverts the commit that broke VID/PID and custom controller config parsing.

PR #948, opened 2025-01-07
Add decoding of VHD cookie, validation, and parsing

kdedon:
# Problem
I am attempting to install Openstep 4.2 but the first major issue is that the IDE driver suffers an overflow with larger SPT values. I would like to manually set the hard drive specifications to allow for the operating system to access the hard drive.

# Implementation
The VHD specification allows for hard drive metadata to be specified on a per file basis. I have added support files that provides the initial VHD footer parsing to allow for the metadata to be extracted for fixed size VHDs. I use this data in the hard drive loading path for pcxt and ao486 to allow for the file to define the CHS values.

I have created a support directory for this functionality in the case that VHD support is extended in the future to support DYNAMIC or DIFFERENTIAL types.

In the case that a VHD footer cannot be properly parsed, the functionality falls back to the current state of treating the file like a standard raw image.

### DIFF
--- a/support/vhd/mister_vhd.cpp
+++ b/support/vhd/mister_vhd.cpp
@@ -1,0 +1,91 @@
+#include <stddef.h>
+#include <string.h>
+#include <byteswap.h>
+
+#include "mister_vhd.h"
+
+vhd_error validate_vhd_checksum(vhd_footer *footer)
+{
+    uint32_t checksum = 0;
+    const uint8_t *fp = (const uint8_t *)footer;
+    uint16_t checksum_offset = offsetof(vhd_footer, checksum);
+
+    for (uint16_t i = 0; i < sizeof(vhd_footer); i++)
+    {
+        if (i >= checksum_offset && i < checksum_offset + sizeof(footer->checksum)) 
+        {
+            continue;
+        }
+        checksum += fp[i];
+    }
+    checksum = ~checksum;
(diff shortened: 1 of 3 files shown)


sorgelig:
I'm not sure if adding header at the end is a good idea.
May be it could be a better idea to add a text file with the same name where VHD parameters will be present? In that case, it will be easier to adjust/create than adding it in binary form.

sorgelig:
> A third option is to change ide_set_geometry to mirror calc_geometry in ide.cpp to attempt to fit the total size starting from lower values first. A bit more inefficient but it should only run once per file so it doesn't seem like that would be too big a waste.

it won't always work. You will attempt to tune the function to specific file but it will not work with many other files. So i prefer to keep CHS detection in current form (and keep it compatible with existing files) and add a config file in text form to support files where automatic CHS detection fails. Binary footer is too cryptic and requires HEX editor or other tools to edit.

So better to use a text form. It can be a simple text form like

```
sectors=XXX
heads=XXX
cylinders=XXX
```

I don't know how many real parameters are required, but better keep minimum for easier use. `cylinders` parameter is probably useful if you want to limit it in case if there is a "tail" at the end which may add unwanted additional cylinders. Probably "cylinders" can be used to extend the file on first use if original file is smaller.

I see you are motivated to do it, so it's ok if you will implement it.

PR #956, opened 2025-01-19
Fix asm operand has impossible constraints error in modern gcc versions

bootsector:
CPU registers r0...r3 are reserved to function parameters and more modern gcc versions will enforce that, making inline asm statements to fail if those registers are being used inline.


birdybro:
Is this possibly related to https://github.com/MiSTer-devel/Main_MiSTer/issues/958 ?

Just curious, I'm not an expert.

bootsector:
Different optimization options might use/allocate different registers inside the function. I recommend compiling the affected source file with the -S flag in order to generate the assembly output and see which registers are actually being used where and adjust the asm parameters accordingly so it compiles correctly for the intended optimization.

PR #968, opened 2025-02-16
Add Sinden Light Gun

terminator2k2:


sorgelig:
If i remember right, this gun requires white frame around the screen to work, so adding it in Main won't make it work.

terminator2k2:
yes requires a white frame , which will need alternative cores, but it saves having to keep a separate alternate repo for MiSTer Main

PR #972, opened 2025-03-09
Add CLI trigger UI widgets

wizzomafizzo:
Adds:

- File picker UI that can be triggered from CLI
- Notice message UI that can be triggered from CLI
- New file in tmp to advertise features are available
- Turn recents and log_file_entry on by default
- Adds Z to end of version string in info screen


wizzomafizzo:
@sorgelig Hi Sorg, I was originally not planning to make a PR for this, but I thought I would ask your opinion on it in any case. Is there any implementation of these features you would be willing to accept?

The intention of them is to let external processes have some true integration with the OSD. It lets an external process set up and display an on-screen list picker and show a text notice.

I have tried my best to make this work using only external scripts, but the result is prone to error managing conflicting linux processes, triggering the console display, and looks like garbage over analog video assuming the user has even configured it (which is confusing in itself). This OSD method is basically superior in every way, and has the added benefit of displaying the element without needing to quit an active core.

The implementation I've gone with is likely not the best way to do it, but it is (probably) the easiest for me to maintain if I had to keep it as a separate fork. If it actually were to go in, it would be excluding the ini default changes (they're personal nice to haves, not required for the feature), the version string change and the MAINFEATURES addition which are both just make it easier to detect if the fork is in use.

Based on your past feedback to requests for adding a way to show a notification on screen, you probably have a problem with potential conflicts of notices overwriting each other or interfering with an official OSD display. I don't have a solution for that, I can only say there are several areas conflicts like this happen on the linux side of MiSTer and people manage to make it work.

sorgelig:
I don't like to frustrate people who spend a time to implement something they think very useful, but in fact it's useful only for few people. 
1) MiSTer is managed directly by user input. There is no point to manipulate OSD outside of user direct input. What is the purpose to open the files outside of user interaction? It even may interfere the user while playing. 
I've already added MGL files to launch specific files at start and even more.
2) Changing default ini options because you like it is not acceptable. You may change it in your own INI if you don't like default state.

wizzomafizzo:
Please, let me assure you, I am not frustrated and this is basically what I expected you to respond with. Have a good one!

PR #1001, opened 2025-07-08
Virtua Racing sound fix - replays

Kuba-J:


srg320:
You have set the minimum seek time to 10 frames. What are your values based on?

PR #1009, opened 2025-07-23
Lookahead 3 positions when scrolling

misteraddons:
Scrolling from the very top/bottom row doesn't allow the user to see upcoming files. This modification keeps the cursor 3 positions from the top/bottom when scrolling. Works with page up/page down. Only allows the top/bottom 3 rows to be selected on the first/last page of the file list.

### DIFF
--- a/file_io.cpp
+++ b/file_io.cpp
@@ -1758,7 +1758,8 @@ int ScanDirectory(char* path, int mode, const char *extension, int options, cons
 			if(iSelectedEntry + 1 < flist_nDirEntries()) // scroll within visible items
 			{
 				iSelectedEntry++;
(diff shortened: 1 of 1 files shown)


sorgelig:
do we really need 3 lookahead positions? May be 1 would be enough?

misteraddons:
1 lookahead position is better than 0. Maybe 2 would be best? See what you think if you test it.

sorgelig:
May be add an option in INI where you can set 0-3, so if someone doesn't like it, then will be able to set to 0

sorgelig:
I think you forgot to use this parameter in the code.

misteraddons:
Thanks for the reminder. I tested 0-3 values of lookahead in MiSTer.ini. I also checked for out of bounds conditions (lookahead>3).

PR #1012, opened 2025-07-30
Automatic Direct Video Mode

misteraddons:
Automatically set Direct video based on HDMI EDID information. Useful for users who swap between HDMI and direct video.

- Use EDID information to automatically toggle direct_video.
  - Relies on generic HDMI resolution (1024x768@60) or whitelist of known DAC make and model.
    - Whitelist devices sets direct_video and hdmi_limited.
      - Continue adding new models to the whitelist
- Supports HDMI hot-plugging
- Adds two new direct_video settings:
  - direct_video=2: Auto-detect HDMI DAC, uses vga_mode and composite_sync settings
  - direct_video=3: Auto-detect HDMI DAC, forces vga_mode=rgb and composite_sync=1 for 15kHz RGB

- Known issues:
  - Menu background doesn't reliably re-scale when switching between HDMI and Direct Video
  - RetroTink 4K: MiSTer starts in Direct Video mode, but switches to standard HDMI after a few seconds

### DIFF
--- a/cfg.cpp
+++ b/cfg.cpp
@@ -64,7 +64,7 @@ static const ini_var_t ini_vars[] =
 	{ "FB_SIZE", (void*)(&(cfg.fb_size)), UINT8, 0, 4 },
 	{ "FB_TERMINAL", (void*)(&(cfg.fb_terminal)), UINT8, 0, 1 },
 	{ "OSD_TIMEOUT", (void*)(&(cfg.osd_timeout)), INT16, 0, 3600 },
-	{ "DIRECT_VIDEO", (void*)(&(cfg.direct_video)), UINT8, 0, 3 },
+	{ "DIRECT_VIDEO", (void*)(&(cfg.direct_video)), UINT8, 0, 2 },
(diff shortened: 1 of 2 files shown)


sorgelig:
Direct video is by nature a quirky mode. It's not a good idea to enforce it this way. It gives a wrong impression to the end user. It requires an experience user and specific hardware. And experienced user has enough knowledge to add direct_video=1 to INI file manually.

P.S.: looking at the added code - it looks over-complicated.

misteraddons:
I like the idea as an option (not the default). It works well in practice - the system reliably detects and enables direct_video when an HDMI DAC is connected. I removed direct_video=3 and simplified a few other things with the code.

To me, worst-case scenario is that the user has to revert to direct_video=1. As we gather more HDMI DAC EDID info, we can white list more devices. The AG6200 and CS5213 already cover some of the most popular DACs.

Open to feedback and thanks for evaluating :)

misteraddons:
Will submit a simpler change

PR #1021, opened 2025-08-17
Change composite_sync default value to 1

TheJesusFish:
From a troubleshooting perspective, composite_sync=1 seems like a more common gotcha because more people are using consumer CRTs than monitors.

I'm wondering if it makes sense to make this change as it covers more use cases.

Sorry for the double PR - I didn't realize I had to change the value in cfg.cpp as well.

### DIFF
--- a/MiSTer.ini
+++ b/MiSTer.ini
@@ -5,7 +5,7 @@ forced_scandoubler=0   ; set to 1 to run scandoubler on VGA output always (depen
 ;ypbpr=0               ; set to 1 for YPbPr on VGA output. (obsolete. see vga_mode)
 vga_mode=rgb           ; supported modes: rgb, ypbpr, svideo, cvbs. rgb is default.
 ntsc_mode=0            ; Only for S-Video and CVBS vga_mode. 0 - normal NTSC, 1 - PAL-60, 2 - PAL-M.
-composite_sync=0       ; set to 1 for composite sync on HSync signal of VGA output.
+composite_sync=1       ; set to 1 for composite sync on HSync signal of VGA output.
 vga_scaler=0           ; set to 1 to connect VGA to scaler output.
 hdmi_audio_96k=0       ; set to 1 for 96khz/16bit HDMI audio (48khz/16bit otherwise)
 keyrah_mode=0x18d80002 ; VIDPID of keyrah for special code translation (0x23418037 for Arduino Micro)
@@ -371,3 +371,4 @@ osd_lock_time=5
 
 ; use custom main for specific core. This option should be used only inside specific core.
 ;main=some_binary_file
+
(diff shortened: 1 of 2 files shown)


birdybro:
This should make it so computer CRT's work out of the box with VGA via RGBS. RGBHV won't work by default with this on only some monitors, but that didn't work by default without tweaking some settings anyways. I personally think this is a good change that in the long run will reduce troubleshooting issues.

sorgelig:
I don't have a preference on this option as i seldom use analog output. 
May be need to hear more opinions. Generally speaking monitors expect RGBHV, not RGBS so i think many monitors won't work with composite sync. TVs aren't so versatile in supporting different refresh rates so even RGBS won't make it compatible with many TVs.

misteraddons:
I fully support making composite_sync=1 the default setting. 

For signal types: YPbPr, S-Video, and Composite video all require composite sync. YPbPr can also use the vga_sog=1 setting but composite_sync=1 is sometimes required depending on the IO board type.

For display types:
PC CRT: many support composite sync (mostly mid-1990s and newer), some require separate sync
Consumer CRTs: require composite sync
Professional monitors (PVM/BVM): require composite sync
Arcade monitors: require composite sync
External video upscales: require composite sync

So really, aside from select PC CRTs, all devices and signal types prefer composite sync. This is often the only MiSTer.ini setting that users need to change. In my experience, most analog users are connecting to professional monitors or consumer CRTs.

miklosaubert:
I'm not a contributor, just an enthusiastic user, but I would like to add my voice in favor of this pull request if it can help.

I mainly use my MiSTer with a CRT TV, but I also have an LCD computer monitor with VGA in, and a fairly old Samsung flat panel TV that also accepts VGA. I tested both the monitor and the Samsung TV with minimal MiSTer.ini files containing only the below lines:

| 1st test    | 2nd test | 3rd test | 4th test |
| ------------- | -------------  | ------------- | ------------- |
| vga_scaler=1 | vga_scaler=1 | vga_scaler=1 | vga_scaler=1 |
|  | composite_sync=1 | | composite_sync=1 |
| | | video_mode=8 | video_mode=8 |

I got the menu to show up, for all tests. One thing to note is that with `composite_sync` unset, both the monitor and the TV went through an extra auto-adjustment cycle, with an OSD message coming from the device saying something like "auto adjustment in progress...". The image came up fine and I don't think the auto-adjustment was really needed, but I guess that's just a feature of these screens when they detect an RGBHV signal.

Of course, this is all anecdotal, with only 2 screens tested. But overall, I think having the default value of `composite_sync=1` would simplify the first-time setup for many users, and it doesn't seem to be an issue with VGA displays (at least, not with mine)

PR #1027, opened 2025-08-26
Add volume button support for GRS Ultimate Control Panel (iiRcade + Viper KVM)

HonkeyKong:
The MiSTer itself works great with the Viper KVM switcher for iiRcade, however when in XInput mode (the default on the GRS Ultimate Control Panel through the KVM), the volume buttons don't work. This fixes that issue.

### DIFF
--- a/input.cpp
+++ b/input.cpp
@@ -5368,6 +5368,16 @@ int input_test(int getchar)
 									}
 								}
 
+								// Volume controls for GRS Ultimate Deck for iiRcade + Viper KVM (XInput Mode)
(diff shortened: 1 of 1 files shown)


HonkeyKong:
It just occurred to me that this may need to be more specific, as it could cause conflicts with wired Xbox 360 controllers due to the cloned VID/PID. I'll do some more testing with some 360 controllers and see what I can find.

HonkeyKong:
Looks like axis 2 is right stick X, and axis 5 is the right trigger. That won't do without some specific casing. I'll close this one and work on breaking it out into its own option.

PR #1029, opened 2025-08-26
Add optional ABS axis to volume mapping via MiSTer.ini

HonkeyKong:
Some controllers/encoders (e.g. GRS Ultimate Control Deck for iiRcade + Viper KVM)
expose their panel volume buttons as EV_ABS axis events instead of joystick buttons
or consumer keys. In XInput mode this previously meant volume controls were not
usable.

This patch introduces four new MiSTer.ini keys:

```
  volume_abs_vid   = <USB vendor id>
  volume_abs_pid   = <USB product id>
  volume_abs_down  = <ABS code for Volume Down>
  volume_abs_up    = <ABS code for Volume Up>
```

When defined, MiSTer will map those ABS events to set_volume(-1/+1).

Default is disabled (0 / -1), so normal Xbox 360 pads and other controllers are
not affected.

### DIFF
--- a/cfg.cpp
+++ b/cfg.cpp
@@ -134,6 +134,10 @@ static const ini_var_t ini_vars[] =
 	{ "LOOKAHEAD", (void *)(&(cfg.lookahead)), UINT8, 0, 3 },
 	{ "MAIN", (void*)(&(cfg.main)), STRING, 0, sizeof(cfg.main) - 1 },
 	{"VFILTER_INTERLACE_DEFAULT", (void*)(&(cfg.vfilter_interlace_default)), STRING, 0, sizeof(cfg.vfilter_interlace_default) - 1 },
+	{ "VOLUME_ABS_VID", (void*)(&(cfg.volume_abs_vid)), HEX16, 0, 0xFFFF },
+	{ "VOLUME_ABS_PID", (void*)(&(cfg.volume_abs_pid)), HEX16, 0, 0xFFFF },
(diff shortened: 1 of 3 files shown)


sorgelig:
why it needs to introduce a new ini options? If it's a quirk of specific device then hardcoded VID/PID should be enough as in your first commit.

HonkeyKong:
Because the VID/PID are identical to original Xbox 360 controllers. My initial commit had it this way, and then I thought to go back and check with a real 360 controller, only to find that the axes mapped to volume controls were the right stick X axis, and the right trigger. 

Adding the INI options allows the GRS Ultimate Control Deck + Viper KVM to use its volume controls and is constructed in such a way that it can be used for other oddball joystick encoders as well. I'm open to other ways of doing this if you have suggestions though.

sorgelig:
Why not just let it be as it is? Volume control isn't the main functionality required to be available all the time. Volume can be controlled through OSD.
I just don't fee it right to add so much attention to badly designed controller. Especially it seems cannot even succeed the kickstarter campaign.

HonkeyKong:
That's a fair enough point, I guess. It had two production runs, but still, few enough people own it that if they want that feature, they can patch it themselves.

wickerwaka:
> Because the VID/PID are identical to original Xbox 360 controllers. My initial commit had it this way, and then I thought to go back and check with a real 360 controller, only to find that the axes mapped to volume controls were the right stick X axis, and the right trigger. 

The device name is probably different and unique, you can use that along with the vid/pid to identify this specific device.

RyanBashGaming:
Unfortunately, the device name is "Xbox 360 Controller".

PR #1034, opened 2025-09-11
Skipped destination bytes in maps parsed correctly. Fixes #1033

### LINKED ISSUE #1033: map in interleave not parsed correctly
This MRA expression is not parsed correctly:

```
<interleave output="32">
    <part name="061d14.12l" crc="b9440924" map="0201"/>
    <part name="061d13.12k" crc="9f999f0b" map="2010"/>
</interleave>
```

The bytes do not get mapped correctly. I built a small test file [maptest.cpp](https://github.com/user-attachments/files/22247193/maptest.cpp) that calls the rom_data function. You can run it directly as it is selfcontained. 

It tries to use map=0201 on a buffer that just goes `buf[8]={0,1,2,3,4,5,6,7}`. The output it produces is:

```
0 1 0 0 2 3 0 0 - got
0 0 1 0 2 0 3 0 - expected
```

Showing that the `map="0201"` is not parsed correctly.

jotego:
Adds the capability to skip bytes in the destination map. Like `map="0201"` or `map="2001"`

The following combinations have been tested against a buffer containing `01234567`, the first number is the map and the string is the expected result. All these tests pass.

```
unitlen=2;
test_map(0x01,"00102030");
test_map(0x10,"00010203");
test_map(0x21,"01234567");
test_map(0x12,"10325476");
unitlen=4;
test_map(0x0001,"00001000");
test_map(0x4321,"01234567");
test_map(0x0012,"10003200");
test_map(0x1234,"32107654");
test_map(0x1200,"00100032");
test_map(0x2100,"00010023");
test_map(0x0021,"01002300");
test_map(0x0201,"00102030");
test_map(0x2001,"00012003");
unitlen=8;
test_map(0x0000'0001,"00000000");
test_map(0x0000'0021,"01000000");
test_map(0x0000'4321,"01230000");
test_map(0x8765'4321,"01234567");
test_map(0x0043'0021,"01002300");
test_map(0x4300'0021,"01000023");
test_map(0x0403'0201,"00102030");
```
The test is in a different [branch](https://github.com/jotego/Main_MiSTer/tree/romdata/support/arcade) that you can check for reference. I do not see tests in the original repository so I am not including the test in the PR.

### DIFF
--- a/support/arcade/mra_loader.cpp
+++ b/support/arcade/mra_loader.cpp
@@ -250,12 +250,19 @@ static int rom_data(const uint8_t *buf, int chunk, int map, struct MD5Context *m
 		return 0;
 
 	map_reg = map;
+	bool first = true;
+	int gaps = 0;
 	for (int i = 0; i < unitlen; i++)
 	{
 		if (map_reg & 0xf)
 		{
-			offsets[bytes_in_iter] = idx + (map_reg & 0xf) - 1;
+			offsets[bytes_in_iter] = idx + (map_reg & 0xf) - 1 + gaps;
 			bytes_in_iter++;
+			first = false;
+		}
+		else if(!first)
+		{
+			gaps++;
 		}
 		map_reg >>= 4;
 	}


zakk4223:
I think this may have broken some ST-V games that have maps like this: 
```
<interleave output="64">
			<part crc="2a5a8c3a" name="350-mpa1.u19" map="01030507" />
			<part crc="735e23ab" name="350-mpa1.u16" map="10305070" />
</interleave>
```

Memory layout with this commit:
```
34000000: 00 00 00 54 2d 53 41 45 00 00 00 4e 29 41 49 28    ...T-SAE...N)AI(
34000010: 00 00 00 4c 41 43 41 43 00 00 00 20 20 20 20 4d    ...LACAC...    M
34000020: 00 00 00 20 20 20 20 20 00 00 00 30 30 30 31 20    ...     ...0001 
34000030: 00 00 00 30 37 35 36 39 00 00 00 20 20 41 54 41    ...07569...  ATA
34000040: 00 00 00 20 20 20 20 55 00 00 00 20 20 20 20 20    ...    U...     
34000050: 00 00 00 20 20 20 20 20 00 00 00 20 20 20 20 20    ...     ...     
34000060: 00 00 00 20 46 4e 4d 41 00 00 00 20 20 52 56 52    ... FNMA...  RVR
34000070: 00 00 00 20 46 4e 4d 41 00 00 00 20 20 52 56 52    ... FNMA...  RVR
34000080: 00 00 00 20 46 4e 4d 41 00 00 00 20 20 52 56 52    ... FNMA...  RVR
```

Previous memory layout:
```
34000000: 54 2d 20 53 47 41 53 45 4e 29 54 41 54 49 56 28    T- SGASEN)TATIV(
34000010: 4c 41 43 43 29 41 28 43 20 20 20 20 20 20 49 4d    LACC)A(C      IM
34000020: 20 20 20 20 20 20 20 20 30 30 2e 30 56 31 20 20            00.0V1  
34000030: 30 37 30 35 39 36 31 39 20 20 2d 41 52 54 43 41    07059619  -ARTCA
34000040: 20 20 20 20 45 20 4a 55 20 20 20 20 20 20 20 20        E JU        
34000050: 20 20 20 20 20 20 20 20 20 20 20 20 20 20 20 20                    
34000060: 20 46 41 4e 54 4d 42 41 20 20 45 52 45 56 4f 52     FANTMBA  EREVOR
34000070: 20 46 41 4e 54 4d 42 41 20 20 45 52 45 56 4f 52     FANTMBA  EREVOR
34000080: 20 46 41 4e 54 4d 42 41 20 20 45 52 45 56 4f 52     FANTMBA  EREVOR
```

jotego:
I am afraid the MRA for those games may have been working because of a fluke. I tested the original code -prior to this PR- with the pattern:

```
buf[8]={1,2,3,4,5,6,7,8};
test_map(0x7050'3010,"01030507");
```
And the result was:
```
01020304 - got
01030507 - expected
FAIL
```
The map is asking to skip bytes in the input data sequence, but the code cannot do that. The buffer is always fully read:
``` C++
while (chunk)
{
	for (int i = 0; i < bytes_in_iter; i++)
	{
		*(romdata + romlen[idx] + offsets[i]) = *buf++;
		chunk--;
	}
	romlen[idx] += unitlen;
}
```
See how `*buf++` always goes up regardless of the map. The map only tells you where the byte will be placed. Valid maps cannot skip bytes, you cannot go 1,3,5. It must be 1,2,3. 

I think that the map you need to use might be this one:
``` XML
<interleave output="16">
  <part crc="2a5a8c3a" name="350-mpa1.u19" map="01" />
  <part crc="735e23ab" name="350-mpa1.u16" map="10" />
</interleave>
```
This will merge the two files taking one byte from each.

zakk4223:
The old code treated "01030507" and "1357" as equivalent since it just ignored zero elements in the map string unless they occurred on the right hand side. 

Looks like the map I need is actually:
```
<interleave output="16">
  <part crc="2a5a8c3a" name="350-mpa1.u19" map="1357" />
  <part crc="735e23ab" name="350-mpa1.u16" map="13570" />
</interleave>
```

which results in the pattern:
```
a3 b3 a2 b2 a1 b1 a0 b0
```

I think ST-V is the only core that was using the 'gap' style pattern so I'll just re-generate those MRAs since that pattern works on both old and new.

jotego:
Thanks, @zakk4223 for looking at options. However, note that the map you are using is still ilegal and hence the results may change the next time the code is updated.

The numbers in the map string refer to which bytes are used from the file. So a `map="1357"` means that bytes 1,3,5 and 7 should be used. But **that expression is not supported**. Valid maps cannot skip bytes from the origin file. You need to write in a sequence like `map="1234"`, which means use all bytes from 1 to 4 in order.

The total number of bytes used inside the interleave multiplied by 8 is what goes in the `output="16"` part. 

There are literally over a thousand valid MRA files in my [JTBIN repository](https://github.com/jotego/jtbin/tree/master/mra) you can check out as a reference. You will find examples of pretty much anything possible there.

zakk4223:
I don't read map="1357" meaning that at all, at least not how the code was written. That still uses bytes 1 2 3 and 4 from the input file, it just writes them into different positions in the "output=X" sized block of output. (In this case the resulting output for 64bit wide map is "4 X 3 X 2 X 1 X" Then bytes 5,6,7,8 are written into the same positions in the NEXT block of output. 

I see no reason there should be a requirement that the map strings be contiguous sequences. All they're expressing is what position to write into the current output block. 

Or to put it another way, and how I've always read this code's intention:

The POSITION of the number in the map string refers to the input byte position.
The VALUE of the number refers to the index to write it into the current output chunk. (Controlled by output=X)


Every non-zero value in the map string "consumes" an input byte and writes it to the position (value-1). 
*Starting from the rightmost side*

This allows controlling how many bytes of a file go into each output block and where they go. 
It provides for being able to 'explode' files into patterns like "2 X X X X X X 1 4 X X X X X 3" etc
Trailing/right side  zeros shift the pattern by one byte for each zero. (and it controls the "base" starting offset for the file, but that's a bit different than what we're talking about here).

I think if you have a file with contents 12345678 and you want this pattern:
```
4 0 3 0 2 0 1 0   8 0 7 0 6 0 5 0
```

The only way to express that is output="64" and map="1357" (at least in the old way, I haven't looked at how with this change).

jotego:
I have always understood it the other way around:

1. the position indicates the output byte
2. the value indicates the input byte
3. the interleave tells you how many bits must be collected in each round

Point 3 implies that you cannot fill the output buffer with zeros in order to make a part look larger. You always need to merge it with a different part in order to produce the total number of bytes required by the interleave.

I do not mean to be offensive but I honestly think that the ST-V files are working by a fluke. 

I think it is easier to agree with explicit test cases. The old code passes the following tests:

```C++
unitlen=16/8;
test_map(0x01,"10203040");
test_map(0x10,"01020304");
test_map(0x21,"12345678");
test_map(0x12,"21436587");
unitlen=32/8;
test_map(0x0001,"10002000");
test_map(0x4321,"12345678");
test_map(0x0012,"21004300");
test_map(0x1234,"43218765");
test_map(0x1200,"00210043");
test_map(0x2100,"00120034");
test_map(0x0021,"12003400");
unitlen=64/8;
test_map(0x0000'0001,"10000000");
test_map(0x0000'0021,"12000000");
test_map(0x0000'4321,"12340000");
test_map(0x8765'4321,"12345678");
test_map(0x0403'0201,"10203040");
```

The new code will pass all the above plus these new ones:

```C++
unitlen=32/8;
test_map(0x0201,"10203040");
test_map(0x2001,"10023004");
unitlen=64/8;
test_map(0x0043'0021,"12003400");
test_map(0x4300'0021,"12000034");
test_map(0x0403'0201,"10203040");
```

You can test this in the [branch](https://github.com/jotego/Main_MiSTer/tree/romdata/support/arcade). Just go to the folder `support/arcade` and run `run_tests.sh`

See if this makes sense to you and what else you would like to see tested and passing.

PR #1035, opened 2025-09-11
Automatic Direct Video

misteraddons:
- This is a simple patch that checks for 1024x768 EDID resolution and enables direct video
- cfg.cpp setting added: direct_video=2
- This mode is intended for simple HDMI DACs, not for upscalers.
- hdmi_limited is set for known HDMI DACs (AG6200 and CS5213 so far) by looking up the mfg_id in EDID bytes 0x08-0x09
- Requires power cycle when switching between HDMI and HDMI DACs.

### DIFF
--- a/dv_dac.txt
+++ b/dv_dac.txt
@@ -1,0 +1,24 @@
+# Direct Video DAC Configuration File
+# Format: mfg_id,hdmi_limited,hdmi_audio_96k,composite_sync,name
+# 
+# mfg_id: Manufacturer ID from EDID bytes 0x08-0x09 (hex, e.g. 0x48F4)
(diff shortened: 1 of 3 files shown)


sorgelig:
With direct_video=2 it's acceptable in general. But why you assume that any 1024x768 is HDMI-DAC? Since you already know VID of known DACs then enable direct video only for them.

misteraddons:
That's a great idea. Since we're able to make the DAC detection granular, would you consider making direct_video=2 the default mode? It nicely defaults to preferred EDID resolution for modern displays, and only enables direct video for narrowly defined devices.

sorgelig:
It's better to not enable it as default because a) it's not always direct video is preferable option b) it's not for novice user as it's required a special display/scaler.

PR #1042, opened 2025-09-28
Add Gamate to yc.txt

TheJesusFish:
[Insert joke about Gamate here]


MikeS11:
> [Insert joke about Gamate here]

I'll add that in a couple weeks, and look for more that might be missing or needs updating..

In the meantime, if anyone needs it: GAMATE_60.8=221958035000

PR #1045, opened 2025-10-04
Write unique game info to /tmp/GAMENAME

misteraddons:
This is the game equivalent of writing the current core to /tmp/CORENAME. 

The idea was inspired by Robert Peip's PSX and N64 cores, which identify internal titles for games.

This new function writes the path, game_id (if applicable) and crc32 (if applicable) to /tmp/GAMENAME.
`user_io_write_gamename(const char *path, const char *game_id, uint32_t crc32_val);`

- Cartridge games are identified already identified by crc32
  - Special cases
    - N64 - crc32 and internal title (already done by Robert Peip)
    - NeoGeo - internal title instead of crc32
- Disc games are identified in unique ways (internal title)
  - Tested
    - MegaCD - internal title
    - NeoGeo CD - disc author date + internal title
    - CD-I - internal title
    - PSX - internal title (already done by Robert Peip)
    - Saturn - internal title
    - TGFX16-CD - md5 hash of first 100 sectors
- MGL files work as expected
- MRA files export the mame setname as game_id
- Computer support may vary

Inspired by @niemasd's [GameID](https://github.com/niemasd/GameID) project.


sorgelig:
Looking at the amount of code added with a lot of heuristics, it doesn't look as robust feature.

misteraddons:
Ok, let me look into it some more. The idea is for each game to be uniquely identified. Do you have specific things you don't like? I'm thinking the best way to ID a CD game is the product code (internal title) combined with a crc32 of the first 100 sectors. Combined, that takes 50-100ms. What do you think?

sorgelig:
What is the practical use of this? For consoles, games can be distinguished by either cartridge CRC or CD crc already.

misteraddons:
What part of the cd image is already CRC32'd, and does it work across bin/cue and chd formats? The purpose is to advertise the current game along with the current core for scripts and third party utilities to display metadata.

We agree that cartridges are easy. Maybe we start with file path and cartridge CRC32?

I'm mostly following Robert's idea with the PSX and N64 core, both of which extract product code. I know the N64 core uses the local database to determine save time and region. In the future, could main set the region for cores which don't have auto region?

sorgelig:
> Maybe we start with file path and cartridge CRC32?

it's a;ready done. That's how cheat engine works.

birdybro:
> In the future, could main set the region for cores which don't have auto region?

Usually cores that don't have auto region are because there is no identifier of what the region should be in the ROM itself typically (or if there was in the manufacturer's spec, it wasn't consistently used). Or it isn't useful for a region swap (NTSC/PAL) like on the Sega Master System where there is only Japan or Export or International, none of which designate if it should be NTSC or PAL (source: https://www.smspower.org/Development/ROMHeader).

So I think that would be the practical benefit of using such a database, systems where the NTSC/PAL switch is not currently automatic due to missing metadata in the ROM spec.

The other benefit would potentially being a refactor so we don't have multiple different redundant instances of the same kind of auto-region switching in different forms either in Main or in the core HDL.

misteraddons:
What’s the minimal code needed to reliably identify CD games?

Current CD Core Behavior:
- PSX: Uses product code (from disc) or filename for cheats.
- TurboGrafx-CD & Mega CD: Cheats use filename only; no product code or CRC32.
- Saturn, CD-i (WIP), Neo Geo CD: No cheats or product code support yet.

Current Game Identification Use Cases:
- N64: Uses N64-database.txt to set region, save type, etc.
- PSX: Uses region to select the correct BIOS

Future Game Identification Use Cases:
- TurboGrafx-CD: Auto-enable Arcade Card.
- Saturn Auto RAM cart/region settings - I need to check if Saturn still has region specific boot roms.
- Saturn/CD-i: Game ID might help cheat support (CD-i has 2 cheats haha).

Using CRC32 on the first 100 sectors, combined with product code extraction is fast, deterministic, and works with both .bin/.cue and .chd formats. 

Example ID timing (first 100 sectors):
```
Saturn Product Code: T14310G (0 ms)
Saturn CRC32: 300C0682 (4 ms)
```

Some platforms (e.g., TurboGrafx-CD) lack product codes entirely, and others may share or miss them—making CRC32 useful for differentiation. While CRC32 could be a universal fallback, product codes remain the standard.

For /tmp/GAMENAME, we could start with filename + CRC32 or product code where available.

Want to see the updated code with a common CRC32 sector function and per-core product code extraction code?

sorgelig:
Only N64 is awkward in this case. It needs a game ID to setup myriads settings. Other cores use some reliable ID or CRC32 to support cheats (which has more relaxed requirement).
There is no reliable way to identify a game name for every core, so i suggest to give up such idea and do it core by core if it really required (such as cheats).

misteraddons:
I removed all the product code. Now it's just file path and crc32 for systems that already have that functionality.

sorgelig:
There is already tracking info stored in `/tmp/CURRENTPATH`, `/tmp/FULLPATH`, `/tmp/FILESELECT`. Your addition is kind of doubling it.

misteraddons:
You're right! I forgot about the `log_file_entry=1` option. I updated the pull request to only write CRC32 and Product Codes (N64 and PSX only since no other system has support for this) to /tmp/GAMEID, excluding bios files (boot*.rom and cd_bios*.rom).

sorgelig:
Finally we came to some adequate change :)
GAMEID should obey `log_file_entry` option, same as other tracking files. Also it's better to remove excessive tracking info except failure.

misteraddons:
Haha finally! I think it's ready for testing and inclusion. I only print a single line for successful GAMEID:
`Game ID: Soldier Blade (USA).pce [CRC32: 4BB68B13]`

Games without crc32 or serial don't print anything to stdout.

misteraddons:
I added MRA and MGL support because I didn't see support for those previously. Please provide feedback if you see a better way to handle these.

misteraddons:
Any feedback?

PR #1046, opened 2025-10-10
PSX: Persistent noreset.txt state until core is reset or reloaded.

Bemjo:
The current PSX core behaviour is to reset upon loading a CD image in a different directory from the previously loaded CD image. A magic file called noreset.txt can be placed beside a CD image file in it's directory to override this behaviour and force it to not reset upon loading a new disc outside of it's directory. This type of functionality is used for a few games, notably Monster Rancher 1 & 2, and Vib Ribbon. 

However in the case of Monster Rancher 1 & 2 especially, unless the swapped in disc also has a noreset.txt file present in it's directory, you cannot swap back to the original Monster Rancher disc which is required to continue to play the game, without having to do an intermediary step of going into the Miscellaneous menu, and opening the CD lid, inserting a new cd, then closing the lid, continuing with the game, opening the lid, inserting the original disc again, and closing the lid. Trying to swap back to the original disc just resets the core, as the noreset.txt file is checked against the last loaded disc image every time a new disc is loaded, which is extremely likely not to exist in the typical use case of these games.

This change is to improve the user experience for the select few games that use this disc swapping mechanism by indefinitely holding the noreset state when such a game is loaded, allowing the user to easily and quickly swap between discs from the Load CD OSD option on the PSX core main menu and not having to juggle the lid state through the miscellaneous menu. This state persists until the user either reloads the core entirely, or uses the Reset OSD option.

### DIFF
--- a/menu.cpp
+++ b/menu.cpp
@@ -2349,6 +2349,7 @@ void HandleUI(void)
 									if (is_pce() && !bit) pcecd_reset();
 									if (is_saturn() && !bit) saturn_reset();
 									if (is_n64() && !bit) n64_reset();
+									if (is_psx() && !bit) psx_reset();
 
 									user_io_status_set(opt, 1, ex);
 									user_io_status_set(opt, 0, ex);
--- a/support/psx/psx.cpp
+++ b/support/psx/psx.cpp
@@ -18,6 +18,7 @@
 static char buf[1024];
 static uint8_t *chd_hunkbuf = NULL;
 static int chd_hunknum;
+static int noreset = 0;
 
 static int sgets(char *out, int sz, char **in)
 {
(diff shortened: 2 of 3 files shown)


sorgelig:
I don't want to interfere here but as far as i understood, once noreset.txt is discovered, core stops rebooting. So if user would like to load another game it will be treated as additional disk of the same game. This may impact user experience as usually people forget about such hidden switches. 
Anyway, if @RobertPeip is ok with such change then i'm fine too.

RobertPeip:
Yes, we discussed in Discord about it before.

Issue is currently not with games that have multiple disks, those wouldn't need it.
It's only for games that have special feature to load completly different disks. Noreset.txt will allow you to load these, but as soon as you back to the previous CD, core reboots.

I think it's fine. If someone really adds Noreset.txt to a directory, they want this behavior. It's hidden enough that only the ones experienced enough to add the file in the first place will encounter this behavior.

PR #1049, opened 2025-10-13
Make subcarrier a vga_mode option instead of standalone

misteraddons:
This makes more sense since it will automatically set composite sync, rgb, etc.


sorgelig:
RGB mode doesn't imply composite sync. There is still H/V sync mode.

misteraddons:
I agree, but subcarrier mode requires composite sync and uses the vsync
line to send the chroma subcarrier signal. Unless I’m misunderstanding,
composite sync is only set when vga_mode=subcarrier, not when vga_mode=rgb.

sorgelig:
ok, i see now

PR #1051, opened 2025-10-23
Update MiSTer.ini

kaloun34:
subcarrier : unknown option


misteraddons:
Thank you for testing and catching this issue. I forgot to remove the superfluous option when I updated MiSTer.ini.

PR #1054, opened 2025-10-28
Enforce required settings for vga_mode=ypbpr/svideo/cvbs

misteraddons:
`rgb` isn't affected
`ypbpr` enforces composite_sync and vga_sog
`s-video` and `cvbs` enforces 15kHz resolution and composite_sync
`subcarrier` is updated to disable vga_scaler


misteraddons:
Tested with and without direct_video to confirm no regressions

sorgelig:
This is kind of redundant. parameters csync=1, vga_scaler=0, scandoubler=0 are already default.
I don't see the point to enforce it which will provide more confusions to those whole would like to use other values of above parameters even if some combinations won't work.

misteraddons:
The default settings are much better now with composite_sync=1.

What’s the advantage of allowing the user to set incompatible ini options?
There’s only one valid setting for subcarrier, svideo, and cvbs.

YPbPr can be scandoubled or upscaled so I only forced sync settings
and I didn’t
change RGB at all since it has multiple sync configurations.

I tested each vga_mode with and without direct_video by adding invalid
settings to MiSTer.ini and they all work as expected.

Let me know what you think!

sorgelig:
I already wrote above what I think.
Some options may look incompatible but may work for some specific hardware setup. I don't see a point to restrict it.

misteraddons:
S-Video and composite video are inherently 15 kHz formats that rely on
composite sync. These parameters are not optional, as they define the
signal standards themselves. Similarly, YPbPr requires sync on the Y
channel. Allowing users to disable these settings will result in invalid
configurations that won’t produce a usable signal on any display.

Do you agree with this?

TheJesusFish:
I don't know nearly enough about the edge cases here (if there are any), but if this would allow users to simply set their output mode, and not have to worry about enabling additional options to make their analog output work, it would be a big win from a support perspective. 

Sorting out what you need to enable on top of your `video_mode` can be a pain.

Just my 2 cents.

sorgelig:
> Sorting out what you need to enable on top of your `video_mode` can be a pain.

it's already made that way, so you won't need to change any sub-parameters. Just set vga_mode to required mode. Generally speaking, for novice users HDMI is the way to go. VGA output is kind of nerdy mode where users who really need this output, already know what they want. But anyway, even for VGA you just need to set vga_mode. Additional parameters just for special cases.

misteraddons:
What's the result of the following settings?
```
vga_mode=svideo
composite_sync=0
```
```
vga_mode=svideo
forced_scandoubler=1
```
```
vga_mode=svideo
vga_scaler=1
```
```
vga_mode=cvbs
composite_sync=0
```
```
vga_mode=cvbs
forced_scandoubler=1
```
```
vga_mode=cvbs
vga_scaler=1
```
I agree that the default analog video settings are good now, but it's easy for the user to accidentally break analog video. 
What's the benefit of allowing invalid settings?

TheJesusFish:
Just giving this a nudge due to the increase in new users thanks to the SuperStation and Qmtech/Hamgeek clones being available. 

Unless you anticipate an issue, streamlining these options would make things easier for the new folks.

misteraddons:
Any new thoughts on guarding against invalid analog video settings?

sorgelig:
I don't see a reason to revert modes if user explicitly supplied additional parameters.

misteraddons:
The reason is because the additional parameters break composite, and s-video. Neither format can be separate sync without breaking. Neither format can be scandoubled without breaking. Neither format can be upscaled without breaking.

For YPbPr, it's safe to enforce vga_sog. If the user has an IO board with the switch it either won't matter, or will be correct.

sorgelig:
Not every parameter breaks the things. For example i have scaler supporting separate sync. Also it's possible to have an external circuit for mixing SOG and sync. Probably only scandoubler can be enforced to 0.

misteraddons:
Interesting. For which video formats does your scaler support separate sync? I know RGB allows for composite sync or separate sync. I'm not aware that YPbPr, S-Video, nor composite video ever support separate sync.

How about the following required settings?
YPbPr:
composite_sync=1
vga_sog=1 (external circuit for mixing SOG and sync wouldn't be bothered by Arduino_IO9 outputting sync, right?)

S-Video and composite:
composite_sync=1
forced_scandoubler=0

manfredi90:
> What's the result of the following settings?
> 
> ```
> vga_mode=svideo
> composite_sync=0
> ```
> 
> ```
> vga_mode=svideo
> forced_scandoubler=1
> ```
> 
> ```
> vga_mode=svideo
> vga_scaler=1
> ```
> 
> ```
> vga_mode=cvbs
> composite_sync=0
> ```
> 
> ```
> vga_mode=cvbs
> forced_scandoubler=1
> ```
> 
> ```
> vga_mode=cvbs
> vga_scaler=1
> ```
> 
> I agree that the default analog video settings are good now, but it's easy for the user to accidentally break analog video. What's the benefit of allowing invalid settings?

I might be stating the obvious, but I'd like to point out that I don't think inhibit vga_scaler is correct. It can also be useful with vga_mode=svideo or vga_mode=subcarrier, for example, if you want to use a specific 15kHz video_mode (e.g., AO486 or x68000 cores on a CRT via composite).

misteraddons:
You're right, that was a mistake to include vga_scaler=0. I updated my suggested required settings in the previous comment.

PR #1056, opened 2025-11-01
Fix next/prev letter jump when directories start with _

zakk4223:


sorgelig:
why you need to skip "_" char? It's valid character participating in sorting.

zakk4223:
You couldn't use it to jump around the 'alternatives' directory because every directory in there starts with '_'

I just realized I forgot to limit it to only directories AND I forgot that's only a valid thing when SCANO_CORES is used. Ooops. I'll push a revised commit soon.

PR #1061, opened 2025-11-16
SPD Infoframe Fix

TheJesusFish:
I imagine there is a more elegant solution than this, but this uncomments the corename over HDMI code and fixes VRR.

Here is what I did:
- Rolled back to before the first SPD implementation
- Uncommented the spd_config_hdmi function code
- Added a vrr_requested variable to store the vrr state as it was getting zeroed out by the `if (!supports_vrr() || cfg.vsync_adjust) use_vrr = 0;` line

Tested this on several screens and scalers
- Tink4k
- Morph4k
- LG CX 
- iPad Screen
- Random Amazon 2400p portable monitor

I realized that I wanted to do rotation direction (as well as `is_rotated`) so unless you have a solution in mind for that, I can take a crack at it this week (well, Claude and I will take a crack at it).

Hopefully this closes Pandora's SPD InfoFrame

### DIFF
--- a/video.cpp
+++ b/video.cpp
@@ -1568,7 +1568,7 @@ static void spd_config(uint8_t *data)
 	}
 }
 
-static void spd_config_dv()
+static void spd_config_update()
 {
 	VideoInfo *vi = &current_video_info;
 	if (!vi->width)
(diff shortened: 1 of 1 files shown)


TheJesusFish:
Please make sure I did the rollback correctly. This built fine on my machine, but I would hate for Zakk to have to clean up my mess again :)

My branch that built fine: https://github.com/TheJesusFish/Main_MiSTer/tree/Rollback

sorgelig:
I can fix the issue differently. 
So, as i understood, VRR doesn't like SPD to be sent?

TheJesusFish:
> I can fix the issue differently. So, as i understood, VRR doesn't like SPD to be sent?

Correct. I believe VRR hijacks that? @zakk4223 can confirm

sorgelig:
ok, i've pushed relevant changes

Issue #1062, opened 2025-11-16
Autofire Overhaul


tonytoon:
I've been working on the autofire system and wanted to get some feedback on my proposed changes as I work towards implementing them.

Done:
- Display which buttons currently have autofire enabled anytime autofire is toggled for any button. Helps identify accidental activations of autofire.
- Calculate and display autofire in hertz rather than milliseconds to better match how hardware devices and emulators present the information.
- Allow defining custom autofire rates in mister.ini, with sane defaults if nothing is entered.

TODO:
- Per-button autofire rates. This has been a frequent ask I've seen from users in the shooting game community and the way I've rewritten autofire timing it should not be very complicated.
- I propose that we have two autofire rates by default (low and high) and pressing menu/osd while holding a button steps through them. This would be the most discoverable and least disruptive for users since it would visually indicate what is happening when they press a button combo they already use for autofire.
- Advanced users could take advantage of the custom autofire rates in mister.ini I've already implemented if low/high or the specific rates are insufficient.
- Remove left/right + menu/osd as a way to change global autofire rates. Users (or at least I) tend to hit that on accident and simply stepping each button through off/low/high would replace that functionality.

Thoughts? On the backend I've implemented a single 60hz timer for autofire which would make handling per-button rates straightforward and still easy to read.

Kitrinx:
You should probably just convince Zakk to use his. Ideally there would also be two modes, one that used simple hz based repetition like NES and one correlated with polls from the system like TG16

tonytoon:
That would be ideal since I'm sure he'd do it right but I don't have his address to send any bribes

@zakk4223 are you sure you want the guy who wrote gems like https://github.com/MiSTer-devel/Main_MiSTer/blob/master/input.cpp#L5851 submitting another PR?

zakk4223:
You can't use the existing mister timer mechanism to do a framerate-matched timer; it only has a resolution of 1ms so even at a bog standard 60hz rate it is wrong (the fractional milliseconds _matter_) and will just suffer the same problems the current one has; maybe if you're lucky the misses will be spaced  farther apart but maybe not. 

You can try to compensate for the loss/gain over time but that vastly complicates the code and probably doesn't fully solve the issue anyways. 

Maybe a super high resolution timer would work, I'm not sure.

ideally the framework would just increment a counter (3 bits is more than enough) that is query-able via something like UIO_GET_FRAMECNT. 

[...]

PR #1063, opened 2025-11-18
Extend Button/Key remap to support chording

zakk4223:
Don't merge yet! 

Creating this as a draft PR because it is kinda big and will probably need changes if https://github.com/MiSTer-devel/Main_MiSTer/issues/1062 happens.

The diff is big, sorry. A lot of it is just that menus take a lot of code. The biggest functionality changes are update_advanced_state/process_abm_entry in input.cpp

This allows setting button chords in button/key remap menu. Both input and output can be multi-button.

It also adds an 'advanced' menu you access by holding MENU_OK/Tab during button/key remap. 
There you can delete individual entries for that controller/keyboard and create remaps directly. 
Just like the previous joy/key map you can do joy->joy, joy->keyboard or keyboard->keyboard

The advanced menu also lets you directly map to core buttons even if they aren't bound to a physical button. Useful for functions you don't have room for/want on a single button, but still want to quickly access with a button combo: (Pause/Fastforward on PSX core etc).

The order of the input buttons does not matter, as long as all the buttons are pressed it is activated. The only exception is that if the input chord contains the OSD button(s) they have to be held down first to prevent it from conflicting with autofire activation. 

**The OSD has been changed so it activates when the assigned button is released instead of  on-press. This is so the OSD button can be used for chording.** 

The code attempts to handle weird cases like multiple activated chords that share/input output buttons. It generally does a good job with anything that maps to core buttons. If there are multiple activated entries that map to overlapping keyboard keys the core may see some keyboard events that press/release keys that are already in that state. 
I don't think many people are going to create remaps like this and preventing the extra key events would be very complicated and possibly involve having to track the state of all keyboard keys. 

A few questions:

- The old code allowed you up to 1024 mappings (so more than every keycode). The new code only allows 32 remaps. Is that enough? Should I extend it? I could possibly make it infinite it we really want to. Or maybe enough to fit an entire keyboard? So 120 entries? I feel like 32 is enough but maybe you know of some uses I don't.
 
- I ended up creating a whole separate way to capture the input/output event codes separate from the 'mapping' stuff in input_cb() since I needed multiple buttons and a few special keys for clear/advanced etc. It also pushes a lot of the state management into the menu code; let me know if you think it should be changed. Although I think it might just be moving complexity from one file to another....
 
- Is the 'advanced' menu maybe too hard/obscure to get to?

 

The biggest advantage to this feature is the NeoGeo people who keep asking for every possible combination of 2,3 or 4 simultaneous button presses to be mappable can just use this!


sorgelig:
that's big...
The question is how to maintain it in the future. It's quite intrusive in current code. Filled half of OSD code with remapping like MiSTer is about button remapping rather than emulating.. 
Listening the people who don't want to play how it supposed to be may lead to more and more weird request. Today they need a combo which they "can't" press in their crippled controllers, Tomorrow they will need a macro "press to win".

zakk4223:
The NeoGeo thing was a bit of joke ;)  Although I do think those people would be able to take advantage of it I think it has more/better uses than that.

Being able to do N->M buttons/keys or N->core button mappings helps with:

- Some cores have more _gameplay_ buttons than can comfortably fit on even a PS4/PS5 controller. Stuff like Intellivision or Jaguar with the number pads. You could just bind them all to a key combo and still be able to access them from the controller.

- The extra buttons on many cores (rewind/fastforward/pause/savestates etc) would be very comfortable to have on user defined button-combos. Makes it much harder to accidentally press them and if you're using a controller with limited buttons allows you to still have easy access to those features.

- I'd finally have a way to access the annoying arcade core "2 Player Start" without wasting a "real" button on it.

- You'd be able to bind any mister keyboard shortcuts to controller button combos. No more keyboard needed just for a screenshot.

- Some old computer cores have extra keys that aren't available on modern keyboards. Even if the core provides some built in key combo for them, users could bind them to something more comfortable for them. Or if they found some USB keyboard with some extra keys they could bind those however they want.

I updated the code with some changes; I tried to integrate the button/key capturing more into the existing mapping_* parts of input.cpp instead of writing an entirely separate one. It makes it so the menu code doesn't have to deal with as much of the actual capture mechanics. 

It helped a bit but I think there's no way around things like the edit menu just being a lot of menu/UI/support code. 
If I got rid of that and only allowed button->button and button->key it would be a lot smaller, but I think the button/key -> core button part is really useful so I'd prefer not to...

I will try and think if there is a different approach to that functionality that doesn't involve a bunch of menu screens...

sorgelig:
In general, i'm ok to merge your changes. May be, before merging, give other people to test it for some time.

zakk4223:
I've updated this to work with the latest autofire changes, along with a few bug fixes noticed in testing.

seascape:
@zakk4223, I have been hoping for years to have PPP/KKK shortcuts for Street Fighter games on MiSTer. But this implementation is so much more, so flexible and system-wide. Being able to take screenshots while playing without massive hassle alone is a big improvement. Thank you!

PR #1065, opened 2025-11-19
Rotation Direction in SPD

TheJesusFish:
Ask and you shall receive :)
 
No idea if this is the best way to do this but, when plugging in debug messaging, it returns the correct value based on the MRA. 

Also not sure if the `cfg.direct_video ?` is needed in the SPD, but I wanted to keep it to just direct video in the event that it gums up the normal HDMI frame somehow.

This also requires the `<rotation>` string in MRAs - so, if you dig this, I will go through and add them to any rotated games that are missing them.

Let me know what you think, duder.


sorgelig:
You are free to update MRAs

PR #1066, opened 2025-11-21
Send standard compliant SPD infoframe, when direct_video is not enabled

chriz2600:
I propose to not send the modified "DV1" ("VI1") (SPD) IF when direct_video isn't enabled.

This will greatly improve compatibility with standard HDMI equipment, like TVs using the SPD vendor/product information to name the HDMI input used.

Background: Unfortunately, DV1 infoframe and standard SPD infoframe are incompatible:
```c
typedef struct __attribute__((packed)) DirectVideo_Data {
    uint8_t id[2];          // 'D', 'V'
    uint8_t version;        // '1'
    uint8_t interlaced      : 1;
    uint8_t                 : 1;
    uint8_t menu_open       : 1;
    uint8_t rotated         : 1;
    uint8_t rotation        : 2;
    uint8_t                 : 2;
    uint8_t pixel_repetition;
    uint16_t horiz_de;
    uint16_t vert_de;
    uint16_t width;
    uint16_t height;
    uint8_t name[15]; // 12 according to IF length set
} DirectVideo_Data_t;

typedef struct __attribute__((packed)) SPD_IF_Data {
    uint8_t vendor[8];
    uint8_t name[16];
    uint8_t type;
} SPD_IF_Data_t;
```
The name does not line up and the additional DV1 data might create characters outside the allowed 7bit ASCII range (for the standard SPD IF), which is ok for direct_video, but can easily be avoided for standard HDMI output mode.

Most of the additional DV1 parameters are not used by scalers when receiving 1080p (or other high resolution) inputs. By using SPD vendor/product as proposed, any scaler or TV device can use this infoframe without needing special knowledge of the format (DV1 in this case). 

I've attached the relevant pages from the HDMI specification and ANSI/CTA-861-I:
[HDMI-Specification-1.4 Table 5-8 Packet Types.pdf](https://github.com/user-attachments/files/23681742/HDMI-Specification-1.4.Table.5-8.Packet.Types.pdf)
[ANSI CTA-861-I Errata SPD IF.pdf](https://github.com/user-attachments/files/23681749/ANSI.CTA-861-I.Errata.SPD.IF.pdf)

### DIFF
--- a/video.cpp
+++ b/video.cpp
@@ -2961,36 +2961,62 @@ static void spd_config_update()
 {
 	if (use_freesync_spd) return;
 
-	VideoInfo *vi = &current_video_info;
-	if (!vi->width) return;
-
-	uint8_t data[31] =
+	if (cfg.direct_video)
 	{
-		0x83, 0x01, 25, 0,
-		cfg.direct_video ? 'D' : 'V',
-		cfg.direct_video ? 'V' : 'I',
-		cfg.direct_video ? '1' : '1', // version
-		(uint8_t)((vi->interlaced ? 1 : 0) | (menu_present() ? 4 : 0) | (vi->rotated ? 8 : 0) | (cfg.direct_video ? (arcade_get_direction() << 4) : 0)),
-		(uint8_t)(vi->pixrep ? vi->pixrep : (vi->ctime / vi->width)),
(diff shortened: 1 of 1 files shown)


sorgelig:
Ok, let's send a standard SPD.

PR #1067, opened 2025-11-25
Add subcarrier.txt file

misteraddons:
During testing, some cores require different PHASE_INC values compared to yc.txt.


sorgelig:
I don't like the way it develops.. There is already yc.txt file for this purpose.
At first: need to understand why different values are required and if there is a conversion formula exists (preferred way). If it's impossible to do an automatic conversion then yc.txt probably can be updated to include both values (ugly way).

birdybro:
PSX at 59.8 was the only one that required a different value it seems? So is the yc.txt value incorrect maybe then?

image

misteraddons:
As we tested cores, we noticed unexpected results. I'm using a Sony CXA encoder, which is what many fourth and fifth generation consoles used. Most of the yc.txt values work great. The PSX banding issue is only visible with certain resolutions. MikeS did a great job tuning these cores for the YC active hardware. What are your suggestions for improvements? It's very nice to output RGB, S-Video, and Composite video with a single analog video setting.

sorgelig:
I suggest to try a tweaked value in original YC output to check if it fits or not

PR #1069, opened 2025-12-07
file_io: Expanded ScanDirectory file skipping to include boot#.rom files

Bemjo:
Some cores such as the Intellivision and Atari Jaguar have multiple bios files they can automatically load labelled boot0.rom, boot1.rom, etc. These cores also support loading ROM files with the .rom extension. The file skipping already skips any files named boot.rom, but these extra bios files are not skipped and still display in the file loading menus for these cores.

This expands the existing file skipping rule to include these numbered boot.rom files, from 0-9.

### DIFF
--- a/file_io.cpp
+++ b/file_io.cpp
@@ -1625,7 +1625,14 @@ int ScanDirectory(char* path, int mode, const char *extension, int options, cons
 					//skip non-selectable files
 					if (!strcasecmp(de->d_name, "menu.rbf")) continue;
 					if (!strncasecmp(de->d_name, "menu_20", 7)) continue;
-					if (!strcasecmp(de->d_name, "boot.rom")) continue;
+					if (!strncasecmp(de->d_name, "boot", 4))
+					{
+						int len = strlen(de->d_name);
+						if ((len == 8 || (len == 9 && de->d_name[4] >= '0' && de->d_name[4] <= '9')) && !strcasecmp(de->d_name + len - 4, ".rom"))
+						{
+							continue;
+						}
+					}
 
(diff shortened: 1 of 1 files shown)


TheJesusFish:
@Bemjo If you can, add cd_bios.rom as well.

birdybro:
The only issue I see is that there are some cores where people load a rom on boot on purpose, such as NES using boot1.rom as an NES cart file to load on boot and using boot2.rom for FDS images. The person might want to load that file again manually for whatever reason, maybe testing, etc...

Bemjo:
@birdybro I did consider this possibility, but my opinion on that would be the user who's doing that could rename these bios files to something more descriptive of what they're for specifically if the core allows manually loading BIOS rom files, as these boot roms seem to be meant for autoloading on core initialization. I don't see this as being any different then cores that allow manually loading these files, and use boot.rom, as main is already hiding that. The current Jaguar and CD-i core being an example of this.

@TheJesusFish Which core(s) use this and also allow loading ROM files for games, or other scenarios is this an issue?

birdybro:
Yeah, it's not a major issue, just the only thing I could see coming up possibly. Not voicing disapproval, just something I thought of.

TheJesusFish:
> @TheJesusFish Which core(s) use this and also allow loading ROM files for games, or other scenarios is this an issue?

Good question. I had in my brain that megacd and neogeo did, but I just checked and they do not, so maybe there isn't a situation where blocking that makes sense.

sorgelig:
i think it's ok to hide bootX.rom. Files for other purposes must be renamed.

PR #1070, opened 2025-12-08
Filter more bios files

TheJesusFish:
Some cores use more than boot.rom (jaguar uses boot1 and boot2). There is probably a better way to do this, but I didn't want to start futzing with string matching or other things way above my pay grade.


TheJesusFish:
Ahh nevermind. This is a dupe of https://github.com/MiSTer-devel/Main_MiSTer/pull/1069

PR #1075, opened 2025-12-18
SNES: add Competition Cart support.

srg320:


wwark:
For information, there is two versions of PowerFest 94 10k points and 1 million points. Each rom uses différent headers.
The 1m points version uses the headers, so the rom doesn't load.
image-1

I need to check if the header is also shared with prehistorik man game ...

sorgelig:
So, does it need further tweak or ready to use?

wwark:
It is ready to use for 
Nintendo Challenge (Japan) (En) (Competition Cart)
PowerFest 94 (USA) (Competition Cart, 10k points)

Still those not boot
PowerFest 94 (USA) (Competition Cart, 1m points)
Nintendo Campus Challenge 1992 (USA, Europe) (Competition Cart)

The header of Prehistorik Man (USA) (En,Fr,Es)
image

wwark:
Thank you !

Issue #1055, opened 2025-10-30
bootcore setting don't work


mambrino:
Looks like the bootcore setting don't work after 20250828 main update. Tested with previous release (20250525) and works perfectly.

n-batalha:
I can replicate this issue (as did another user today) with the latest release.

To reproduce, with this SNES core in the SD card `/media/fat/_Console/SNES_20250605.rbf`

Then this ini fails to load the core on boot:

```
[MiSTer]

bootcore=SNES
```

But this one boots the core:

```
[MiSTer]

bootcore=SNES_20250605.rbf
```

birdybro:
Since Nelson fixed this, it can be closed.

PR #1076, opened 2025-12-21
Fix bootcore config param for generic core names

n-batalha:
Candidate fix for https://github.com/MiSTer-devel/Main_MiSTer/issues/1055 for generic core names in `bootcore`. E.g. `bootcore=SNES`.

Tested on a MiSTer with a few generic and precise references in `bootcore`. I also did a POC of unit tests, and checked [these cases](https://github.com/n-batalha/Main_MiSTer/blob/6d99c99b2bba6b56a186895b49285737820d4e39/tests/test_bootcore_utils.cpp) but I kept the testing setup as out of scope of this PR.

From a user point of view, the generic corename seems handy since a reference to a precise core version can break each time there is a version upgrade, and the previous version gets deleted. It's also easier to setup, without needing to look up precise filenames.

(Warning: I haven't touched a language like C++ in >10 years, expect basic mistakes!)

### DIFF
--- a/bootcore.cpp
+++ b/bootcore.cpp
@@ -167,53 +167,92 @@ bool matchesCore_yyyyMMdd_rbf(const char *A, const char *B)
 	return true;
 }
 
-bool preciseMatchOrGenericCoreName(const char *A, const char *B)
+
+struct CoreMatch
 {
-	if (!A || !B)
-		return false;
+	char *path;
+	int date;
+	bool exact;
 
-	if (strcmp(A, B) == 0)
-		return true;
+	CoreMatch() : path(NULL), date(0), exact(false) {}
+};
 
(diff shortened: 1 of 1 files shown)


sorgelig:
If you add such functionality then it needs to be matched against most recent core version (by date in the file name) instead of first found one.

n-batalha:
Got it, thanks for checking. I had thought of doing that search in a separate PR to keep them smaller for you (and fix for the default case of the updates just keeping the latest version). I now added logic to search for all dated cores and pick the latest, and manually tested against:

- `bootcore=PSX` (with ~5 versions installed)
- `bootcore=PSX_20250710.rbf` (with ~5 versions installed)
- `bootcore=DonPachi.mra` (as per https://github.com/MiSTer-devel/Main_MiSTer/pull/993)
- no bootcore

PR #1081, opened 2026-01-06
Detect PCXT variants in is_pcxt

spark2k06:
Expand is_pcxt() to recognize Tandy1000 and PCjr as PCXT-class systems, while keeping the cached detection behavior.

I am working on splitting the current PCXT core into two (PCXT and Tandy1000) and in the future I will try to add the PCjr system. This change is essential for the identification of these systems to work correctly.

### DIFF
--- a/user_io.cpp
+++ b/user_io.cpp
@@ -283,7 +283,16 @@ char is_archie()
 static int is_pcxt_type = 0;
 char is_pcxt()
 {
-	if (!is_pcxt_type) is_pcxt_type = strcasecmp(orig_name, "PCXT") ? 2 : 1;
+	if (!is_pcxt_type)
+	{
+		if (!strcasecmp(orig_name, "PCXT") ||
+		    !strcasecmp(orig_name, "Tandy1000") ||
+			!strcasecmp(orig_name, "PCjr")
+		   )
+			is_pcxt_type = 1;
+		else
+			is_pcxt_type = 2;
+	}
 	return (is_pcxt_type == 1);
 }
 


sorgelig:
If these computers share many things then why don't keep a single core with option to switch between them?

spark2k06:
> If these computers share many things then why don't keep a single core with option to switch between them?

The core currently has two models linked (PCXT and Tandy1000), but resources are being wasted, especially memory, with elements that do not make sense between them, such as TandyVideo for PCXT or HGC/MDA for Tandy1000. These are the resources consumed by each module with respect to the base system (without modules):

| Resource           | Max       | Base   | CGA    | CGA+TandyVideo | HGC/MDA | TandyAudio | TandyKBD | OPL2  | CMS   | EMS   | A000_UMB | All   |
|--------------------|-----------|--------|--------|----------------|---------|------------|----------|-------|-------|-------|----------|-------|
| ALMs               | 110,000   | 12,551 | 14,229 | 14,442         | 13,546  | 12,802     | 12,549   | 13,266| 13,505| 12,599| 12,648   | 17,605|
| Registers          | ~220,000  | 17,861 | 19,755 | 20,081         | 19,680  | 17,974     | 17,883   | 18,509| 18,815| 17,776| 17,739   | 24,311|
| Block memory bits  | 5,662,720 | 945,035| 2,124,705| 3,042,209    | 2,026,404| 945,035   | 945,035  | 952,077| 945,035| 945,035| 945,035  | 4,313,183|
| RAM blocks (M10K)  | 553       | 138    | 282    | 394            | 270     | 138        | 138      | 147   | 138   | 138   | 138      | 553   |
| DSP blocks         | 112       | 33     | 35     | 35             | 33      | 33         | 33       | 34    | 37    | 33    | 33       | 40    |

Currently, with everything combined, it is at its memory limit, and this prevents me from experimenting with new features, such as 8087, or others... which also make more sense in one model or another:

| Resource           | Used     | % of Max |
|--------------------|----------|----------|
| ALMs               | 17,605   | 16.0%    |
| Registers          | 24,311   | 11.1%    |
| Block memory bits  | 4,313,183 | 76.2%    |
| RAM blocks (M10K)  | 553      | 100.0%   |
| DSP blocks         | 40       | 35.7%    |

Ultimately, it's a matter of being practical and facilitating the maintenance of each model... plus, I also want to try the IBM PCjr version, which is based more on the Tandy1000 than on the PCXT.

On the other hand, another reason is that it is interesting to have two entries (in the future there may be three with PCjr) in the core menu (in _Computer) and to be able to enter each of them... with their particular configurations, without having to enter only one and then switch between them. Although they are based on the same x86 system, they have some different features that make them unique.

sorgelig:
in most cases BRAM usage can be reduced after optimizations. Other resources don't have much use as i see.

spark2k06:
> in most cases BRAM usage can be reduced after optimizations. Other resources don't have much use as i see.

Thanks for the merge. I agree with you that optimizations need to be made, including using SDRAM instead of BRAM despite the complications. I have no doubt that at some point we will have to get started on this.

PR #1089, opened 2026-01-18
Frame-synced, per-button autofire system.

tonytoon:
New autofire system.

Autofire button presses are synchronized to core's display refresh rate.
Uses framecounter from core when available, falls back to timerfd using video timings.
Autofire rate can be toggled per-button. Button+OSD to step through rates or reset to off.
Custom autofire rates can be configured in mister.ini, including custom patterns.

Also implemented reference counting for button presses to handle overlapping buttons. Replaces old hack I implemented to fix same problem and hopefully makes future updates such as chorded buttons easier to implement.

Resolves issue #1062

### DIFF
--- a/cfg.cpp
+++ b/cfg.cpp
@@ -597,6 +597,7 @@ void cfg_parse()
 	has_video_sections = false;
 	using_video_section = false;
 	cfg_error_count = 0;
+	strcpy(cfg.autofire_rates, "10,20,30");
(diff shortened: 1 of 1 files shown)


tonytoon:
I need to make sure reference counts to buttons are cleared if a player changes controllers/changes assignments/etc. Any idea where the best place to do that would be?

tonytoon:
need to make some more changes, will squash before submitting pr

PR #1091, opened 2026-01-20
autofire overhaul

### LINKED ISSUE #1062: Autofire Overhaul

tonytoon:
implements frame timer, autofire revamp, tweak to button reference tracking to remove limit on number of duplicate buttons tracked (previously 2). touching input.cpp in so many places has made the diff quite large so I've included a summary of changes below.

resolves #1062

frame_timer.*
* implements frame timer. uses core frame counter if available, falls back to timerfd otherwise.
* exposes fpga_vsync_timer boolean from frame_timer.h as extern/global.
* frame_timer() has been added to the scheduler before input_poll().
* as basic sanity check, timerfd based frame_timer will discard values outside 50-75hz and instead use 60hz as the frame timer.
* FRAME_TICK() macro is available in frame_timer.h to execute code when a new frame is detected. (currently only used by autofire and i'm not sure what else could use it but it's there for the using)

autofire.*
* implement frame-synchronized autofire, per-button autofire rates and support for custom rates.
* depends on new frame_timer
* autofire now defaults to 10/15/30hz. I picked these values because they feel good to me and more than three would be cumbersome. If anybody has better choices chime in. 99% of users will be stuck with these.
* autofire now internally uses bitmask with 1/0 representing frames for which a button is being held/released. reports rate to user in hz.
* hold button and press menu (OSD or two button combo) to step that button through autofire rates or to disable autofire. direction+menu no longer adjusts autofire. You are still blocked from enabling autofire on directions.
* custom autofire rates can be defined in mister.ini as rates in hz or custom bitmasks up to 64 bits long (64 frames). extremely complicated custom patterns may display a wildly inaccurate estimated rate in hz.
* you can get some interesting non-obvious functionality out of the custom bitmasks but won't advertise that to avoid confusion.

other files touched:

mister.ini
* sample lines (commented out) to demonstrate setting custom autofire rates.

input.*
* added set_key_state(), build_joy_mask() and build_autofire_mask() functions for implementation of button reference counting.
* joy_digital() now uses reference counting to track keypresses. It stores this data in a KeyStates array and builds the final joystick and autofire arrays using simple ORs before sending the mask to user_io_digital. This will expand the number of effective physical keys that can map to the same button and be handled cleanly from 2 to unlimited.
* joy_digital() now uses code from autofire.cpp when setting/checking/clearing autofire on buttons.
* code for enabling/setting/disabling autofire has been moved from joy_digital() to a dedicated handle_autofire_trigger() helper, which joy_digital() now calls (cleans up joy_digital() a bit)
* input_poll() calls FRAME_TICK() macro to check for new frame and set new autofire masks if true.
* NUMPLAYERS moved to input.h so that player_input.* and autofire.* could size arrays based on it.
* Fix to BTN_OSD handling when autofire is toggled (uncaptured release event was previously causing false positives and interrupting inputs)

user_io.*
* reverted user_io_digital_joystick() to prior state where it expects a simple 32-bit bitmask.
* reference counting was previously done (by me) by sending a 64-bit bitmask (containing primary and alternate mappings), splitting it in two and ORing them together.

cfg.*
* added autofire_rates to config reading. since autofire data is so dissimilar to every other config string, it is instead processed by parse_autofire_cfg() within autofire.cpp

video.*
* current_video_info made global/extern for use in frame_timer.*

scheduler.cpp
* added frame_timer() to scheduler in co_poll between user_io() and input_poll()

main.cpp
* added frame_timer() between user_io() and input_poll() in case scheduler is disabled.

notes:
a side effect of frame-sychronized autofire is that the displayed autofire rate in hz may not precisely match the real rate. autofire rates are typically calculated with the assumption that a game's refresh rate (and thus polls inputs) at 60hz. a pattern of every other frame would thus be 30hz. if a game's display refresh rate is 50hz, every-other frame will have an effective rate of 25hz. if anybody using autofire on PAL microcomputers or a wonderswan takes issue with this discrepancy, you may report your concerns in person at: 1060 W Addison St, Chicago, IL 60613, USA

### DIFF
--- a/MiSTer.ini
+++ b/MiSTer.ini
@@ -341,6 +341,10 @@ vrr_vesa_framerate=0
 ; disable autofire if for some reason it's not required and accidentally triggered
 disable_autofire=0
 
+; custom autofire rates (in hertz). up to five allowed.
+; can also use literal bit patterns for on/off cycles, i.e. 0b00111. each bit is one frame.
+;autofire_rates=10,0b00111,20,30
+
 ; Specify a default video processing preset that will be applied to cores.
 ; Path is relative to the presets/ directory and can optionally include the .ini extension
 ;preset_default=General Hardware/Console - 3rdGen
--- a/autofire.cpp
+++ b/autofire.cpp
@@ -1,0 +1,342 @@
+#include <math.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <strings.h>
+#include <stdint.h>
+#include "input.h"
+#include "autofire.h"
+#include "cfg.h"
+
+/*
+    New autofire system.
+    We take the desired autofire rate in hertz and convert that to a bitmask
+    0 == button released
+    1 == button pressed
+    We advance through a single bit of this mask every frame on the assumption that
+    most games read their inputs every vsync (or is it vrefresh? once per frame.)
+
+    We display autofire rates to the user as if the game is running at 60hz, but internally
+    the rate is scaled to match the game's actual display refresh rate.
+
+    e.g. A PAL game (running at 50hz) with 30hz autofire will internally be toggling the button at 25hz
+    actual_refresh / 60 * autofire_rate_hz == real_autofire_rate_hz
+*/
+
+#define MAX_AF_CODES 16
+#define MAX_AF_RATES 6
+#define AF_NAME_LEN 32
+
+// global autofire cycle data.
+struct AutofireData {
+	char name[AF_NAME_LEN];				// display name
+	uint64_t cycle_mask;				// bitmask representing the autofire cycle
+	int cycle_length;					// length of the cycle in frames
+	int bit;							// current bit in the cycle
+	int frame_count;					// current frame in the cycle
+};
+
+// per-player autofire table; index 0 means disabled.
+struct AutofireTable {
+	int count; 								// number of entries
+	int index[MAX_AF_CODES];				// index of matching autofire rate in autofiredata[]
+	uint32_t mask[MAX_AF_CODES];			// bitmask representing which buttons this code represents 
+	uint32_t autofirecodes[MAX_AF_CODES]; 	// codes that have autofire set (or codes that had it set, then disabled)
+};
+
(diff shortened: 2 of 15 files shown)


tonytoon:
Deleted previous comment. Realized I overthought things and was able to drastically simplify the reference counting.

Instead of directly storing button data in joy[], joy_digital() now stores button masks per ev->code in KeyStates. It performs an OR on all masks to build the final bitmask which is sent to user_io_digital_joystick(). It does similar to build and apply the appropriate autofire mask. This will totally remove the current limitation of only properly handling 2 sources for button press/release events.

tonytoon:
Two button combos for menus are not working quite correctly now. All testing has otherwise gone well.

tonytoon:
@zakk4223 I hope that my simplified and not-overthinked approach to reference counting of buttons assists in your planned addition of advanced button mapping including chords. as long as they have different ev->codes it should all come out in the wash and require no to minimal manipulation of the input flow on your part. i think.

PR #1092, opened 2026-01-20
Add SNI uartmode

NobodyNada:
This adds a 6th uartmode for SNI support implemented in MiSTer-devel/SNES_MiSTer#462, allowing a host PC to remotely read and write the memory of the SNES core. 

See that PR for the complete details; the TL;DR is that the UART interface exposes a simple command/response protocol by which a process running on the Linux side can make requests to access SNES memory. [A daemon process](https://github.com/NobodyNada/snid) will then listen for requests from user applications over TCP. This PR adds the UART mode used by the SNI daemon process to the main menu.


sorgelig:
This patch needs to be improved.
1) What is the use of new UDP mode?
2) SNI mode must be visible only for SNES core.

jnkel:
@sorgelig

1. The UDP UART mode is not new, it existed in the `uartmode` script but was missing an entry in the menu. IIRC, selecting it via the commandline would have caused an out-of-bounds array access when drawing the menu, so it appeared to me that it was unintentionally missing from the menu. If that's not correct and it was intentionally hidden for some reason I can fix the logic.
2. Sounds good, I'll update that logic this weekend.

sorgelig:
UDP wasn't available and probably never tested/finished. So it needs to be hidden.

NobodyNada:
Fixed:

1. The UDP mode is hidden unless it was already selected (e.g. via the command line)
2. The SNI mode is hidden unless either it was selected manually, or the user is using the SNES core and has installed the SNI daemon.

Issue #823, opened 2023-09-19
Feature request: Add analog trigger polling for future Saturn support


YLFAndy:
Currently analog triggers are not polled as full byte values. They can be mapped as digital but cannot be read for xinput byte value. This is (currently) specifically an issue for the Saturn core, which has a number of emulated controllers with analog inputs apart from the L and R sticks (perhaps analog trigger quirk could be added?) 

NOTE: The core *currently* can write values to the emulated Saturn pad for the inputs, but it is not hooked up because an appropriate analog read source is non existent. See: https://github.com/srg320/Saturn_MiSTer/blob/master/rtl/hps2pad.sv
Applying value to the Z1/Z2 inputs in-core will work, but cannot be read from any analog currently apart from L and R sticks A test core has been created that maps Y2 up to Z1 and Y2 down to Z2 to verify this.

Two known xinput cases:
Modern console trigger/8bitDo ultimate controller - both triggers map in Xinput as Z. RT as 0-255, LT as -255-0.
Reflex Adapt using Saturn 3D pad - R maps to 'Throttle' L maps to 'Brake' both 0-255. 

In-menu work arounds failed - attempting to map triggers to analog stick 2 inputs causes the mapper module to hang after stick 2 down is entered on a number of controllers with analog trigger support. The mapper graphical menu shows up and hangs on dpad R assignment - tested with multiple controllers with analog triggers.
Repro steps:
Using a controller with analog triggers, launch Define Joystick Buttons in the main MiSTer OSD
Map the first set to D pad
Map stick 1 to stick 1
When "Press right on Stick 2" is displayed, pull the right trigger.
When "Press down on Stick 2" is displayed, pull the left trigger.

Result: This hangs the remainder of the controller mapping.

If this is an issue that can be solved per-core, please close (thought if it was mappable for racing games in general as throttle/brake, that would be a bonus).

sorgelig:
I will think about implementation.

dazzer69:
Any update on implementation?

thorr2:
[...]

PR #1093, opened 2026-01-31
Add support for analog triggers

n-batalha:
WIP, opening draft PR in case some of an unexpected objection/blocker to adding analog triggers, and further discussion.

Plan for PR:

- [x] Add mapping steps for analogue triggers
  - [ ] Simplify mapping (less steps)
- [x] Add runtime use of analogue triggers
- [x] Modify InputTest to support analog triggers ([PR here](https://github.com/MiSTer-devel/InputTest_MiSTer/pull/17))
- [ ] Add support in the gamepad controller db (currently disabled)
- [ ] Add controller input mapping migration v3->v4
- [ ] Account for path of controllers with digital triggers, emulating analogue triggers

- Testing
  - [x] Tested with the modified InputMapper and a Dualsense controller
  - [ ] Test more controllers (digital triggers)
  - [ ] Test more controllers (analogue triggers)
  - [ ] Test cores without changes, that use triggers (PSX, N64?)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,3 +28,5 @@ host
 *.user
 .vs
 MiSTer
+
+.DS_Store
--- a/input.cpp
+++ b/input.cpp
@@ -1155,6 +1155,8 @@ typedef struct
 	uint16_t jkmap[1024];
 	int      stick_l[2];
 	int      stick_r[2];
+	int      trigger_l;
+	int      trigger_r;
 
 	uint8_t  has_kbdmap;
 	uint8_t  kbdmap[256];
@@ -1403,7 +1405,9 @@ static int mapping_current_key = 0;
 static int mapping_current_dev = -1;
 
 static uint32_t tmp_axis[4];
+static uint32_t tmp_triggers_axis[2];
 static int tmp_axis_n = 0;
+static int tmp_triggers_axis_n = 0;
 
 static int grabbed = 1;
 
@@ -1424,9 +1428,11 @@ void start_map_setting(int cnt, int set)
 	mapping_clear = 0;
 	mapping_finish = 0;
 	tmp_axis_n = 0;
+	tmp_triggers_axis_n = 0;
 
-	if (mapping_type <= 1 && is_menu()) mapping_button = -6;
+	if (mapping_type <= 1 && is_menu()) mapping_button = -8;
 	memset(tmp_axis, 0, sizeof(tmp_axis));
+	memset(tmp_triggers_axis, 0, sizeof(tmp_triggers_axis));
 
 	//un-stick the enter key
 	user_io_kbd(KEY_ENTER, 0);
@@ -1487,8 +1493,8 @@ static char *get_map_name(int dev, int def)
 	static char name[1024];
 	char *id = get_unique_mapping(dev);
 
-	if (def || is_menu()) sprintfz(name, "input_%s%s_v3.map", id, input[dev].mod ? "_m" : "");
-	else sprintfz(name, "%s_input_%s%s_v3.map", user_io_get_core_name(), id, input[dev].mod ? "_m" : "");
+	if (def || is_menu()) sprintfz(name, "input_%s%s_v4.map", id, input[dev].mod ? "_m" : "");
+	else sprintfz(name, "%s_input_%s%s_v4.map", user_io_get_core_name(), id, input[dev].mod ? "_m" : "");
 	return name;
 }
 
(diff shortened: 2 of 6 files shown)


sorgelig:
1) what is the purpose of UIO_ATRIGGER_L/R I/O codes?
2) Analog triggers have so many variants. It can be a single/dual axis, it can be 0-max or max-0.
3) Better to make it transparent, so besides analog report to the core, it should generate button presses as well. Core will decide what input to use.

zakk4223:
>     2. Analog triggers have so many variants. It can be a single/dual axis, it can be 0-max or max-0.

I've found a number of controllers where the analog stick axes are inverted (right/up is negative). 
I think these are unmappable currently. the gamecontrollerdb can indicate these too, but that is also not supported.

If we have to change the on-disk mapping file version for triggers, it might be a good time to add whatever needs to be added for the weird sticks too. Probably just a flag indicating which direction from reported center the axis is? (which I think works for triggers too?)

sorgelig:
There are some controllers use the same axis for both triggers, so if you press one, then it will go 0..+max if you press other trigger then it will go 0..-max. If you press both then result will be 0. 

For generic user it's hard to explain why his lovely controller doesn't work. The key point is: if you start to implement this feature then need to be sure it will work with as many controller as possible. Not only specific controllers.

n-batalha:
>if you start to implement this feature then need to be sure it will work with as many controller as possible. Not only specific controllers

Absolutely, I intend to test with more controllers. I would just hesitate if it's a rare behaviour (normally not supported, and with rare+old controllers), I need to research this further, thanks for noticing this.

>1. what is the purpose of UIO_ATRIGGER_L/R I/O codes?

I mainly replicated the behaviour for other input events (mainly sticks), you can see how these 2 are used in the core side [here](https://github.com/MiSTer-devel/InputTest_MiSTer/pull/17/changes#diff-306cec83e9a332d7af55930f75cfbd19ad332e2ed5eff0557442f64fbfbe71a4R479-R493). I separated left/right for readability, instead of joining 2 values in one command.

>3. (...) besides analog report to the core, it should generate button presses as well. Core will decide what input to use.

I was thinking in a follow-up PR, to suggest replacing the virtual SNES controller with a "PSX" one to cover triggers. And then all input can be covered via the user mapping in main, or better yet, via the gamepad db. Solving https://github.com/MiSTer-devel/Main_MiSTer/issues/1095 . This would involve doing what you're saying? The current change doesn't cancel the triggers usable in say PSX/N64 when doing the mapping via the core, the triggers work as buttons (I think! Also tested this). It's just to solve https://github.com/MiSTer-devel/Main_MiSTer/issues/823 by making a new cmd available for controllers only usable via snac currently. If it takes a while between both changes, you could see users confused with defining the triggers in main and these not working in the PSX core and requiring a mapping there still, but I reckon it's not too different from now (as covered [here](https://github.com/MiSTer-devel/Main_MiSTer/issues/1095)). Seemed sensible to split changes in smaller ones?

I might miss some context as new to this type of project and codebase, appreciate the feedback.

tonytoon:
I'm down for switching a virtual PSX pad. Seems the most abundant layout nowadays by far.

Well virtual Xbox pad unless somebody wants to make custom glyphs for the PSX buttons and get them added to the font.

sorgelig:
virtual SNES controller is for digital buttons only. There is no point to change to another one. Analog triggers reaction test need to be done separately like it's done for analog sticks.

zakk4223:
The original psx pad didn't have analog triggers anyways, so it would be a digital mapping.

There are three cores that can't do complete auto mapping to a virtual SNES pad: 
PSX, Saturn and N64. 

Seems the point of the virtual gamepad was to provide a 'configure once' experience but for those cores it can't. 

It would be nice to have that work, but on the other hand it would add 4 more stages to the main map dialog. And people that are trying to map something like usb SNES/genesis pads would have a lot to skip over.

n-batalha:
> The original psx pad didn't have analog triggers anyways, so it would be a digital mapping.

Sorry, I used quotes on "PSX" as it wouldn't be exactly the same, we'd have trigger steps and we'd have detect whether the user has digital/analog sticks (although leaving out configurable controllers that support both, like the 8bitdo Ultimate 2 Wireless) and generate signals for both analog and digital - emulating analog if the controller doesn't have it. But seems we can drop that approach via the virtual controller

n-batalha:
(going to be a month or so inactive, closing to clear the PR list and reopen when resuming work)

misteraddons:
One advantage of using the PSX controller standard for mapping is exposing L2 and R2. Arcade games that use 6 action buttons (e.g. Street Fighter II) don't have enough mappable buttons for Pause using the standard SNES pad, so there's no default Pause button mapped in those .mra files. This isn't critical but it's a nice to have feature.

`<buttons names="LP,MP,SP,LK,MK,HK,Start,Coin,Core credits" default="A,B,X,Y,L,R,Start,Select,-" count="6"/>`

sorgelig:
It's not a problem to add L2/R2 to virtual SNES controller

tonytoon:
> It's not a problem to add L2/R2 to virtual SNES controller

My concern with that would be consistency with how inconsistent controllers are.

Controllers that do use the YX/BA SNES layout (Switch) don't call them L2/R2 - they call them ZL/ZR.

Xbox uses XY/AB and calls them RT/RB LT/LB

It's only on PlayStation that they call them L2/R2, but they use the square/cross/etc.

I'm sure everybody would *know* what was meant by L2/R2 but it wouldn't match any controllers in the wild, to the best of my knowledge.

Not sure it's a problem worth worrying about though if it gets the point across. (and even then those wouldn't work with the 8bitdo m30 which is extremely popular with MiSTer users and mimics the *saturn* pad buttons with abc/xyz/lr physically written on them... sigh)

birdybro:
Yeah I think people generally know what L2 and R2 are, it's probably fine. It's more about communicating it to the user what button to press or map, not as much about a canonical accuracy :P

PR #1094, opened 2026-02-02
per key tracking of autofire timer

tonytoon:
switch to tracking autofire timer per key from globally.


tonytoon:
@zakk4223 noticed that using a global timer it's possible for an autofire action to take a couple frames before it triggers due to starting in the middle of the global cycle. added per key tracking of it's location in the cycle as a frame count.

Issue #789, opened 2023-07-18
Feature Request: Copy the save file to the SD card without to open OSD 


caiosilva1993:
Feature Request: Is it possible to implement autosave without the need to open the OSD menu to copy the save file to the SD card?

sorgelig:
This is unsafe feature as it may write to SD card at any time, including the time when you may press reset or power MiSTer down.
This is why write is implemented in determined times (when you open OSD).

machine-check-exception:
Hi
Just an idea: it could be interesting and option inside the osd, under the reboot option, for example, that says: turn off. That could do all the writes left (if any) and leave the system in a semi standby state. It would be great if it could auto shut off like pi4 (raspberry) does, for example, but that is probably a hardware limitation. Even if that button is almost placebo, it could help people to take the habit of open the osd and activate that option before shutting down the power (and hence not loose any nvram change)
Hope not to say any stupidity :)
Thanks

sorgelig:
MISTer concept originally is about using as a non-computer device. So whenever you want it turn off then just turn off. Adding computer style menu option "Turn off" is opposite the main concept. By the way, there is no power control on the board, so it won't able to be turned off nor sleep programmatically. The only available option is to terminate all Linux processes and halt.

I found it funny you forget to just open OSD to automatically save the data but you want to have menu where you have to navigate to specific option ;)

machine-check-exception:
Damn!! You missed my point!! My bad. Let me try again :D

You could create an option that says "turn off". That options **help** users to get **the habit** of: **open the OSD** (that's the important part, of course xD. I omitted that because I found it so obvious as it is the main purpose of the "turn off" option) and then select "turn off" (that can even be completely placebo; Just print a message saying "good bye" or, if needed/desired, do some work). 

I suspected about the hardware power control limitation :(

Jokes apart, I would like to add/link with your comment:

a) Agree with the primordial idea of being a stateless system but it looks that as far as we go, the systems recreated are getting more complex, for example with the nvram or maybe other techniques/technologies that are currently present or may be present in a future. It would be great that mister could behaves, for example, as a replacement for an arcade cabinet, but with the nvram stuff that most arcades have, you can't just turn on and at the end of the day turn it off without loosing data. If I'm correct, you loose at least the scores and if you do some changes in the service and don't open the OSD, those changes are also lost. So the stateless original idea, I think is "already lost".

[...]

PR #1097, opened 2026-02-16
Add autosave interval option

birdybro:
Putting here as a draft, I need to test this later. It also should be discussed before anything pulled in. This was AI generated via Claude, but it looks ostensibly correct.

Goal is to have a timer on Main, send pulse via spi to core that it should trigger an autosave, there is a coinciding template update to go along with this.

https://github.com/MiSTer-devel/Template_MiSTer/pull/99

https://github.com/MiSTer-devel/Main_MiSTer/issues/789

### DIFF
--- a/MiSTer.ini
+++ b/MiSTer.ini
@@ -370,8 +370,9 @@ controller_unique_mapping=0
 ; set to 0 for manual lock from OSD
 osd_lock_time=5
 
-; Autosave interval in seconds. When set, periodically sends a pulse (UIO cmd 0x44) to the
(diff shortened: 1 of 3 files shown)


birdybro:
Added an update to use ddr3 shadow copy as per wickerwaka's idea in a chat.

This adds a configurable autosave interval that periodically saves game SRAM to the SD card without requiring the user to open the OSD. Instead of slow SPI byte transfers, the FPGA copies dirty save RAM into a DDR3 buffer, and the ARM reads it directly via memory mapping — then writes it to the save file in one bulk operation.

birdybro:
Closing as most of this should not be handled in Main ideally.

PR #1098, opened 2026-02-17
HDMI-CEC via ADV7513

misteraddons:
HDMI-CEC support via the ADV7513 integration path, with conservative defaults and low runtime chatter.

Features:
- Enable/disable CEC - default disable (`hdmi_cec`)
- Device name (`hdmi_cec_name`)
- Automatically switch TV to correct input on boot/wake
- TV remote inputs control MiSTer menu (`hdmi_cec_input_mode`)
  - Up, Down, Left, Right, OK
  - Back (also default OSD) (`hdmi_cec_osd_key`)
- Wake TV (power on or cold reboot) or with controller input when system is idle (`hdmi_cec_wake`)
- Sleep TV  (`hdmi_cec_sleep`) when idle timeout is reached (`video_off`)
  - Note: `video_off` previously sent black frames when OSD is visible and osd_timeout was reached, with simple enable/disable option. Now, `video_off` will sleep video regardless of OSD state, and has presets 1-4 (1=15m, 2=30m, 3=45m, 4=60m, >5=timeout duration in minutes)

Note: DE10-nano requires hardware modification to enable CEC - missing CEC 12MHz clock. Feature was tested on third party hardware.

### DIFF
--- a/HDMI_CEC.md
+++ b/HDMI_CEC.md
@@ -1,0 +1,89 @@
+# HDMI CEC Behavior
+
+This document describes current HDMI CEC behavior implemented in `hdmi_cec.cpp`.
+
+## Config
+
+- `hdmi_cec=1`: enable HDMI CEC.
+- `hdmi_cec_input_mode=0|1|2`:
+  - `0`: ignore all incoming CEC remote key input.
+  - `1`: accept CEC keys only when OSD/menu is active, except the configured OSD trigger key.
+  - `2`: always accept mapped CEC keys.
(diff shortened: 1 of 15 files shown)


TheJesusFish:
Confirmed that this works on a QMTech clone board.

Kitrinx:
This works great on some clone boards, and theoretically the de10 itself could be modded to also support it. Great change.

sorgelig:
It's not possible on original HW, it's waste of CPU power and make code complicated.
Bought new TV receintly and already got problems with CEC even on originally supported devices. Most popular solution for CEC fixes is CEC blocker :)

PR #1099, opened 2026-02-17
Try and use sqlite to have automatic saves.

Tombert:


birdybro:
I think this is probably a bit overkill, to import sqlite directly into the project. There is probably another way to handle it. We have a few devs talking amongst ourselves about the strategy currently.

Tombert:
I figure it's worth a shot just to try it out.  Worst case scenario I wasted a half hour with codex. 

It probably is overkill and I think the quasi journaling thing I mentioned earlier would probably be fine.

Tombert:
@birdybro It took a bit more than half an hour, but for what it's worth it works now, using exclusively sqlite.  

I suspect we probably don't want to mandate the exclusive use of sqlite since people probably like to use the `.sav` files, but at least as a POC the sqlite thing works fine.

sorgelig:
Please stop suggesting more and more wild solutions.
Current save code is perfect and lightweight. Problem is not in code structure but in moment of writing. Changing the code using DDR3 or sqlite doesn't fix the problem. I even won't look into this PR as it's too wild already. No one will use such a heavy API in place where each microsecond matters.

Tombert:
@sorgelig It is decidedly not perfect. The fact that people have complained because they lose save progress because you seem to be under the erroneous impression that it's not irritating and easy to forget to open the OSD indicates that it's not perfect.  Also, in your current system I believe if the system crashed or lost power mid-save it would be corrupted because exfat is terrible. 

I do find it extremely amusing that I get a pissy comment telling me to "make a PR" so as to try and shut me up, I say I don't want to do that because I figured I'd get some sanctimonious response and I'd waste my time, and I'm called "entitled" like I'm asking for free labor.  Then I make the PR with the fix it, and you tell me you won't even look at this PR because it's "too wild".  It's almost like I was completely right to call this out beforehand. 

Yeah, it's so wild that people want something to work the way that they expect it to work. What people who "save" in their game *really* want is a a diatribe about why the irritating current system is "perfect".  That's so crazy, I'm so quirky for wanting my device to do the thing that everyone expects it to do. 

I don't really understand how you think I changed "code structure", *all* I addressed was the moment of writing and reading.  I could have reinvented journaling like my previous suggestion but I figured it would be less convoluted to use sqlite. The overall structure is mostly unchanged, I basically just changed how data is written and read.  

And I *did* get it working, to be clear.  It works fine, I played Link to the Past for several hours, having it auto-save every five minutes, and it worked *fine*.  I didn't notice any stuttering, and I purposefully killed the power mid-save to make sure it didn't corrupt the file and it didn't.  SQLite isn't really a database, it's basically fopen with atomicity guarantees, so the API isn't really that "heavy". 

But you're right, I don't need to waste my time with more "wild" solutions, because I actually solved the proble. Despite what you seem to think, it *really* isn't that hard to get this working.

sorgelig:
If you could tell that you are going to use a database system in advance then you would save a lot of your time. Making PR doesn't mean it will be automatically accepted. You are focusing only on single issue, but i think about whole project and further supporting. So database is definitely "NO". For power outage you need an UPS, not the change of code. It doesn't matter if you write a plain file or add a record to database which will eventually write a file too. You don't to need add a whole new API to achieve the same result by writing a sav file every 5 minutes.

Tombert:
> If you could tell that you are going to use a database system in advance then you would save a lot of your time.

@sorgelig I didn't waste my time.  I got my device working how it's expected to work.  From my perspective the problem is solved.  I was just hoping *a* solution could be upstreamed so it's not living in some random fork. You can run your project however you'd like.  I can run my forks however I'd like. It's a beautiful system. 

Again, I don't really consider SQLite a "database", it's just a disk-writing library.  It's not like I installed Postgres or Oracle on there. 

> It doesn't matter if you write a plain file or add a record to database which will eventually write a file too.

I know it writes files.  Sqlite very explicitly creates files.  I have them under the /media/fat/saves directory. 

How you write files is not created equal. The choice of exfat means you have to do extra work to ensure that writes aren't corrupted because exfat sucks.  My argument was that writes should be "all or nothing", either the write was successful or it's like it didn't happen at all.

My initial idea was to use temporary files but since, again, exfat is terrible, you don't get that kind of atomicity against crashes or power loss because something could get borked mid `mv`. 

My second idea was to have versioned files with a checksum to make sure that the writes weren't corrupted upon write, and if the latest file *was* corrupted, then we go back to the latest non-corrupted file.  This, in a hand-wavey way, is how journaling works in modern filesystem and this wouldn't be necessary if we used a filesystem that didn't coexist with dinosaurs. 

I'm pretty sure that that idea would mostly work fine but it would create a lot of files and because (and I cannot overstate this) exfat is terrible this would create a lot of metadata.  exFAT has the lovely and unique ability to corrupt the metadata in addition to the file, potentially breaking the entire directory.  

So I decided to, instead, use a library that handles the journaling and versioning automatically, and it can handle it in a single file reducing metadata churn. 

> For power outage you need an UPS, not the change of code. 

No, this is a bad response.  There are plenty of reasons that the writes can be interrupted independent of a power outage. The computer can crash, it can be knocked off the table it's resting on, the power adapter plugged into it could give out.  There's dozens of other ways that a write can be interrupted that can't be handwaved away with "LOL GET A UPS BRO". 



> You don't to need add a whole new API to achieve the same result by writing a sav file every 5 minutes.

Even I granted your premise with the UPS, which I don't, there's still the issue of "you could forget to hit the OSD button".  Forcing the users to hit the OSD is weird.  It makes it so that the system isn't doing what the user thinks it's doing. 


---- 

Again, you can run your project however you'd like, and you're free to enforce whatever terrible decisions and policies that you'd like.  I'm happy enough to maintain my own fork.   I won't open any more PRs.

PR #1105, opened 2026-02-24
Fix: ignore vga_mode=ypbpr unless direct video is active

misteraddons:
direct_video=2 is treated as a request and then resolved to:

- 1 when a known HDMI DAC is detected

- 0 for normal HDMI displays

Apply vga_mode=ypbpr only when direct video is actually active (direct_video==1).

Refresh HDMI CSC after direct_video=2 resolution so non-DAC HDMI stays RGB.


sorgelig:
As far as i remember i made YPbPr working in direct video and non-direct video.

PR #1106, opened 2026-02-25
CD-i: Buffer overflow fix + Initial default NvRAM

Slamy:
I'm feeling slightly ashamed about this. I'm fixing something I've fixed and broken again, just for the need to fix it again.


birdybro:
Don't worry, you aren't the only one who runs into this every now and then. ;)

PR #1107, opened 2026-03-02
Add video_mode FIFO command for CRT video mode switching

theypsilon:
### The problem:
Many users are interested in running the menu core with vga_scaler=1 so that they can see what the scripts are doing directly on their CRT display, but they have trouble finding video modes compatible with their consumer TVs.

The only way they have to check video modes at the moment is modifying the MiSTer.ini by hand, testing different video modes, and rebooting the system to see the result, which is a very tedious and complicated process for most users, and furthermore needs to be done with extreme care, because a wrong modeline can produce out-of-spec frequencies that can damage users' CRTs.

### The solution:
This PR implements a new command for the MiSTer_cmd pipe API, that can be used like this:
```bash
echo "video_mode 320,13,31,52,240,2,3,16,6515" > /dev/MiSTer_cmd
```

When used through some tool, it provides very fast feedback to the user, making it easy to decide whether a video mode works for them or not.

It's important to note that this new command does not apply video modes blindly. It contains a horizontal frequency guard that makes sure the applied levels are withing a conservative safe range. This safe range (15625-16500 Hz) is taken from the [Switchres project](https://github.com/antonioginer/switchres), which is a modeline generator tool that also serves as one of the best references for safely operating CRT monitors. This is the exact source that has been checked: https://github.com/antonioginer/switchres/blob/master/monitor.cpp

This PR does not support non-15kHz video modes for pragmatic and safety reasons. The vast majority of CRT users have 15kHz equipment, and 15kHz CRTs are much more prone to damage from out-of-spec frequencies, especially those intended for 25kHz or 31kHz monitors. Conversely, higher frequency monitors are generally not damaged by 15kHz signals, so they can are not in danger when users apply 15KHz video modes by mistake.

### Summary of the changes:

- video.cpp: Add video_mode_cmd() that parses custom modelines and applies them via video_set_mode(), with hfreq safety check blocking modes outside 15625-16500Hz to protect 15kHz CRT monitors
- video.h: Declare video_mode_cmd()
- input.cpp: Dispatch "video_mode" commands from MiSTer_cmd FIFO to video_mode_cmd()

### Demo

Here are a few demos that use the video command to set video modes and also make positional adjustments:

https://youtube.com/shorts/URBvRN13Q5I
https://youtube.com/shorts/i0HwIST7sNU
https://youtube.com/shorts/2wZEj8kSq7E

### DIFF
--- a/video.cpp
+++ b/video.cpp
@@ -3910,16 +3910,6 @@ void video_mode_cmd(char *cmd)
 		return;
 	}
 
-	int htotal = v.param.hact + v.param.hfp + v.param.hs + v.param.hbp;
-	double hfreq = (v.Fpix * 1000000.f) / htotal;
-	// Safe hfreq range for 15kHz CRTs based on Switchres arcade_15ex preset (15625-16500 Hz)
-	// https://github.com/antonioginer/switchres/blob/master/monitor.cpp
-	if (hfreq < 15625 || hfreq > 16500)
-	{
-		printf("video_mode_cmd: BLOCKED. hfreq=%.0fHz is outside 15625-16500Hz safe range.\n", hfreq);
-		return;
-	}
-
 	v_def = v;
 	v_cur = v;
 	video_set_mode(&v, v.Fpix);


sorgelig:
I think, this command needs to be improved because i already predict a new request to support any refresh rate.
Probably an additional parameter like this "video_mode tv,... " will add a limiter if required.

theypsilon:
I removed the runtime check. The rationale was to avoid trusting other software, so relying on other software to trigger the check defeats the purpose.

paulb-nl:
Is this just for seeing the scripts on the Menu core? The only thing you need to do is enable Directvideo and you can see the wallpapers and scripts on your crt.

theypsilon:
@paulb-nl good to know, I though the only way to enable this was through the vga_scaler option.

But this is still useful for finding better resolutions, according to the needs of the users and their equipment.

PR #1117, opened 2026-03-13
Suppress N64 "Waiting for N64 game... " debug message

misteraddons:


misteraddons:
Any thoughts on this?

sorgelig:
i'm not familiar with N64 core but together with message you removed early return too, which breaks the logic.

misteraddons:
Good catch. I just commented out the print statement.

PR #1119, opened 2026-03-15
autofire tweaks

tonytoon:
increased number of allowed autofire rates from 5 to 30 (some people wanted to have upwards of nine rates and be able to step through all of them)

added autofire_on_directions to mister.ini to allow toggling autofire on the dpad. defaults to off. i was surprised people asked for this but who am i to argue with those who love varth.


tonytoon:
re: increase in number of rates. some users felt that as nice as the much more accurate and user configurable autofire rates were, it was a regression not being able to step through rates arbitrarily. five was admittedly a pretty arbitrary limitation so i raised it to thirty which should be plenty. those users can now set autofire_rates=5,7.5,8.6,10,12,15,20,30 or whatever in mister.ini and get more, and more granular control over their speed. default behavior remains a simple "as fast as most people can mash fire" (10hz), "as fast as the fastest humans can mash fire" (15hz), and "as fast as possible while timed to vsync" (30hz)

Issue #1123, opened 2026-03-15
Optimize screenshots to prevent stalls in CD-based cores and reading input.


tonytoon:
I noticed some performance issues with how screenshots are currently handled. I'm working on a PR that should significantly reduce that impact.

If profiling functions are correct, main presently stalls out for around 250ms when a screenshot is captured. This breaks down to 80ms to copy the scaler's data into a buffer, and 170ms to transform that buffer into a PNG file and write it to disk (this is using SNES core for testing, other cores are typically faster.)

People who know better than me: does that mean main is non-responsive for that entire ~250ms? Do screenshots make inputs/chd loading/etc. pause/hiccup?

My plan of attack.

* I've rewritten the scaler copy to use simd instructions, which sped up the copy significantly because it can interleave the data in preparation for our final image (8ms, down from 80ms for 480 line buffers.)
* I plan on dumping imlib2 for a small rgb->bmp function. That'll be lossless which means larger files (~330KB), but zero cpu overhead on compression. That should save a huge chunk of the 170ms image processing and writing to disk step. (current tests show 45ms, down to 170ms for 480 line buffers)
* I've already moved image processing / write to offload_add_work so that it's not taking time from main to perform non-time-critical work.

Question for @alanswx who wrote the original code, or @sorgelig if you know more about the backend of how the scaler's memory is exposed.

The original copy from the scaler to a buffer took around 80ms, which is almost 5 full frames @ 60hz. But we never got corrupted screenshots from the buffer getting modified mid copy. What magic is going on there to make that happen?

sorgelig:
I don't know why it takes 80ms to copy from DDR3 to DDR3. I believe it should be much faster considering the screen size of SNES. Also, there are at least 2 buffers for screen which gives about 16-20ms to copy without tearing.
What is the purpose of rework? Screenshot function is not intended for movie capturing.

tonytoon:
> I don't know why it takes 80ms to copy from DDR3 to DDR3. I believe it should be much faster considering the screen size of SNES. Also, there are at least 2 buffers for screen which gives about 16-20ms to copy without tearing. What is the purpose of rework? Screenshot function is not intended for movie capturing.

To speed it up. Taking that long to copy a buffer was a surprise to me as well, plus converting to PNG and writing to disk added up. I worry that main might be getting blocked while it happens. I believe that would mean no input polling, no chd loading, etc, until the entire job completes, correct?

[...]

PR #1126, opened 2026-03-17
screenshot optimizations to remove stalls in inputs / cd based cores

tonytoon:
issue:
screenshots currently hog the main thread from the moment the user requests one until the file is written to disk. see #1123 

this was noticeable as inputs getting dropped/ignored and cd based cores getting starved for data. worst case in my testing showed main stalled for up to 240ms when a screenshot was requested. This is now down to between 3 and 8ms, which is well within our frame budget and a significant improvement.

before
```
233513us spike over 1000us limit.
| co_poll                                            |  233513 |       8 |
|   user_io_poll                                     |      13 |      13 |
|   input_poll                                       |  233491 |     128 |
|     user_io_screenshot                             |  233363 |  143403 |
|       mister_scaler_read_32                        |   88593 |   88593 |
|       OsdUpdate                                    |    1367 |    1367 |
```

after
```
4031us spike over 1000us limit.
| co_poll                                            |    4031 |      23 |
|   user_io_poll                                     |      12 |      12 |
|   do_screenshot                                    |    3942 |     676 |
|     mister_scaler_read                             |    3238 |    3238 |
|     offload_add_work                               |      27 |      27 |
|   input_poll                                       |      53 |      13 |
|     input_test                                     |      39 |      39 |
```
the decrease in mister_scaler_read was the result of rewriting how the memory is copied, the overall reduction in user_io_screenshot was the result of offloading the remainder of screenshot work to a worker thread (which I think runs on the other core) so that it wouldn't stall main.

fixes:

* copying from the scaler to a local buffer was taking an unexpectedly long time (40-80ms). rewrote copy to use SIMD operations which sped up the copy by an order of magnitude. now takes between 3 and 8ms to complete, depending on core's output resolution.
* all screenshot processing after the initial copy is now offloaded to a worker thread to be completed without further stalling main. we also refuse to process further screenshots until the existing one has completed.
* /dev/MiSTer_cmd now supports scaled as an optional parameter to indicate that the screenshot should be scaled to the scaler's output resolution rather than kept at native.
* screenshots are tied to vsync via a callback to minimize the risk of the scaler's buffer changing before the copy was completed resulting in tearing/corruption in screenshots. autofire frame checks were moved to the same callback handler for consistency.
* the same callback now waits for the results as an atomic to safely call Info() since Info() is not threadsafe.
* user can choose between png or bmp format in mister.ini using screenshot_image_format. Continues to default to png.
* split screenshot handling from user_io to screenshot.cpp/h and image_save.cpp/h, trimming uiser_io.h/cpp down a bit.

other notes:
* copying from the scaler could probably be moved completely out of main to a worker, but i've not tested that so i'm not sure if that'd drastically increase the risk of getting tearing/corruption in screenshots since we'd have no control over when it started.
* removed the FRAME_TICK macro i previously added for autofire in favor of a callback handler. this should make it easier for anybody else to setup code that runs every vsync and ensures that any functions that depend on vsync are launched at the same time. (it was already checking for vsync every time it was called, so this should reduce overall overhead by not making each vsync function also check the framecounter for themselves.)
* the new screenshot handling is admittedly overengineered, using atomics to pass data and results between the callback and worker but i'm considering it practice in case we find other places that could benefit from offloading work -- we now have a known working pattern to copy.


tonytoon:
@Slamy did a round of testing with their WIP CDi core and reported that my test build completely removed all issues with the CD data cache getting starved when screenshots were taken. I've requested they take one more look at my final draft to verify I didn't introduce any regressions, then I'll switch this from draft.

Slamy:
I'm not sure about regressions in other parts of MiSTer Main. I can only speak for issues with the optical drive emulation.
As of ae8e2431 in mainline and before, doing a screenshot of the 1080p scaled framebuffer, stalls the HPS for about 95 CD sectors. That is more than a second. A raw screenshot stalls for 15 sectors. That is something, the CD-i core is able to catch using the 32kB of CD sector buffer in block RAM, which is holding 27 CD sectors. Even doubling the size would not fix the screenshot issue right now.
MiSTer Main stalling after plugging in USB devices is still an issue with a certain randomness, though. But that is not what this PR is about.

I've tested the state of b236220b. Doing scaled or raw screenshots have no impact on the FIFO level of my sector cache. So, I guess it seems to help in that regard. I'm not sure what else to test though. Screenshots still seem to take about 1 second to do. But during that time, input from USB devices are still forwarded to the core. So that is an improvement too.

tonytoon:
> I'm not sure about regressions in other parts of MiSTer Main. I can only speak for issues with the optical drive emulation. As of [ae8e243](https://github.com/MiSTer-devel/Main_MiSTer/commit/ae8e24319e5f61003d03bb5a98ad606dc6126017) in mainline and before, doing a screenshot of the 1080p scaled framebuffer, stalls the HPS for about 95 CD sectors. That is more than a second. A raw screenshot stalls for 15 sectors. That is something, the CD-i core is able to catch using the 32kB of CD sector buffer in block RAM, which is holding 27 CD sectors. Even doubling the size would not fix the screenshot issue right now. MiSTer Main stalling after plugging in USB devices is still an issue with a certain randomness, though. But that is not what this PR is about.
> 
> I've tested the state of [b236220](https://github.com/MiSTer-devel/Main_MiSTer/commit/b236220bb981a0c8a8c7c1c8c31c27d88f8557f4). Doing scaled or raw screenshots have no impact on the FIFO level of my sector cache. So, I guess it seems to help in that regard. I'm not sure what else to test though. Screenshots still seem to take about 1 second to do. But during that time, input from USB devices are still forwarded to the core. So that is an improvement too.

screenshots will still take time to process (which is unavoidable) but throwing them to a worker thread should make it so main doesn't get stuck waiting on them and blocking input/cd/etc. -- which it sounds like what you're seeing in your testing.

I'm going to switch this from draft to ready for review.

tonytoon:
switching back to draft to fix issue mentioned in conversation - writing the screenshot filename extension from cfg. could be incorrect if the user put garbage data and it's falling back to .png.

tonytoon:
FileGenerateScreenshotName() now requires the extension be passed to it - that way we can check for and correct invalid file extensions beforehand.

sorgelig:
May be it's better to reduce amount of new files? scaler, image_save and screenshot - are all for a single purpose - screenshot. Basically all these files being combined will have a single calling function - make_screenshot.  Isn't?

tonytoon:
Moved all screenshot related code back into scaler.cpp. Removed custom bmp/resize code I used during testing so everything goes through imlib2 which also reduced amount of code substantially.

PR #1129, opened 2026-03-19
Fix F12 and F12+modifier for OSD

zakk4223:


odiaboeeu:
I think this modification ended up breaking the ability to enter the OSD using controllers. Now I can only exit the OSD using the button mapped on the 8BitDo M30. It works in the menu, but not in the cores.

zakk4223:
Yes, incredibly dumb typo/editing mistake on my part :1st_place_medal: 
I will fix tonight

PR #1139, opened 2026-03-24
Atari800: reset fix for oscillating keyboard triggered reset

woj76:
I knew the problem existed already a few weeks ago when migrating the firmware to Main, I just couldn't wrap my head around it until now. The symptom was ugly and annoying, when resetting from keyboard F9/F10, while the key was pressed, even shortly, the reset signal was oscillating violently with very broken video and buzzing audio, this fixes it (essentially a de-bounce of a sort ;)).

EDIT: converting it to draft for a quick fix to the 5200 core too, totally forgot about that.

EDIT2: fixed.

### DIFF
--- a/support/atari8bit/atari5200.cpp
+++ b/support/atari8bit/atari5200.cpp
@@ -151,7 +151,7 @@ void atari5200_open_cartridge_file(const char* name, int match_index)
 
 	if (FileOpen(&f, name))
 	{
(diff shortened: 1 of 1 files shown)


woj76:
Converted this to draft again, for whatever reason the 5200 core got broken by this, the whole MiSTer gets reset on F10, I presume this is because of the C function overload.

woj76:
Third time is a charm, now it is tested and ready.

PR #1150, opened 2026-04-04
conslidate SPD/CSC hdmi packet transmission to reduce overhead

### LINKED ISSUE #1127: Info() causing main thread spikes.
I've noticed that calling Info() seems to cause spikes in the main thread. Attached is output from profiling. 11ms isn't a full frame, but could accumulate with other sources to cause an occasional missed input or other data loss. *Something* seems to happening as a side effect of Info() that is making user_io_poll do a bunch of work - twice.

```
11260us spike over 1000us limit.
| co_poll                                            |   11260 |      20 |
|   user_io_poll                                     |   11213 |   11208 |
|     set_vfilter                                    |       5 |       5 |
|   input_poll                                       |      26 |       5 |
|     input_test                                     |      21 |      21 |

Opened /dev/i2c-1 for device 0x38
Opened /dev/i2c-1 for device 0x39

11207us spike over 1000us limit.
| co_poll                                            |   11207 |      11 |
|   user_io_poll                                     |   11169 |   11164 |
|     set_vfilter                                    |       5 |       5 |
|   input_poll                                       |      26 |       7 |
|     input_test                                     |      19 |      19 |
```

I plan on digging into this one myself once I finish up my screenshot PR, but if anybody has looked at this before and has insights as to why this is happening, let me know.

tonytoon:
Consolidate sending of SPD and CSC data into sending a few packets with a block of data vs a packet for every byte of data. This should speed up transmission of those packets (and reduce blocking in main) by reducing overhead.

This might help with issue #1146 if it is related to SPD. It also partially resolves #1127

### DIFF
--- a/cfg.cpp
+++ b/cfg.cpp
@@ -137,6 +137,7 @@ static const ini_var_t ini_vars[] =
 	{ "AUTOFIRE_RATES", (void *)(&(cfg.autofire_rates)), STRING, 0, sizeof(cfg.autofire_rates) - 1 },
 	{ "AUTOFIRE_ON_DIRECTIONS", (void *)(&(cfg.autofire_on_directions)), UINT8, 0, 1 },
 	{ "SCREENSHOT_IMAGE_FORMAT", (void *)(&(cfg.screenshot_image_format)), STRING, 0, sizeof(cfg.screenshot_image_format) - 1 },
+	{ "DV1_ENABLE", (void *)(&(cfg.dv1_enable)), UINT8, 0, 1 },
 
 };
 
@@ -602,6 +603,7 @@ void cfg_parse()
 	cfg_error_count = 0;
 	strcpy(cfg.autofire_rates, "10,15,30");
 	strcpy(cfg.screenshot_image_format, "png");
+	cfg.dv1_enable = 1;
 	
 	ini_parse(altcfg(), video_get_core_mode_name(1));
 	if (has_video_sections && !using_video_section)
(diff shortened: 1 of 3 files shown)


sorgelig:
I don't like the idea to add an option for every small thing making it linux-like hard to manage settings. There is no point to disable SPD sending unless if it's absolutely required - this part needs more test to understand why QMTech board has problem. Try can you supply more compact PR avoiding old text formatting to provide more clear changes. Basically you could replace for cycle with i2c block sending which is just couple of lines.

znam1980:
**tonytoon**
I’ll post this here as well, since the thread about the Qmtech issue has grown so much that some messages are getting lost in the mix 🙂
I wanted to report that your MiSTer_spd_offloadv2 build helped me resolve the issue with OSD lag and the subsequent lag in other cores like PS1 and ZX Next.
It also helped another user, although not everyone.

Nevertheless, I’ve been using it for three days now, and there have been no issues at all.

tonytoon:
> **tonytoon** I’ll post this here as well, since the thread about the Qmtech issue has grown so much that some messages are getting lost in the mix 🙂 I wanted to report that your MiSTer_spd_offloadv2 build helped me resolve the issue with OSD lag and the subsequent lag in other cores like PS1 and ZX Next. It also helped another user, although not everyone.
> 
> Nevertheless, I’ve been using it for three days now, and there have been no issues at all.

I'm going to close and resubmit this PR - I didn't realize I was pushing stuff to an existing PR as I was working on it and it got pretty cluttered. I'm very bad at git, much better at C I promise :D @sorgelig 's comments on it being disjointed are much appreciated, I need to get better at making sure my PRs are concise.

That said - did the non-offload version I posted work as well? The offload version uses a worker thread which I'd prefer to avoid since it a) complicates the code and b) defers the sending of SPD packets to "whenever it gets around to it." and I'm not familiar enough with how time critical these packets are.

PR #1152, opened 2026-04-06
SPD optimizations and fixes

tonytoon:
changes hdmi_packet_set_data to use block writes when data is > 1 byte. will reduce number of SPD packets sent and overhead of multiple writes. falls back to original call when data only contains one byte.

partially helps #1127 and might help #1146

### DIFF
--- a/video.cpp
+++ b/video.cpp
@@ -11,6 +11,7 @@
 #include <sys/types.h>
 #include <unistd.h>
 #include <math.h>
+#include <time.h>
 
 #include "hardware.h"
 #include "user_io.h"
@@ -1128,14 +1129,25 @@ static void hdmi_packet_enable(uint8_t mask, bool enable)
 	}
 }
 
+static bool spd_dv_update_failed = false;
+
 static void hdmi_packet_set_data(uint8_t mask, uint8_t offset, uint8_t *data, int size)
 {
+	// certain qmtech boards have issues with i2c, causing excessive blocking
+	// if we see this happen once, we disable future spd updates.
+	struct timespec start_ts = {};
+	struct timespec end_ts = {};
+
+	bool timer_started = false;
+	
+	timer_started = (clock_gettime(CLOCK_MONOTONIC, &start_ts) == 0);
+
 	if (!data || size <= 0)
 	{
 		hdmi_packet_enable(mask, 0);
 		return;
 	}
-
+	
 	int fd = i2c_open(0x38, 0);
 	if (fd >= 0)
 	{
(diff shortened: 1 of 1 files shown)


sorgelig:
i think this change is ok even if it's not fully solve the problem. I don't know if i2c_smbus_write_block_data accepts lenght of 1 to avoid the check.

tonytoon:
Unfortunately, I don't think this will do too much to help with #1146. The test build I posted that helped the most offloaded SPD packets to a worker thread and I don't know enough about the timing of SPD packets to be sure that won't impact them negatively. It also didn't resolve it for all the users and I worry it just masks the problem rather than fixing it.

This PR *does* cut the overhead of SPD packets in half, which helps reduce the chances of main blocking in general. I can't recall the exact values now but it was several milliseconds.

In my testing the block write function worked fine with a data size of one but I don't have a good enough HDTV and don't use direct video so I couldn't fully test if that caused any regressions vs just falling back to writing a single byte which felt safer. I'll want some people with VRR or using DV1 to test this unstable when the PR is accepted.

tonytoon:
This additional commit does the following:

- Only sends DV/DV1 SPD updates if the data it is sending has actually changed
- If it detects an excessively long SPD update transfer (>= 32ms) it disables future SPD updates (the impacted qmtech boards were seeing over 2,000ms for these to complete.)

This has substantially improved behavior in issue #1146 for the users whose qmtech boards were impacted the worst.

sorgelig:
But there is a problem: it silently (user won't look into logs usually) disables SPD which may lead to bug reports related to external scalers or some TVs. So it must be visible. Probably through info() message. But this need to be carefully tested as it may not appear or hidden by subsequent info() calls due to resolution change.

And 32ms is too optimistic i think. Generally speaking, SPD should be sent only once per resolution change or OSD open/close. 

An improvements idea: when SPD gets disabled, disable it only for OSD open/close but keep it for resolution change as SPD is less important for OSD but still required for new resolution. Make sure menu_present flag in SPD is always cleared if it's disabled for OSD open/close.

tonytoon:
I am going to put this one on ice for a while. While it seems to have fixed the OSD lag for people that reported the issue, it's making external scalers not work properly. It's possible that's the result of the optimizations I made or possibly the check for whether SPD data has been changed. With neither an impacted board nor an external scaler, I'm not confident I can test this as well as we need - despite my enthusiasm to try. 

Until somebody can replicate the issue and has the hardware to make sure any fixes don't break existing functionality it might be easiest to fall back on the INI file option to simply disable SPD packets since that would let impacted users fix it for themselves without breaking any functionality for current users.

PR #1158, opened 2026-04-16
Mapping v4

misteraddons:
- Add L2/R2/L3/R3

- Separate Joystick and Mouse Mapping dialogs

- Update virtual joypad graphic and activity indicator for current button (flashing fill instead of flashing outline)
joymapv4

- Add hotkeys
  - Skip (hold Dpad right for 1 second - same as IO User button)
  - Previous (hold Dpad left for 1 second)
  - Save (hold Start for 1 second - core mapping only - same as keyboard Enter)
  - Clear (hold Select for 1 second - core mapping only - same as keyboard F12)
- Backwards compatible with V3 mappings (new button mappings are after previous button mapping slots in the .map file)


sorgelig:
1) new graphics is worse than original one, more cryptic as it has no button names
2) while adding analog triggers, are you sure you handled all possible variations? Some gamepads have inverted range, some share the same axis. Some have button + analog report - was it filtered out? Are digital triggers supported? do they emulate analog triggers? How about opposite to emulate a button from analog triggers, and so on. 
If you think i didn't add triggers support yet just because i'm lazy or don't want - you are wrong. It's just because there are so many different options - need to handle all variations...

Btw, your lookahead OSD list navigation is so broken! Strange that no one reported it. So, i have to fix it.
So i'm not confident in your changes, especially in such large change in very sensitive part. I don't want to fix it later digging in new logic.

misteraddons:
1) The name of the button is in the mapping queue. I like this look better but if you prefer button names, we can change that.
2) I didn't implement analog triggers. Only digital buttons for now.
3) How is the skip option broken? I tested myself and it works. Others tested too. What needs to be fixed?

misteraddons:
I tweaked the skip/previous mapping commands and replaced the virtual joypad buttons with ABXY

image

sorgelig:
i was talking about lookahead list navigation, not skip. It was other PR. I want to say, you don't test your changes well, so it requires further improvements.

btw, graphics is still bad.

misteraddons:
I'm sorry, I wasn't aware of issues with lookahead code. I did test it, and had others test as well. Can you tell me specifically what's wrong with lookahead?

Regarding this code, can you tell me what's wrong with the controller? I agree that having L2 and R2 on an SNES controller isn't accurate. You expressed disinterest in making the default controller PSX, which has a superset of buttons compared to SNES.

sorgelig:
Current controller graphics is good enough. L2 R2 can be added to the graphics. PSX and N controllers have fundamental change of A/B X/Y which doing it in already existing project isn't a good idea. It's not really SNES controller but some virtual controller agreed as a base before. Adding additional buttons to it isn't against rule. The problem is how to make mapping reliable.

As for lookahead - it simply doesn't work reliably. Cursor stops at 2nd/3rd row when you do paging up, and same at the end. Also i see page up/dn keys act a little different than LB/RB buttons - as cursor at first/last page stops at different rows.
Generally speaking, lookahead need to be rewritten. Amount of lookahead rows is not required as cursor should simply be in the middle row as long as there enough records above/below. At the end, last record must be always on the last row. There must be no empty lines at the end of list (as it is now).

misteraddons:
We can revert the virtual controller change but the original one is pretty crude. The dpad arrows are not nearly as nice as the new dpad. The blinking on/off rather than blinking fill/empty is also worse user experience in my opinion. The proportions of the old mapping controller area really unnatural. Here's a video of the current mapping process:

https://github.com/user-attachments/assets/c424286f-ffdc-423f-b5a6-d006d13fbe05


Does this look better for Lookahead? Keyboard page up/down and dpad left/right both behave the same:

https://github.com/user-attachments/assets/af0746c3-2b89-43aa-a584-c636bfbe4f84

sorgelig:
> Does this look better for Lookahead? Keyboard page up/down and dpad left/right both behave the same:

look better. How about up/down navigation?

misteraddons:
https://github.com/user-attachments/assets/dff44f20-740f-4914-a079-84c4abbc1952

misteraddons:
Any new thoughts on the input mapping code?

sorgelig:
There are no new thoughts, sorry

misteraddons:
Which parts do you dislike?
- R2, L2, R3, L3
- Separate joypad and mouse mapping dialogs
- Updated graphic
- Hotkeys (skip, back, save, clear)

sorgelig:
I don't like to change current mapping code because it's very sensitive thing and very easy to break something without thoroughly testing. 
R2/L2/R3/L3 are used in couple of cores and it can be easily mapped in those cores. For hundred of other cores it has no meaning. Those changes would have a reason if analog triggers will be supported, but as i've pointed out analog triggers support is complicated and needs more thinking. Graphics is worse than original. 
Hotkeys - this part probably can be improved in original mapping (btw, Menu button on gamepad can be used for skipping when buttons are get defined in cores).

misteraddons:
https://github.com/user-attachments/assets/456a6ed0-ed4f-4b21-8a6a-98c7657c06c1

Understood. You suggested you would accept a pull request for L2 and R2. 

I had several testers check out the functionality. This code is backwards compatible with V3 mappings. I put the new buttons at the end of the .map file, but changed the mapping wizard to prompt in a different order. I avoided analog trigger implementation based on your suggestion. Having L2 and R2 (at a minimum) allows gamecontrollerdb.txt be used for 6 button arcade games with pause. This is very interesting to me.

What about the graphics looks worse than the original? Have you built the binary and tested? The Dpad, shoulders, and filled-flash (compared to disappearing blinking) are all very helpful. I restored the SNES button names at your request.

The mouse button mapping seems inconvenient to have coupled with controller mapping. Breaking it out to a separate dialog really tightens up the feeling. Each mapping dialog only writes to the correct parts of the .map file.

The problem with using menu button to skip is that not all controllers have a menu button. Holding left or right feels very intuitive. There's a "Skipping" status message that appears when you start holding right.

Can we agree on any part of the implementation? Do you want analog triggers added? I know that there are lots of different standards for analog controls. Could we look for axis change during L2/R2 mapping to figure out the axis changes? I'm open to ideas but I think there's a lot of benefit to this code as is.

Bemjo:
I do hope something can be worked out on this, as being able to map L2/R2, L3/R3 would allow expanding the buttons available to the linux framebuffer for controller support, expanding what the controller can do in the PDFViewer for example by allowing L2/R2 for page rotation, L3/R3 for jumping to first/last page, for multi-page changes, or other functionality -- regardless of what core is actually loaded.

misteraddons:
Reworked in new PR https://github.com/MiSTer-devel/Main_MiSTer/pull/1201

PR #1163, opened 2026-04-21
video: add hdmi_spd option to disable HDMI SPD InfoFrames

drizzt:
## Summary

Adds a new `hdmi_spd` option to `MiSTer.ini` (default `1`) that lets users disable HDMI SPD InfoFrame transmission. Workaround for boards — notably QMTech clones — where SPD register writes to the ADV7513 are slow enough to cause visible OSD/menu lag (see #1146).

## Behavior when `hdmi_spd=0`

- `spd_config_update()` early-returns, skipping the DV1 and standard SPD InfoFrame payload construction and i2c writes triggered by video mode changes / OSD toggle.
- In `set_vrr_mode()`, Freesync VRR is treated as unavailable: autodetect (`vrr_mode=1`) falls through to VESA Forum VRR if available; forced Freesync (`vrr_mode=2`) leaves `use_vrr=0`. This keeps `use_vrr == VRR_FREESYNC` from being set when no Freesync signaling can actually reach the display, so the rest of the VRR state stays consistent.
- VESA Forum VRR (`vrr_mode=3`) is unaffected — it uses `hdmi_spare_config`, not SPD.
- `direct_video` still works over HDMI but loses the DV1 scaler metadata carried in the SPD InfoFrame.

Default `hdmi_spd=1` preserves existing behavior for everyone not affected.

## Test plan

- [x] Build passes on the ARM Linux cross toolchain (verified locally with `arm-none-linux-gnueabihf-g++`).
- [x] With `hdmi_spd=1` (default): normal operation on a DE10-Nano, Freesync and direct_video external scalers keep working.
- [ ] With `hdmi_spd=0` on an affected QMTech board: OSD opens responsively when F12 is hammered on the Minimig core; no lag buildup across core switches (Genesis, SNES, PS1, Amiga).
- [ ] With `hdmi_spd=0` and `vrr_mode=1`: autodetect selects VESA Forum VRR if the display supports it.
- [ ] With `hdmi_spd=0` and `vrr_mode=2`: VRR is disabled (no Freesync without SPD transport).

Fixes #1146

### DIFF
--- a/MiSTer.ini
+++ b/MiSTer.ini
@@ -25,6 +25,9 @@ hdmi_limited=0         ; 1 - use limited (16..235) color range over HDMI
 direct_video=0         ; 1 - enable core video timing over HDMI, use only with VGA converters.
                        ; 2 - auto-mode for HDMI DACs that report 1024x768 resolution (AG6200, CS5213). 
                        ; Power cycle when switching between HDMI displays and DACs. No hot-plug detection.
+hdmi_spd=1             ; 0 - disable HDMI SPD InfoFrames. Workaround for boards where SPD i2c writes
+                       ; are slow and cause OSD lag. Disabling loses Freesync VRR and direct_video
+                       ; scaler metadata; VESA Forum VRR (vrr_mode=3) still works.
 
 hdr=0                  ; 1 - enable HDR using HLG (recommended for most users)
                        ; 2 - enable HDR using the DCI P3 color space (use color controls to tweak, suggestion: set saturation to 80).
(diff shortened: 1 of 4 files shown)


sorgelig:
This feature is still in discussion.
If nothing new will be offered then i will make an option myself. I think your option is too cryptic.

tonytoon:
No real progress has been made in fixing the persistent HDMI/SPD issues on qmtech boards referenced in #1146. We assume it's bad hardware. I think several qmtech users are using a test build I uploaded that disables SPD in a similar way day-to-day to remove the issue. I support a PR like this - especially if the submitter does a better job than I did (aka, any job at all) of identifying and separating VRR from DV1 data just in case it's only one or the other that causes the issue.

Alternately, QMTech could replace the boards under warranty or they could respond to the messages people have sent them and submit their own PR but I suspect that's not happening.

PR #1169, opened 2026-04-26
Lookahead Update

misteraddons:
Lookahead now scrolls from the center list item instead of 2-3 positions from the top/bottom entry. MiSTer.ini has a 0/1 setting to disable/enable lookahead. Please test before accepting if you have concerns.


sorgelig:
by the way, did you check behavior when you page down and current items has 2 rows?

PR #1187, opened 2026-05-06
F7 = Joypad Map

misteraddons:
Keyboard F7 key at menu core or when OSD is open activates Joypad mapping.

### DIFF
--- a/menu.cpp
+++ b/menu.cpp
@@ -1323,6 +1323,13 @@ void HandleUI(void)
 			}
 			break;
 
+		case KEY_F7: // added: F7 activates joystick map if OSD is visible, or in menu core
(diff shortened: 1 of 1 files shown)


sorgelig:
what is the problem with specific menu item already exiting?
There are 2 drawbacks with this method:
1) you have to remember all these hot keys
2) you may run out of hotkeys, so it's better to use it when absolutely required.

misteraddons:
It's not required, but it can be handy if the IO board user button isn't accessible. Since bluetooth pairing has a hotkey, it makes sense that joy map also has a hotkey.

F1: change background (main menu)
F2: toggle core dates
F7: joy map
F9: terminal
F10: light gun map
F11: bluetooth controller
F12: osd

PR #1193, opened 2026-05-09
Allow custom USB devices to use bcdDevice for gamecontrollerdb matching

misteraddons:
Support devices that use the USB `bcdDevice` field to encode subversion info, such as the input mode of a multi-mode adapter.

This change allows MiSTer to use USB `bcdDevice` as the gamecontrollerdb version field for whitelisted USB hardware IDs. All other controllers continue using the normal Linux input device version, so existing gamecontrollerdb matching behavior is unchanged.

Tested with whitelisted devices and normal controllers to look for regressions. Normal USB HID controllers often report input device version `0x0111`, and should continue matching existing gamecontrollerdb entries.

### DIFF
--- a/input.cpp
+++ b/input.cpp
@@ -35,6 +35,7 @@
 #include "str_util.h"
 #include "frame_timer.h"
 #include "scaler.h"
+#include "file_io.h"
 
 #define NUMDEV 30
 #define UINPUT_NAME "MiSTer virtual input"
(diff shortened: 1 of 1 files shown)


sorgelig:
Do you use AI for writing code? It looks not optimized and inserted without checking what code already has. For example sysfs path is already present in structure. Using MAX_PATH arrays in stack 3 times doesn't look good as it may hiccup with such deep stack. resolved[] looks like a dummy huge array just to check if path exists. There is a function https://github.com/MiSTer-devel/Main_MiSTer/blob/71c4946ef6c7fefd844be18fd78c77124d0b2615/file_io.cpp#L835
which is probably what you need. 
When you add something to the code, you need to be aware of existing code and use existing functions instead of blindly add whatever AI suggested.

misteraddons:
Oh, good catch. I didn't know about FileExists() function. Does it look better optimized? I re-tested normal and special controller devices.

sorgelig:
much better now

PR #1194, opened 2026-05-11
Add HDMI-CEC support for compatible hardware

misteraddons:
Adds HDMI-CEC support for MiSTer systems with compatible hardware.

The DE10-Nano HDMI output does not support CEC directly, but some third-party MiSTer hardware can expose CEC through the HDMI path. This PR adds the core MiSTer-side support needed for those devices.

I was considering using a microcontroller to implement this, but HDMI port address of the display is embedded in HDMI DDC/EDID, which the microcontroller can't easily access.

Features:
- Adds HDMI-CEC initialization and message handling (device name = MiSTer)
- Adds optional config settings
  - hdmi_cec=1 (won't affect DE10-nano)
  - hdmi_cec_input_mode=1 (enable/disable remote input)
  - hdmi_cec_sleep=1 (turns off display after 15-60 minutes - requires video_off>0)
  - hdmi_cec_wake=1 (turns on display after sleep when controller activates)
- Supports basic TV/remote control interaction through CEC
  - Up, Down, Left, Right, Ok, Back (toggles OSD)
- Keeps behavior unchanged on hardware without CEC support
- No /tmp file writing, as requested

Please feel free to ask questions or make suggestions on improvements,

### DIFF
--- a/MiSTer.ini
+++ b/MiSTer.ini
@@ -22,10 +22,14 @@ rbf_hide_datecode=0    ; 1 - hides datecodes from rbf file names. Press F2 for q
 menu_pal=0             ; 1 - PAL mode for menu core
 hdmi_limited=0         ; 1 - use limited (16..235) color range over HDMI
                        ; 2 - use limited (16..255) color range over HDMI, for VGA converters.
+hdmi_cec=1             ; 1 - enable HDMI-CEC via ADV7513, 0 - disable.
+hdmi_cec_input_mode=1  ; 0 - ignore CEC remote input, 1 - accept mapped CEC keys.
+hdmi_cec_sleep=1       ; 1 - send CEC STANDBY after video_off minutes of no input activity. Requires video_off>0.
+                       ; video_off values 1-4 are presets: 1=15m, 2=30m, 3=45m, 4=60m.
+hdmi_cec_wake=1        ; 1 - on input after that idle period, send wake (ImageViewOn/TextViewOn/ActiveSource).
 direct_video=0         ; 1 - enable core video timing over HDMI, use only with VGA converters.
                        ; 2 - auto-mode for HDMI DACs that report 1024x768 resolution (AG6200, CS5213). 
                        ; Power cycle when switching between HDMI displays and DACs. No hot-plug detection.
-
 hdr=0                  ; 1 - enable HDR using HLG (recommended for most users)
                        ; 2 - enable HDR using the DCI P3 color space (use color controls to tweak, suggestion: set saturation to 80).
 fb_size=0              ; 0 - automatic, 1 - full size, 2 - 1/2 of resolution, 4 - 1/4 of resolution.
--- a/MiSTer.ini
+++ b/MiSTer.ini
@@ -22,11 +22,12 @@ rbf_hide_datecode=0    ; 1 - hides datecodes from rbf file names. Press F2 for q
 menu_pal=0             ; 1 - PAL mode for menu core
 hdmi_limited=0         ; 1 - use limited (16..235) color range over HDMI
                        ; 2 - use limited (16..255) color range over HDMI, for VGA converters.
-hdmi_cec=1             ; 1 - enable HDMI-CEC via ADV7513, 0 - disable.
+hdmi_cec=0             ; 1 - enable HDMI-CEC via ADV7513, 0 - disable.
 hdmi_cec_input_mode=1  ; 0 - ignore CEC remote input, 1 - accept mapped CEC keys.
-hdmi_cec_sleep=1       ; 1 - send CEC STANDBY after video_off minutes of no input activity. Requires video_off>0.
+hdmi_cec_power_on=1    ; 1 - on power-on, try once to make MiSTer the active HDMI input once EDID provides its CEC address.
+hdmi_cec_sleep=0       ; 1 - send CEC STANDBY after video_off minutes of no input activity. Requires video_off>0.
                        ; video_off values 1-4 are presets: 1=15m, 2=30m, 3=45m, 4=60m.
-hdmi_cec_wake=1        ; 1 - on input after that idle period, send wake (ImageViewOn/TextViewOn/ActiveSource).
+hdmi_cec_wake=0        ; 1 - on input after that idle period, send wake (ImageViewOn/TextViewOn/ActiveSource).
 direct_video=0         ; 1 - enable core video timing over HDMI, use only with VGA converters.
                        ; 2 - auto-mode for HDMI DACs that report 1024x768 resolution (AG6200, CS5213). 
                        ; Power cycle when switching between HDMI displays and DACs. No hot-plug detection.
--- a/audio.cpp
+++ b/audio.cpp
@@ -22,6 +22,7 @@ static uint8_t vol_att = 0;
 static uint8_t corevol_att = 0;
 unsigned long vol_set_timeout = 0;
 static int has_filter = 0;
+static int has_boost = 0;
 
 static char filter_cfg_path[1024] = {};
 static char filter_cfg[1024] = {};
@@ -31,6 +32,8 @@ static void setFilter()
 	fileTYPE f = {};
 
 	has_filter = spi_uio_cmd(UIO_SET_AFILTER);
+	has_boost = (has_filter & 2) && !is_menu();
+	if (!has_boost) corevol_att &= 7;
 	if (!has_filter) return;
 
 	snprintf(filter_cfg_path, sizeof(filter_cfg_path), AFILTER_DIR"/%s", filter_cfg + 1);
@@ -140,8 +143,11 @@ int get_volume()
 
 int get_core_volume()
 {
+	int boost = corevol_att & 0x60;
+	if (boost > 0x40) boost = 0x40;
 	corevol_att &= 7;
 	if (corevol_att > 6) corevol_att = 6;
+	corevol_att |= boost;
 	return corevol_att;
 }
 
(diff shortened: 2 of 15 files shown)


sorgelig:
There are several reasons why this feature is arguable:
1) MiSTer originally uses gamepad, so adding remote control gives nothing but "let it be". Current implementation of screen saver is more than enough for extended away mode. Turning off only TV is doubtful feature which will make you forget that other equipment isn't turned off yet.
2) it's based on intensive i2c/smb communication which already has problems for many users - that's why i had to add option to disable SPD which is just seldom sends the data. Here is non-stop spamming of i2c.
3) I've tried your previous implementation on MiSTer Pi which according to my measurements has CEC connected to ADV7513 but i couldn't make it work - i didn't see any CEC messages. CEC code also was delaying the Main from starting. I will try this PR later to see how it works.

misteraddons:
Thanks for taking a look at it. Maybe we can optimize this more to fix i2c bus contention. I don't think MiSTer Pi has HDMI CEC connected. Based on user feedback, QMTech does have CEC connected. I tested the code before submission on third party hardware and it displays "MiSTer" as the device name, lets the user navigate the menu with the TV's remote and fully sleep and wake the display.

I submitted this simplified version because you wanted me to remove the /tmp/IDLE status tracking from the previous code.

sorgelig:
I modified my original de10-nano: 
1) disconnected pin 1 of U34 to avoid grounding CEC when de10-nano is powered off
2) removed R248 and soldered R244, R242, D6
3) removed R247 and soldered R252 to enable CEC clock.
4) Added your patch to Main

CEC doesn't work. TV doesn't see the name of MiSTer. What i forgot to do?

sorgelig:
update: 
i found that 12MHz is absent. Probably firmware of EPM chip doesn't enable it, or one of CLK_MAX_24 or UB2_CLK_OUT clock sources coming through non-installed resistors to EPM needed to generate that 12MHz clock. 
Anyway, UB2_CLK_OUT is 12MHz, so i wired it to pin 32 of ADV7513, so now CEC block has 12MHz clock. But CEC still has no signs of life.

sorgelig:
update: CEC works now, i forgot connect one signal after earlier experiments.
Will test it more. It seems have some hiccups in Main startup.

sorgelig:
This is my report from couple hours of usage:
1) CEC definitely should not be enabled by default considering very little benefit vs problems it brings. CEC itself (regardless MiSTer) is a buggy feature according to internet, so many users search how to turn it off rather than on.
2) Current implementation produces a long delay in starting of Main, so you have to wait for several seconds before OSD appear (after you already see a wallpaper) vs instant OSD without CEC. This delay even breaks some cores. For example ao486 refuses to boot from mounted HDD on start. It requires reset of core to boot. I'm sure many other cores will misbehave due to this delay (due to CEC initialization).
3) When i turn everything on (TV+AV Receiver + MiSTer) at the same time, TV doesn't know about MiSTer. I have to reload Main (or load any core) to make TV aware about MiSTer. I don't know much bout CEC and how devices aware about each other. May be MiSTer can try to repeat to advertise itself after some time if didn't get reply from TV. Otherwise, in current state, if user turn on devices and go to drink a tea/coffee, then after 1 hour it come back just to find TV is still ON.
4) Button reaction is not good. It's more for "ok, you have no other option to control OSD" than "i want to use TV RC instead of taking the gamepad laying a meter away".
5) I didn't even measure how much time CEC polling steals from FPGA<->HPS communication. Something tells me, CD cores may dislike it. Probably moving CEC thread to other CPU core may fix this problem - not sure.

Funny fact: qmtech clone which is the only option for CEC from the box, has an issue with smbus communication, so even rare non-periodic SPD sending is problematic and most qmtech users would like to turn SPD (and CEC of course) off.

sorgelig:
> I was considering using a microcontroller to implement this, but HDMI port address of the display is embedded in HDMI DDC/EDID, which the microcontroller can't easily access.

could microcontroller sit on DDC bus and sniff the data to grab the port address and other required data?

misteraddons:
> > I was considering using a microcontroller to implement this, but HDMI port address of the display is embedded in HDMI DDC/EDID, which the microcontroller can't easily access.
> 
> could microcontroller sit on DDC bus and sniff the data to grab the port address and other required data?

It's possible an MCU monitoring DDC bus might work. 

Maybe we can improve this enough to do directly through the ADV7513. Will you try this new build with some improvements? 
- CEC init now runs after the menu has started instead of blocking early startup. The old fixed 3 second startup delay is gone; observed menu startup impact is about 0.5s.
- Added `hdmi_cec_power_on` MiSTer.ini option to try to wake the TV and switch to the MiSTer input once on startup. Tested on LG C1; behavior may vary by display.
- CEC work is throttled to a 20ms poll cadence (was previous every poll loop(!))
- MiSTer advertises its CEC identity/name once at startup, then only replies to name queries from other CEC devices.
- EDID retries back off from 15s to 30s, 60s, 120s, 240s, then 300s max to avoid DDC contention.
- If the startup input switch happens before a valid EDID was available, MiSTer does one short 1s follow-up EDID retry so resolution detection can recover.

[MiSTer.zip](https://github.com/user-attachments/files/27693082/MiSTer.zip)

sorgelig:
it would be better if you will add changes to your repo, so i will pick it there as a source change and will see how it's changed instead of testing a black box.

sorgelig:
I plan to make HDMI_TX_INT pin available for polling in HPS which will eliminate spamming of i2c bus. According to ADV7513 datasheet HDMI_TX_INT can indicate CEC packet reception.
It will require framework update for all cores, but I think updating just Menu will fulfill 99% of needs. 
Using i2c for polling is unacceptable anyway.

misteraddons:
Excellent

sorgelig:
i'v pushed changes to Main and Template.
use `fpga_get_hdmi_int()` for int polling. 1 means interrupt pin is active.

sorgelig:
btw, on my another MiSTer setup with original non-modified de10-nano and iPad display i get display off issue if CEC is enabled. I don't know what part of CEC initialization causes HDMI shutdown, but this is another reason why CEC should not be enabled by default.

misteraddons:
> btw, on my another MiSTer setup with original non-modified de10-nano and iPad display i get display off issue if CEC is enabled. I don't know what part of CEC initialization causes HDMI shutdown, but this is another reason why CEC should not be enabled by default.

Interesting - I wonder how this is happening since the CEC pin is normally grounded on the DE10-nano. I'm not sure how the message would even get to the display. Worth looking into later.

sorgelig:
Of course it won't go through CEC pin but ADV7513 could shutdown HDMI because of some registers programming.

misteraddons:
I implemented the CEC interrupts into main, and rebuilt the menu core. [menu.zip](https://github.com/user-attachments/files/27737026/menu.zip)

I tested it successfully on stock DE10-nano and CEC-capable hardware, with hdmi_cec=0 and hdmi_cec=1, without issue.

Will you test your iPad display again with your stock DE10-nano to see if hdmi_cec=1 still breaks video?

misteraddons:
[_Test.zip](https://github.com/user-attachments/files/27742807/_Test.zip)
ao486 and PSX with HDMI CEC interrupts

sorgelig:
iPad display still goes to off state at start if hdmi_cec=1. I will investigate it.

sorgelig:
i found why - i'm using DVI mode (to reduce pixel clock) but your code forces HDMI mode (register AF).
Do CEC really requires HDMI mode? If yes, then CEC should not be enabled when `dvi_mode=1` is used.

sorgelig:
what this part of code tries to achieve?
```
	if (!cec_hpd_pulsed)
	{
		ok &= main_reg_write(MAIN_REG_POWER2, 0x00);
		usleep(100000);
		ok &= main_reg_write(MAIN_REG_POWER2, 0xC0);
		usleep(100000);
		cec_hpd_pulsed = true;
	}
```
If it's trying to pulse HPD pin physically then it's impossible as ADV7513 is a HPD receiver, not sender. But that code gives 100ms delay of startup which is not desirable.

sorgelig:
I'm still trying to evaluate what we are going to achieve by CEC. If main purpose is to provide sleep function then it's already implemented through HDMI shutdown which is more standard than CEC and supported by wast majority of displays. Currently HDMI shutdown works in Menu core only but can be expanded to any core if required.
The only exclusive and somehow useful feature from CEC i see is using TV remote to control OSD. But it's so niche use. MiSTer is a gaming platform and gamepad is a main input device.

P.S.: CEC itself has a lot compatibility problems between manufacturers, that's why CEC breaker is a popular device. If you have several CEC devices then they compete each other and try to fight for display switching it sometimes to a wrong port or shutdown TV in inappropriate time. Overall it looks like a rabbit hole to a world of problems.

sorgelig:
MiSTer is still not detected by TV in my hardware setup till i restart Main. It advertise to early, before my TV is fully booted.
If i cycle the power on MiSTer after everything booted then it correctly detected and TV automatically switches to MiSTer.

But it raises another question: do i want my TV to automatically switch to MiSTer every time? Probably not. Basically i ask the same question to Android TV box connected to TV which also switches the input to itself when i turn on the main switch (where all Audio and Video devices are connected in living room). Modern technologies more and more are trying to decide for me which is not always what i want.

sorgelig:
i've created a separate branch for cec with all your changes.

sorgelig:
it still need further fixes and tweaks such as CEC advertisement on startup. Probably it requires to detect a monitor appearance then additional timeout before advertise.
Important note: my MiSTer is connected to AV receiver. Probably AV receiver provides some dummy EDID and other info till TV get ready, then probably EDID gets update according to TV. That's only my guess.

PR #1197, opened 2026-05-15
Hdmi cec

misteraddons:
I removed the HPD 100ms delay, and added a periodic CEC re-advertise after init.

As for the advantages/purpose of adding CEC to MiSTer:
1) Advertise MiSTer name to compatible equipment.
2) Sleep (already implemented with video_off) and wake displays.
3) Remote control inputs to MiSTer for updating, etc.
4) Switch inputs when MiSTer turns on.

Let me know what you think of the latest changes. Thanks!

### DIFF
--- a/MiSTer.ini
+++ b/MiSTer.ini
@@ -25,9 +25,9 @@ hdmi_limited=0         ; 1 - use limited (16..235) color range over HDMI
 hdmi_cec=0             ; 1 - enable HDMI-CEC via ADV7513, 0 - disable.
 hdmi_cec_input_mode=1  ; 0 - ignore CEC remote input, 1 - accept mapped CEC keys.
 hdmi_cec_power_on=1    ; 1 - on power-on, try once to make MiSTer the active HDMI input once EDID provides its CEC address.
-hdmi_cec_sleep=0       ; 1 - send CEC STANDBY after video_off minutes of no input activity. Requires video_off>0.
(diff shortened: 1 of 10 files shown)


sorgelig:
I didn't get what you changed as all of mentioned were already there. 
Also check my changes

misteraddons:
Whoops, I didn't see your updates. They work great! Do they work for you on your receiver and tv setup? How about the iPad display?

sorgelig:
in setup MiSTer -> AVReceiver -> TV, TV doesn't notice MiSTer till i restart the MiSTer. It seems AV Receiver acts as a TV when TV isn't ready but doesn't forward discovered CEC devices when TV becomes available, or i didn't find how it's done - probably there are some specific messages coming - need further investigation. 
If MiSTer connected directly to TV, then it gets discovered right from power on.
iPad display works fine after i removed forcing of HDMI mode.

misteraddons:
Is it safe to leave enabled now, or does it require more testing? One idea I had was to use HPD interrupts to reinitialize video. It would be especially useful with direct_video=2 where users can hot swap between HDMI displays and DACs. I'm not sure how the HPD interrupts work compared to CEC interrupts. Do you know if they use the same interrupt bit?

sorgelig:
I'm working on HPD now.
Currently it's aggressively tries to establish CEC link which isn't good for cases where CEC is impossible because it will continue to spam ADV7513 registers, so it's not good to make it enabled by default. If i will be able to make it like one-shot operation then we can talk about enabling by default.

Issue #1128, opened 2026-03-17
Serial communication of OSD state


Arrghus8:
Hello,

I'm wondering if it is currently possible or could become possible to access the current state (open/closed) of the OSD using serial commands. For example, would it be possible for the MiSTer to store the current state of the OSD in a file similar to how `/tmp/CORENAME` stores the currently loaded core? I could then make a script to monitor this file for changes and send serial commands, similar to how the TTY2OLED project sends commands when a new core is launched.

The use case I have in mind is in conjunction with an adapter that would automatically switch a native controller's connection between SNAC mode and via a USB adapter. The idea would be to automatically switch to SNAC mode once the appropriate core is launched and the OSD is closed. If the OSD is open (or if any other core is running), the controller would be connected via USB mode for accessibility.

Thanks!

PR #1198, opened 2026-05-16
Add option to log OSD visibility

Arrghus8:
Adds an option to create a file `/tmp/OSD_VISIBLE` that indicates whether the OSD is currently open (see https://github.com/MiSTer-devel/Main_MiSTer/issues/1128).

### DIFF
--- a/cfg.cpp
+++ b/cfg.cpp
@@ -97,6 +97,7 @@ static const ini_var_t ini_vars[] =
 	{ "SHMASK_MODE_DEFAULT", (void*)(&(cfg.shmask_mode_default)), UINT8, 0, 255 },
 	{ "PRESET_DEFAULT", (void*)(&(cfg.preset_default)), STRING, 0, sizeof(cfg.preset_default) - 1 },
 	{ "LOG_FILE_ENTRY", (void*)(&(cfg.log_file_entry)), UINT8, 0, 1 },
+	{ "LOG_OSD_VISIBLE", (void*)(&(cfg.log_osd_visible)), UINT8, 0, 1 },
 	{ "BT_AUTO_DISCONNECT", (void*)(&(cfg.bt_auto_disconnect)), UINT32, 0, 180 },
 	{ "BT_RESET_BEFORE_PAIR", (void*)(&(cfg.bt_reset_before_pair)), UINT8, 0, 1 },
 	{ "WAITMOUNT", (void*)(&(cfg.waitmount)), STRING, 0, sizeof(cfg.waitmount) - 1 },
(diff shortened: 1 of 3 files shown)


sorgelig:
LOG_OSD_VISIBLE option is redundant. You can follow LOG_FILE_ENTRY option.

Arrghus8:
Unless I'm mistaken, the `LOG_FILE_ENTRY` option works slightly differently. Namely, that option writes one of three states to `/tmp/FILESELECT`:
 1. `active`: The file browser is being used (e.g. when loading a rom, core, filter, etc.)
 2. `selected`: Rom/core has been selected from the file browser and is being loaded
 3. `cancelled`: The file browser is not being used

Note: the behavior of the `selected` state is somewhat inconsistent. If one uses the file browser to launch a core, the state changes from `active` to `selected` and remains in this state until the file browser is opened again (at which point the state changes back to `active`). In this situation, the `cancelled` state only occurs after one enters and exits the file browser. On the other hand, if one loads a rom, the state changes from `active` to `selected` and reverts to `cancelled` as soon as the OSD is opened.

In any case, none of these three states are able to identify whether the OSD is open or not, only whether the file browser is being used. Specifically, it is possible to have the OSD open with any of the three states set. If it is preferable, the `active` state could be modified to be set as soon as the OSD is open (whether or not one is using the file browser), but I don't know how these settings are used in external integrations or whether this would cause issues.

sorgelig:
i mean,  you can write to /tmp/OSD_VISIBLE according to LOG_OSD_VISIBLE state.

Arrghus8:
My preference would be to maintain an option to log only the OSD visibility without the additional writes from the current `log_file_entry=1` behavior.

Would this be acceptable?
 - `log_file_entry=0`: No logging (same as current behavior)
 - `log_file_entry=1`: Log file browser status to `/tmp/FILESELECT`, `/tmp/STARTPATH`, `/tmp/FULLPATH`, `/tmp/CURRENTPATH` (same as current behavior)
 - `log_file_entry=2`: Log OSD visibility to `/tmp/OSD_VISIBLE` 
 - `log_file_entry=3`: Log file browser status to `/tmp/FILESELECT`, `/tmp/STARTPATH`, `/tmp/FULLPATH`, `/tmp/CURRENTPATH` and OSD visibility to `/tmp/OSD_VISIBLE`

sorgelig:
I prefer simplicity.
Too many options already - hard to follow

Arrghus8:
I've removed the `log_osd_visible` option and reused the `log_file_entry` option.

Issue #723, opened 2023-01-02
Feature Request: When mapping controllers in Main, allow skipping analog X axis input and only assign Y.


M-Walrus:
Currently when assigning analog sticks in the Main controller mapping screen, if you skip the X-axis for either stick, it will skip the Y-axis as well.  This should be changed so that only the Y-axis can be assigned, for compatibility with many analog control devices that have a throttle.

This is important for devices like the Cyber Stick that only have a Y-axis for throttle, with an example given here by "T4YK" on Twitter:
https://twitter.com/T4YK_68000/status/1609960847660679168

With XE-1AP support now added, this would be needed if someone used a USB adapter for the original controller as well, as it only has Y-axis on the second stick:  https://segaretro.org/XE-1_AP

This will also be helpful when XE-1AP support is added to other cores like Mega Drive, Sega CD, 32X, MSX and X68000.

Koston-0xDEADBEEF:
I ran into the same issue, trying to get throttle working on a CH Fighterstick. It's possible to do it by hex editing the input map file directly, but you need to insert a bogus event number for the X axis - specifying only Y axis doesn't work.

```
00000060  00 00 02 00 01 00 02 00  03 00 02 00 02 00 02 00
```

Bytes at 0x68 and 0x6c are the joystick event numbers for X and Y axis.

PR #1201, opened 2026-05-21
Joypad Mapping v4 with Analog Trigger Support

misteraddons:
- Add v4 joypad mapping with L2, R2, L3, and R3
- Preserve v3 map compatibility by appending new slots after existing v3 layout
- Infer analog L2/R2 triggers during normal L2/R2 mapping prompts
- Support separate, inverted, and centered/shared trigger axes
- Keep digital L2/R2 fallback mappings alongside analog trigger mappings
- Add gamecontrollerdb analog trigger support for aN, +aN, and -aN
- Send normalized 0..255 analog trigger values to cores via UIO_ATRIG
- Split joypad and mouse mapping into separate menu entries
- Keep upstream-style virtual joypad graphics in the mapping dialog

Tested with [InputTest_20260521.zip](https://github.com/user-attachments/files/28124984/InputTest_20260521.zip) (PR https://github.com/MiSTer-devel/InputTest_MiSTer/pull/18)

### DIFF
--- a/gamecontroller_db.cpp
+++ b/gamecontroller_db.cpp
@@ -45,6 +45,10 @@ static const char *sdlname_to_mister_idx[] = {
 	"asysy",
 	NULL,
 	NULL,
+	"lefttrigger",
+	"righttrigger",
+	"leftstick",
+	"rightstick",
 };
 
 typedef struct {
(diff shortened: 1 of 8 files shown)


sorgelig:
Considering how much time i spent to make CEC from concept to a really usable state, i'm already afraid to accept such bug change from you. 
I have to check the code on some test build to see what it does...

sorgelig:
Allow Y-only analog stick mapping - what is this for?

misteraddons:
It's for rare devices that only have a y-axis (like a throttle). Relevant to this issue: https://github.com/MiSTer-devel/Main_MiSTer/issues/723

misteraddons:
Have you had a chance to look at this?

sorgelig:
I plan to look at it in the end on June

misteraddons:
Checking in on this :)

PR #1203, opened 2026-05-25
Reduction of memory footprint

Slamy:
- Removes ~100kB of static memory per CD supporting core for CUE file loading

This change might be controversial. It already got negative feedback on the MiSTer Discord from other devs since it uses C++14 methods of memory allocation.

I could think of an alternative, like a global buffer for this task. Many of the CD cores don't really share code, but I believe this is a small change to at least reduce memory usage in case it might be a problem in the future.

### DIFF
--- a/cd.h
+++ b/cd.h
@@ -4,6 +4,7 @@
 #include <libchdr/chd.h>
 #include "file_io.h"
 
+#define CUE_BUFFER_SIZE (100 * 1024)
 
 typedef enum
 {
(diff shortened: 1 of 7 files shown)


Kitrinx:
The question I would ask here is what actual problem is this solving? Are we getting close to running out of RAM? Changes for the sake of changes without solving a real issue are usually undesirable and can potentially introduce unforseen bugs.

Slamy:
I do have to admit that this change actually fixes no real life issues right now since we probably have enough RAM for now. It was mostly a suggestion based on a sudden notice when I've discovered that this kind of buffer existed multiple times in the MiSTer application. I don't even know how many of these buffers even exist that have a single use purpose and then never again.

Edit: It was meant to show that there are nicer solutions.

PR #1208, opened 2026-06-05
Add serial bridge support for USB memory card adapters

misteraddons:
Lets MiSTer use physical saves exposed by USB memory card adapters on N64, PSX, and Game Boy Color (N64 Transfer Pak).

The bridge discovers compatible serial devices (whitelisted), scans for cards, mounts physical card images into MiSTer save slots, and flushes dirty writes back to the adapter while preserving normal local-save fallbacks.

When a supported physical card is detected, the serial memory-card bridge uses it in place of the matching virtual save file for that slot. If no compatible physical card is found, MiSTer falls back to the normal virtual memory-card or save-file behavior.

Memory cards load in the background during the core boot and are non-blocking (except for Game Boy Color Transfer Pak which waits for the ROM to load before confirming whether the game save can be applied.

When `CONTROLLER_INFO` is enabled in MiSTer.ini, supported cores show memory-card detection through MiSTer's Info() popup path. 
 - PSX shows detected physical memory card mounts
 - N64 folds memory-card detection into the existing game information popup
 - Game Boy / Game Boy Color shows the Transfer Pak save mount after ROM loading completes (and only loads the save file if the rom header matches the Transfer Pak cartridge header)

The serial protocol is line-oriented and uses CARD commands:
- CARD SCAN discovers attached cards and reports type, port, slot, block count, block size, read-only state, and auxiliary flags.
- CARD READ and CARD READRANGE read individual blocks or contiguous ranges from PSX and N64 memory cards.
- CARD WRITEBEGIN / CARD WRITECHUNK / CARD WRITECOMMIT write a single block, with CARD WRITEABORT for cancellation.
- CARD WRITEBLOCKSBEGIN / CARD WRITEBLOCKSCHUNK / CARD WRITEBLOCKSCOMMIT write contiguous dirty block ranges, with CARD WRITEBLOCKSABORT for cancellation.
- CARD GBINFO reads Transfer Pak cartridge identity so GB/GBC saves can be matched to the loaded ROM.
- CARD GBREADSAVERAW reads Transfer Pak save RAM as one raw save image.
- CARD GBWRITESAVEBEGIN / CARD GBWRITESAVECHUNK / CARD GBWRITESAVECOMMIT flush dirty GB/GBC save RAM back through the Transfer Pak.

Tested with Reflex Adapt Classic2USB

### DIFF
--- a/input.cpp
+++ b/input.cpp
@@ -36,6 +36,7 @@
 #include "frame_timer.h"
 #include "scaler.h"
 #include "file_io.h"
+#include "support/serial_memcard/serial_memcard.h"
 
 #define NUMDEV 30
 #define UINPUT_NAME "MiSTer virtual input"
@@ -5620,6 +5621,7 @@ int input_test(int getchar)
 
 			if ((pool[NUMDEV].revents & POLLIN) && check_devs())
 			{
+				serial_memcard_rescan_async();
 				printf("Close all devices.\n");
 				for (int i = 0; i < NUMDEV; i++) if (pool[i].fd >= 0)
 				{
--- a/menu.cpp
+++ b/menu.cpp
@@ -2097,6 +2097,15 @@ void HandleUI(void)
 							//option handled by HPS
 							if (p[1] == 'X') p++;
 							uint32_t x = user_io_status_get(p + 1, p[0] == 'o');
+							if (is_psx() && psx_is_memcard_source_option(p + 1))
+							{
+								uint32_t normalized = psx_memcard_source_normalize(p + 1, x);
+								if (normalized != x)
+								{
+									x = normalized;
+									user_io_status_set(p + 1, x, p[0] == 'o');
+								}
+							}
 
 							// get currently active option
 							substrcpy(s, p, 2 + x);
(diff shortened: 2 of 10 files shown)


sorgelig:
This is absolutely redundant and unneeded complexity.

misteraddons:
This was an effort to move some of SNAC's exclusive feature (memory card reading and writing) over to USB route for greater compatibility and less reliance on SNAC for things that can be handled over USB. It works seamlessly to the user. Will you reconsider?

sorgelig:
SNAC is originally made for such non-standard inputs or other devices. It doesn't require support in Main. So, if memory card is supported in some cores - then it's fine considering very niche use.

misteraddons:
I agree that SNAC is for non-standard inputs. SNAC was created before we had 1ms USB polling, but now we have very fast USB controller adapters that make SNAC less important for controllers. That leaves memory cards and light guns as SNAC-exclusive features. We'll never get accurate USB light guns, but we can add USB memory card support. I'm trying to add parity to USB to reduce dependency on SNAC.

Issue #1207, opened 2026-06-04
Black screen with MiSTer_20260603


marcomto:
Hello,

after updating with update_all.sh, the new MiSTer executable (MiSTer_20260603) causes a black screen at boot. The POWER and USER LEDs keep flashing.

Linux is still running and reachable via SSH. I verified that replacing the new executable with the previous version (MiSTer_20260325) immediately restores normal operation and video output.

Additional information:

OS version: 250402
update_all.sh completed and rebooted automatically
new menu.rbf was also installed
new MiSTer_example.ini was installed
menu.rbf works correctly with the older MiSTer_20260325 executable
old executable had been renamed to .MiSTer.old by the updater
MiSTer_20260325 size: 991 KB (working)
MiSTer_20260603 size: 1011 KB (black screen)

MD5 checksums:

MiSTer_20260325: b5cb22e690dccd2cff6d9a567af9de6c
MiSTer_20260603: 5a6fd2f095002acbb1da372c0e84c5ae

Since the same hardware, OS, and menu.rbf work correctly with MiSTer_20260325, the issue appears to be specific to the MiSTer_20260603 executable.

Thanks.

marcomto:
Edit:
Hardware: Cyclone V-based MiSTer-compatible clone (not an original DE10-Nano).
HDMI output connected directly to a PC monitor.
No HDMI adapters or converters.
VGA IO board installed but not in use.
No MiSTer.ini present (using default settings).

birdybro:
Just as a test, can you download this --> https://github.com/MiSTer-devel/SD-Installer-Win64_MiSTer/blob/master/release_20250402.7z - and then unzip it. Then delete the linux folder on your microSD card, and replace it with the linux folder in this one and boot the MiSTer.

sorgelig:
Try to set specific supported video mode instead of auto detected. May be MiSTer fails to autodetect the supported video mode.

marcomto:
I performed the test you suggested.

I used a new microSD card and created a fresh installation using the utility included in release_20250402.7z. After that, I copied all files from the old /media/fat partition to the new card.

Results:

* The older MiSTer executable worked correctly.
* I then replaced it with MiSTer_20260603.
* MiSTer_20260603 also worked correctly.
* Video output is working normally over HDMI.
* The system boots correctly and the menu is displayed.

[...]

PR #1210, opened 2026-06-08
video: add hdmi_int option to disable HDMI interrupt (hotplug/CEC).

pablopen:
Related to #1207 as I was having the same issue.

**What I observed:** On my board (a Hamgeek clone of the DE10-nano), the MiSTer_20260603 binary gives a black screen on a cold power-on. The disk/user (activity) LED does not light up. A warm reset (reset button) or relaunching the binary over SSH brings the picture back and the menu works normally — only cold boot is affected. 

If I boot without HDMI connected and plug it in afterwards, the picture appears but the menu becomes unresponsive. Reverting to MiSTer_20260325 always works. The same menu.rbf is used in all cases.

**What I think is happening (not 100% certain):** The HDMI hotplug/CEC handling added in 20260603 (video_poll()) runs whenever hdmi_has_int() reports true, which depends on the UIO_HDMI_INT probe and GPI bit 20. On my board this path appears to act on HPD/Monitor-Sense reads in a way that, during the cold-boot window before the HDMI link is stable, either powers the output down (video_hdmi_power(0)) or repeatedly re-initializes video — which matches the black screen and the unresponsive menu. On a warm reset the link is already up, so it doesn't trigger. I haven't been able to verify the exact register behavior on other boards, so I can't say how widespread this is.

**Proposed change:** Add an HDMI_INT ini flag. With it off (default), hdmi_has_int() stays disabled and the board behaves like 20260325, including when there's no MiSTer.ini. Setting HDMI_INT=1 opts back into hotplug detection and CEC. This is meant as a conservative, low-risk escape hatch rather than a behavior change for working boards.

I realize defaulting the new CEC/hotplug feature off may not be ideal for boards where it already works. As a next step, I plan to try a follow-up fix on top of this one to see if I can get the hotplug/CEC path working correctly on my board (for example, handling the cold-boot timing in video_poll() or making the probe more robust). This is my first contribution here, so please let me know if you'd prefer a different approach.

### DIFF
--- a/MiSTer.ini
+++ b/MiSTer.ini
@@ -28,6 +28,7 @@ hdmi_cec_power_on=1    ; 1 - on power-on, try once to make MiSTer the active HDM
 hdmi_cec_sleep=0       ; 1 - send CEC STANDBY after video_off minutes of no input activity. Requires video_off>0.
 hdmi_cec_wake=0        ; 1 - on input after that idle period, send wake (ImageViewOn/TextViewOn/ActiveSource).
 hdmi_cec_clock=0       ; 0 - auto-detect 12MHz/50MHz, otherwise floating point value from 3.0 to 100.0 (3-100MHz).
+hdmi_int=0             ; 0 - disable HDMI interrupt (keep 0 if no image on cold boot), 1 - enable. CEC also requires this =1.
                        ;
 direct_video=0         ; 1 - enable core video timing over HDMI, use only with VGA converters.
                        ; 2 - auto-mode for HDMI DACs that report 1024x768 resolution (AG6200, CS5213). 
--- a/cfg.cpp
+++ b/cfg.cpp
@@ -58,6 +58,7 @@ static const ini_var_t ini_vars[] =
 	{ "HDMI_CEC_INPUT_MODE", (void*)(&(cfg.hdmi_cec_input_mode)), UINT8, 0, 1 },
 	{ "HDMI_CEC_POWER_ON", (void*)(&(cfg.hdmi_cec_power_on)), UINT8, 0, 1 },
 	{ "HDMI_CEC_CLOCK", (void*)(&(cfg.hdmi_cec_clock)), FLOAT, 0, 100 },
+	{ "HDMI_INT", (void*)(&(cfg.hdmi_int)), UINT8, 0, 1 },
 	{ "KBD_NOMOUSE", (void*)(&(cfg.kbd_nomouse)), UINT8, 0, 1 },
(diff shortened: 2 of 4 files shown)


pablopen:
Repro and annotated ADV7513 register dumps for the Hamgeek/QMTech clone are in #1207. With the path on, the ADV7513 reports a powered HDMI link in HDMI mode (HPD/Monitor-Sense high, valid EDID) but the clone shows no image on a cold boot; `HDMI_INT=0` (the default this PR adds) skips the new `video_poll()` path so the board behaves like 20260325, and `HDMI_INT=1` opts back in.

sorgelig:
It's better to fix the problem instead if swiping under the table. Since you are capable to write the code and read the datasheet and have this problem, it would be good to understand on which stage it gets problem. May be it stuck in waiting for EDID ready flag? then adding a timeout and retry may fix the problem without disabling the feature.

pablopen:
Thanks, I see your point — a proper fix is better than just hiding it, however right now for me this is a broken build which is worse. so... would you also consider including the opt-out as a hotfix in the meantime, for new users? That was my case: I followed a tutorial, updated, and ended up with a black screen and no obvious way back. It defaults to off, just restoring the pre-20260603 behavior (HPD hotplug re-detection and CEC are both new this release), so it removes nothing people relied on, and anyone who wants live re-detection can still enable it.


On the cause itself: I could only get a log by booting with HDMI unplugged (otherwise SSH never comes up), then connecting the cable — when I did, the EDID read succeeded and the chip reported a fully powered link, but still no picture. I haven't been able to capture the actual cold-boot hang yet, since SSH never comes up and the log is overwritten on the next boot, so I'll try writing the log to the SD card so it survives, see where it stalls, and give the fix a second try. If it can't be made reliable on my board, I'll try to document why hotplug/CEC can't work there.

drizzt:
this seems more coherent: https://github.com/MiSTer-DB9/Main_MiSTer/commit/hdmi-hotplug-harden
can you test it?

sorgelig:
I cannot test it for several days as i'm away from my computer, but as i see you still use ini option even with your latest fix. Doesn't it solve the problem?

pablopen:
I think there is some confusion, we probably need a new PR with @drizzt changes, if that’s the direction we could then close this one. 
I tested his first commit and looked ok. I got some graph glitches in the menu once, but then I couldn’t reproduce that issue anymore.
I still have pending testing the second one with refinements. 
I only tested the default behaviour tho (no cec) so I don’t know if cec can even work in this boards knowing the limitations

drizzt:
Proposed that in #1212

PR #1212, opened 2026-06-10
video: harden HDMI hotplug state machine

### LINKED ISSUE #1207: Black screen with MiSTer_20260603

drizzt:
This fixes the cold-boot black screen / frozen menu regression reported in #1207 at its source, instead of disabling the new hotplug/CEC path as proposed in #1210.

**What was happening:** the blocking EDID re-read on cold boot fired for every sink whose link wasn't stable yet (or that has no readable EDID at all — DVI, VGA DACs), which froze the menu and gave the black screen on clone boards.

**Changes** (all in `video.cpp`, contained to the interrupt/hotplug poll path):

- Defer the first sample after an `hdmi_off` wake: Monitor Sense reads low for a moment after the power-up write, so adopting it immediately turned every wake into a spurious reinit. HPD edges seen while the TX is down are recorded, and EDID is re-probed at wake only when the display may have changed.
- Decay the reinit burst limiter after a quiet minute so sinks that hold HPD high and toggle only Monitor Sense (AVRs, input switches) aren't suppressed forever; the anti-feedback cap is kept.
- Pre-adopt the link state in `video_init` so the first-sample EDID recovery fires only for a link that came up late, not on every boot for sinks with no readable EDID — those paid a second blocking EDID read.
- Latch interrupt edges until a sample consumes them so a pulse completing inside a blocked poll gap (or across a transient i2c read failure) isn't lost; 0x96 latches are serviced in polled fallback mode too.
- Keep the stuck-INT demotion across reinit by feeding `int_pin_usable` into the 0x94 mask computation; don't demote on latched CEC (0x97) interrupts; count phantom interrupts cumulatively so a periodically re-latching bit is still detected.
- Rate-limit the settle and stuck-detection i2c reads instead of hammering the bus every `co_poll` iteration; treat a 0x96 read error as no-flags so the polled paths still run.
- Flush self-induced latches after a reinit so the link bounce it causes doesn't read back as a fresh hotplug and chain another reinit.

This should make hotplug detection and CEC work on non-genuine ADV7513 chips and edge-case sinks without an opt-out flag.

Fixes #1207.

### DIFF
--- a/video.cpp
+++ b/video.cpp
@@ -1411,10 +1411,36 @@ int hdmi_has_int()
 	return has_int;
 }
 
+#define HPD_SETTLE_MS        250
+#define HPD_SETTLE_SAMPLE_MS 25
+#define HPD_REINIT_GAP_MS    2000
+#define HPD_REINIT_BURST_MAX 3
+#define HPD_REINIT_DECAY_MS  60000
+#define HPD_FALLBACK_POLL_MS 1000
+#define INT_STUCK_WINDOW_MS  3000
+#define INT_STUCK_MIN_CNT    20
+#define INT_STUCK_POLL_MS    100
+
+enum { HPD_IDLE = 0, HPD_SETTLE };
+static int           hpd_sm_state = HPD_IDLE;
+static int           hpd_level = -1, ms_level = -1;   // -1 = not yet sampled
+static unsigned long hpd_settle_tmr = 0, hpd_sample_tmr = 0, hpd_fallback_tmr = 0;
+static unsigned long hpd_reinit_guard = 0, hpd_reinit_decay = 0;
+static int           hpd_reinit_burst = 0;
+static bool          hpd_reinit_suppressed = false;
+static bool          hpd_edge_pending = false;
+static bool          hpd_edge_asleep = false;
+static bool          int_pin_usable = true;
+static unsigned long int_stuck_tmr = 0, int_stuck_poll_tmr = 0;
+static int           int_stuck_cnt = 0;
+static bool          hdmi_tx_on = true;
+
 static void hdmi_config_init()
 {
 	int ypbpr = (cfg.vga_mode_int == 1) && (cfg.direct_video == 1);
-	uint8_t int0 = hdmi_has_int() ? 0xC0 : 0x00; // HPD + SENSE
+	// int_pin_usable: a runtime stuck-INT demotion must survive reinit,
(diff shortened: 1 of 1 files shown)


sorgelig:
I had no time to check very precise this PR. Now, after checking, i see it uses polling through i2c which is unacceptable at all because:
1) clone boards are proven to be glitching in HDMI i2c access. That's why i even added spd_quirk to minimize i2c usage. Here your code adds periodic access to i2c
2) i2c itself is pretty much blocking function which will produce different problems in cores, especially in CD cores.

machine-check-exception:
I have just tested on Mister 260823 and Os 250402. The problem persists. The only way of exit that status (besides reboot) is pressing the OSD button and make the bluetooth menu appears. It is not possible to make the bluetooth menu appears pressing F11 because no input is recognized in that state.
Another thing that happens and that can be related is that when the psx core is running (for example the demo), and the monitor is turned off and on (or switched inputs),  just before the signal comes back, the sound interrupts for a moment. In this experiment the speakers are hooked directly to the audio jack output in the mister itself, not via hdmi. I can't check if also the image freezes for a short period because I don't have in hand a CRT. It is not happening in another cores, for example, sf2: world warrior.
The mister.ini is next:
```
[MiSTer]
video_info=2
controller_info=6
vsync_adjust=0
preset_default=Display Specific/Sony PVM
```

it also happens with this config line added:
```
video_mode=1720,720,60
```

PR #1213, opened 2026-06-10
Fix failed EDID Detection loop

TheJesusFish:
First of all, you probably should not merge this as it's all vibe coded and I don't really understand what I'm doing. I'm only adding it as a PR so that you can see where the bug was.

Basically the gist is that the new EDID detection added with the CEC stuff would loop if it failed to detect a valid EDID, and it could hold the main thread hostage for 12 seconds.

If you want to try to reproduce it, this is my silly setup:
- 2 MiSTers connected to an HDMI display via switch, and a CRT (via different inputs)
- If you are using MiSTer A on the CRT, but have the HDMI switch set to MiSTer B, MiSter A will never get a valid EDID, so any time you start a core or go back to the menu, main would stall for (up to) 12 seconds, and make the MiSTer unusable for that time.

This is pretty low priority, I think, because it probably doesn't show up anywhere else other than in stupid setups like mine (or a setup that somehow gums up the EDID communication). But it was picking at my brain worms, so I wanted to see if it was fixable.

### DIFF
--- a/video.cpp
+++ b/video.cpp
@@ -1793,14 +1793,13 @@ static void cache_raw_edid_mfg_id()
 	raw_edid_mfg_id_valid = true;
 }
 
-static void read_edid_segment(uint8_t segment, uint8_t *buf)
+static bool read_edid_segment(uint8_t segment, uint8_t *buf)
 {
 	i2c_smbus_write_byte_data(hdmi_main_fd, 0xC4, segment);
 	i2c_smbus_write_byte_data(hdmi_main_fd, 0xC9, 0x03);
 	usleep(1000);
 	i2c_smbus_write_byte_data(hdmi_main_fd, 0xC9, 0x13);
 
-	bool edid_ready = false;
 	unsigned long timeout = GetTimer(500);
 	while (!CheckTimer(timeout))
 	{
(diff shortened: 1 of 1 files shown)


TheJesusFish:
Forgot to include Claude's "assessment" of the issue(s)

1. I2C error treated as "EDID ready" — the status register read returns a negative number on I2C failure, and the bitwise-AND check against that negative number always evaluates as true. So any I2C error looked identical to a successful "EDID is ready" signal, causing the chip to read garbage data from the EDID shadow registers.

2. 256-byte read happened even when the ADV7513 never finished fetching the EDID — after the 500 ms polling timeout expired with no interrupt, the code fell through and read all 256 EDID bytes anyway. If the DDC fetch hadn't completed, this just read stale or zeroed data, wasting time and producing a meaningless retry.

3. The retry loop kept going even when the display had no DDC at all — this was the main stall. The loop retried up to 20 times, each with a fresh 500 ms wait, totalling ~12 seconds. It had no way to tell the difference between "display is slow to respond" and "display has no EDID at all" (DAC, passive adapter, VGA monitor). Once the ADV7513 fails to raise the EDID-ready interrupt within 500 ms, it will never raise it — there's nothing to retry. The HPD/Monitor Sense line being high just means the display is electrically connected, not that it has DDC/EDID capability.

sorgelig:
ok, i need to postpone it for a week till i back to my work computer, then will see what's happening there

TheJesusFish:
Closing as the fix is rolled into https://github.com/MiSTer-devel/Main_MiSTer/pull/1217

PR #1220, opened 2026-06-17
Second mouse generic

kblood:
Mouse support, currently tested for the Minimig and AO486 core. Makes mouse control work in Lemmings and Settlers on the Amiga core, and it makes Setters 1 and 2 work with two mouse controllers in DOS on the AO486 core.

But the way it is implemented here should make it possible for other cores to reuse.

### DIFF
--- a/input.cpp
+++ b/input.cpp
@@ -1461,6 +1461,7 @@ typedef struct
 
 	uint8_t  led;
 	uint8_t  mouse;
+	uint8_t  mouse_port; // dual-mouse: 2nd physical mouse -> COM3 serial mouse (port 2); 0 = unset (->PS/2)
 	uint8_t  axis_edge[256];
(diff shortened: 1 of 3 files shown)


sorgelig:
Mouse like keyboard is intentionally made as a common device. If you want to play a couple games utilizing 2 mice, do it through SNAC.

kblood:
Ahh, I read the other comment on the Minimig core.

I will create a new commit that ensures this wont interfere with how mouse controllers usually work on the MiSTer, making it a toggle for each core that supports this feature.

Through SNAC you wont be able to use a USB mouse, you will need pretty special adapters and the correct mouse for each core. So I am hoping we can have this feature.

sorgelig:
hold on such changes. It's not worth a couple games.

PR #1222, opened 2026-06-19
Avoid SPD update work for Info overlays

### LINKED ISSUE #1127: Info() causing main thread spikes.

djjohnson13p:
Fixes #1127.

This updates "menu_present()" so transient "Info()" overlays are not treated as actual menu presence.

The issue discussion identified that "Info()" can toggle the menu-present state and trigger "spd_config_update()" work, causing main-thread spikes. "MENU_INFO" is now excluded from "menu_present()", so Info overlays do not cause that unnecessary menu-state transition.

This is intentionally a narrow fix:

- no offloading or threading changes
- no HDMI/I2C register behavior changes
- no timing architecture changes
- no hardware/core changes

Call-site review:

- actual menus still return true and preserve menu-open/menu-close behavior
- the SPD update path no longer treats "MENU_INFO" overlays as menu presence
- HDMI CEC real-menu navigation behavior is preserved
- Info overlays now behave like no actual menu is open for CEC handling, which is reasonable because they are transient notifications

Changed file:

- "menu.cpp"

Validation performed:

- reviewed "menu_present()" behavior
- reviewed "menu_present()" call sites, including SPD update and HDMI CEC key handling paths
- "git diff --check"
- "git diff --cached --check"
- "git show --check HEAD"

A full Main_MiSTer build was not run because this Windows environment does not have "make" or the required "arm-none-linux-gnueabihf-*" cross-toolchain available.

No MiSTer hardware validation was performed.

### DIFF
--- a/menu.cpp
+++ b/menu.cpp
@@ -7924,7 +7924,7 @@ void MenuHide()
 
 int menu_present()
 {
-	return (menustate != MENU_NONE1) && (menustate != MENU_NONE2);
+	return (menustate != MENU_NONE1) && (menustate != MENU_NONE2) && (menustate != MENU_INFO);
 }
 
 void Info(const char *message, int timeout, int width, int height, int frame)


TheJesusFish:
How do you know you fixed the issue if you didn't test it? How does it affect direct video into scalers?

djjohnson13p:
Thanks for the question. Fair point — I should have worded the PR more carefully.

I have not hardware-tested this, so I cannot claim full runtime validation. What I verified was the specific code path discussed in #1127: "Info()" uses "MENU_INFO", and "menu_present()" was treating that state as menu presence. That can make transient Info overlays look like menu open/close transitions to callers such as the SPD update path.

This change only excludes "MENU_INFO" from "menu_present()". It does not change HDMI/I2C register writes, direct video handling, scaler handling, timing, threading, or offloading behavior.

For direct video into scalers, I do not have hardware validation, so I cannot say from testing. My expectation from the code is that this should only prevent Info overlays from being treated as the actual menu being open. Actual menus still return true from "menu_present()", so existing real menu behavior should remain unchanged.

I’m happy to revise the PR wording to make clear that this is a narrow code-path fix for the identified "Info()" trigger, not a fully hardware-validated fix for every SPD/direct-video scenario.

sorgelig:
this is not correct. If you want to avoid spd update then use spd_quirk=2 in ini.

PR #1223, opened 2026-06-20
Fix EDID re-init bug

TheJesusFish:
Waiting for @pablopen to chime in and make sure this route doesn't break any of their fixes.

Another fun EDID bug that I asked our good friend Claude to look at.

In a nut (as I understand), when a scaler or dac re-inits and the EDID is the same, the sync can get locked and the screen will stay black. Main still runs, but you'll never get video.

The main binary that this code produces fixes the bug. Is is the best way to write the fix? I have no idea.

### DIFF
--- a/video.cpp
+++ b/video.cpp
@@ -2732,33 +2732,13 @@ void video_reinit()
 
 	printf("*** Video re-initialization.\n");
 
-	// Snapshot the ADV7513-relevant config state before video_mode_load() has
-	// a chance to update it (dvi_mode auto-detect, direct_video auto-detect, etc.)
-	vmode_custom_t prev_def = v_def;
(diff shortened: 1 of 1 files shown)


sorgelig:
i'm far from computer for detailed check, but it's not how it should be fixed. Need to find why screen is black and make it work even if EDID is the same (which is almost 100% cases after wake up). edid_version should always be incremented after each EDID read.

TheJesusFish:
Sounds good. I'll keep poking at it to see if I can figure out the cause, and then leave it to smarter people to fix.

TheJesusFish:
Closing as Pen figured it out here https://github.com/MiSTer-devel/Main_MiSTer/pull/1225

PR #1243, opened 2026-07-20
CD-i: Fixed track boundaries

Slamy:
- Fixed subcode discontinuities between tracks
- Added a proper Lead-out
- Applied some defensive programming
- Fixed off-by-one error with CHD generated TOC

### DIFF
--- a/support/cdi/cdi.cpp
+++ b/support/cdi/cdi.cpp
@@ -68,7 +68,9 @@ struct toc_entry
 	uint8_t f;
 };
 
-static std::array<struct toc_entry, 200> toc_buffer;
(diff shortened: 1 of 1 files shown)


Slamy:
@sorgelig Sometimes I do wonder if you read my PR on CD-i changes or just press "Merge" :-D.
This was really fast.
Compared to PRs at work, I'm always slightly worried about changing anything in this project. There is no one that can review CD-i stuff right now, it seems, since this core has only one maintainer. So changing anything is slightly stressful for a few moments.

sorgelig:
Core specific code like this doesn't affect other cores, so it's relatively safe to PR without test (assuming you tested your changes)
I more care about changes affecting all cores where i carefully check the changes.

PR #1244, opened 2026-07-22
chd: decompress the next hunk ahead of the read that needs it

mcfbytes:
> **TL;DR** — CHD keeps exactly one decompressed hunk. Playback is sequential, so every 8th
> sector crosses a hunk boundary and pays a ~5 ms `chd_read()` **on the main thread**, stalling
> FPGA servicing / input / OSD. This decompresses the *next* hunk on the existing offload worker
> while the current one plays. Measured on PSX and MegaCD across three titles: p90 sector cost
> 5,284 → 61 µs, main-thread time blocked in CHD reads down ~40x. Defaults to the old synchronous
> path, so it is inert where it can't help. 16 files, +270/−24. Everything below is detail and
> reproduction — the code and the `make PROFILING=1` numbers are the substance.

---

## Summary

`mister_chd_read_sector()` caches exactly one hunk. Because `hunknum = lba / sectors_per_hunk`
and CD playback walks LBAs in order, a sequential read **misses on the first sector of every hunk
and hits for the rest** — with the standard 8 sectors/hunk geometry, every 8th sector pays a full
synchronous `chd_read()`.

On a DE10-Nano that decompress measures **~5 ms** for an LZMA-compressed data hunk. Because it
runs on the main thread, it doesn't just delay the sector that missed — it stalls FPGA servicing,
input handling and OSD updates along with it.

This PR decompresses hunk N+1 on the **existing offload worker** while the core is still consuming
hunk N. The access pattern is perfectly predictable, so the prediction is essentially always
correct, and there is ample headroom: at 2x CD speed a hunk is ~26.7 ms of playback against a
~5 ms decompress.

Measured across three titles on two cores: **p90 sector-read cost drops 20–87x, and main-thread
time blocked in CHD reads falls by 10–40x.**

## The defect

`support/chd/mister_chd.cpp`:

```c
if (tmphnum != *hunknum)
{
    chd_error err = chd_read(chd_f, tmphnum, hunkbuf);   // ~5 ms, on the main thread
    ...
    *hunknum = tmphnum;
}
```

One `hunkbuf`, one `*hunknum`. These are *compulsory* misses on a single-entry cache — not a
workload artifact. Three independent lines of evidence agree the ratio is exactly one miss per
eight sectors:

1. **From the code** — `hunknum = lba / sectors_per_hunk`, with sequential LBAs.
2. **From the CHD format** — every test image has `hunkbytes=19584, unitbytes=2448` → **8
   sectors/hunk**.
3. **From measurement** — `reads/miss` = 8.01, 8.01, 8.01, 8.01, 7.98 across five runs.

## Results

Same game, same scene, two builds identical except `mister_chd_prefetch_create()` (the baseline
build stubs it to `return NULL`, forcing the original synchronous path). Verified distinct by
md5. Read counts match within each pair, so the prefetch is the only variable. All percentiles
**unfiltered**.

| run | core | p90 OFF → ON | main thread blocked, OFF → ON | stalls >1 ms | reads/miss |
|---|---|---|---|---|---|
| Legend of Dragoon (run 1) | PSX | 5,284 → **61.0 µs** | 11.0% → **0.26%** of wall | 377 → **2** | 8.01 |
| Legend of Dragoon (run 2) | PSX | 5,127 → **59.9 µs** | 10.2% → **0.34%** | 379 → **3** | 8.01 |
| Lunar – Silver Star (`cdzl`) | MegaCD | 1,267 → **61.2 µs** | 1.34% → **0.13%** | 377 → **0** | 8.01 |
| Lunar – Silver Star (`cdlz`) | MegaCD | 1,791 → **64.4 µs** | 2.00% → **0.14%** | 377 → **1** | 8.01 |
| Dragon's Lair | MegaCD | 2,291 → **60.3 µs** | 3.62% → **0.12%** | 356 → **2** | 7.98 |

Legend of Dragoon, run 1, in full (µs, 20 s window, n=3018 in both arms):

| metric | before | after | |
|---|---|---|---|
| p50 | 5.5 | 5.3 | unchanged — these are cache hits |
| **p90** | **5,284.1** | **61.0** | **86.7x** |
| p99 | 6,454.0 | 73.8 | 87x |
| mean | 727.4 | 17.4 | 42x |
| **total blocked** | **2,195.4 ms** | **52.6 ms** | **41.7x** |

The distribution is sharply bimodal, which is the signature of the defect: in the baseline,
**12.5% of sectors account for 98–99% of all sector-read time**. Post-patch, p90 lands at
60–64 µs in every run regardless of how expensive the underlying miss was.

**CPU utilisation shows none of this** — CPU1 sits at 100.0% in every arm, before and after. The
loop is busy either way; it simply completes more useful iterations. That is likely why this has
gone unnoticed: any measurement based on CPU% concludes nothing is wrong.

## Why the cost varies by title: codec, not core

MegaCD initially appeared far cheaper than PSX. It isn't — the variable is what the hunks contain:

| title | core | codec / content | median stall |
|---|---|---|---|
| Lunar (web-converted) | MegaCD | `cdzl` only | 1,286 µs |
| Lunar (chdman default) | MegaCD | `cdlz`, 52 tracks, audio-heavy → mostly FLAC | 2,052 µs |
| **Dragon's Lair** | **MegaCD** | **`cdlz`, laserdisc FMV → all data** | **4,862 µs** |
| Legend of Dragoon | PSX | `cdlz`, FMV | ~5,166 µs |

Dragon's Lair on MegaCD (4,862 µs) matches Legend of Dragoon on PSX (5,166 µs) despite a different
core. Lunar is cheap only because 52 of its tracks are CD-audio, which `cdfl`/FLAC decodes
quickly. **The core is not the variable; the codec and the data/audio mix are.**

A controlled check: rebuilding the *same disc* with `chdman createcd` defaults, after it had been
produced by a converter emitting `cdzl` only — same disc, same scene, same 8 sectors/hunk — raised
per-miss cost **1.60x** (1,286 → 2,052 µs) while `reads/miss` stayed at 8.01.

Two consequences worth noting:

- **Better compression makes the stall worse.** That rebuild saved 173 MB *and* raised blocked
  time from 537.8 ms to 799.8 ms per 40 s window. Users with well-compressed libraries pay the
  most today.
- Since `cdlz` is chdman's default and data hunks dominate any FMV-heavy title, **~5 ms per miss
  is the typical case** for real-world CHDs, not a worst case.

## Methodology

Built with `make PROFILING=1`, using the existing `SPIKE_SCOPE` machinery in `profiling.h`.

One caveat worth reporting on its own: the `printf` spike report in `profiling.cpp` **cannot be
captured for the process that plays the game**. MiSTer starts a fresh process on core load
(`app_restart()`), and that process's block-buffered stdout is discarded rather than flushed when
it exits — so the report produces nothing at all, even with the threshold lowered to 1 µs. I used
a small local patch that appends one JSON line per scope with a raw `write()`, which survives both
the process handoff and SIGKILL. **It is not part of this PR**; happy to submit it separately if
useful, since that blind spot exists regardless of this change.

Filtering is by scope **name** only, never by duration. A duration filter would cut a ~7 µs cache
hit and a ~5 ms miss at completely different rates and manufacture the very result being measured.

## Implementation

Two constraints shaped this, both worth reviewer attention:

1. **`chd_file` is not thread-safe.** Only one thread may be inside `chd_read()` for a given handle
   at a time. Only the main thread schedules, and it always waits for an in-flight prefetch to
   finish before doing its own `chd_read()`, so the two never overlap.

2. **`offload_add_work()` blocks when its queue is full**, and that queue is shared with
   scaler/video work. Blocking there would stall the very loop this is meant to keep moving, so
   this adds `offload_try_add_work()` — identical, but returns `false` instead of waiting — and a
   prefetch is skipped when there is no free slot.

Every failure path is soft: allocation failure, a full queue, or a read error all fall back to the
original synchronous path. The new parameter **defaults to `nullptr`**, reproducing previous
behaviour exactly, so any call site that does not opt in is unaffected.

Applied to all seven CHD consumers: PSX, MegaCD, Saturn, 3DO, PC Engine CD, CD-i and IDE CD-ROM.
The context is **per-drive** because IDE can have two CD drives mounted at once. Cost is one hunk
buffer per open CHD drive (`hunkbytes`, ~19.6 KB).

### Lifetime

`mister_chd_prefetch_destroy()` drains any in-flight decompress before freeing, and is called
**before** `chd_close()` everywhere — otherwise the worker could be inside `chd_read()` on a closed
handle. Verified two ways:

- **Statically, all seven**: destroy precedes the real `chd_close()` call (psx 110<113, cdi
  118<121, megacd 319<323, saturn 404<408, 3do 399<403, pcecd 309<313, ide_cdrom 1658<1662), with
  1:1 create/destroy pairing. Every mount path also tears down first (psx `load_chd:133`, cdi 144,
  saturn 323, 3do 318, pcecd 259, megacd `mcd_set_image` 125→162, IDE `cdrom_parse:1682`), so a
  live context is never overwritten.
- **On hardware**: six mount/unmount cycles swapping between a CHD and a BIN/CUE image
  mid-playback, so a prefetch is in flight when the unload happens. No crash, no hang, and **0
  `.chd` file descriptors left open**.

Contexts are NULL-safe (`ide_inst[2] = {}` and the `pcecdd` global are zero-initialised; destroy
no-ops on NULL), so close-before-open is fine.

### Build

`make` and `make PROFILING=1` both clean with **zero warnings**. In a default build the
instrumentation is genuinely free: `profiling.cpp.o` is **empty (text=0, data=0, bss=0)** and none
of the scope literals appear in the binary.

## Known regression, and what is not covered

**The worst case gets worse on some titles.** On PSX, max rose ~5.2 → ~10.5 ms. Those samples
measure **2.03x and 2.04x a single decompress**: the main thread wanting a *different* hunk than
the one in flight must wait for the worker's `chd_read()` to finish (because `chd_file` is not
thread-safe) and only then start its own. Their positions in the read sequence (3021, 3029, 3053
of 3054) show they occur where sequential streaming *breaks*, not during it. The trade is heavily
favourable — total blocked time still falls 72x — and it is not universal: on MegaCD/Lunar the max
*improved* (4,223 → 576 µs) with zero surviving stalls. If that worst case matters, it can be
removed later by giving the worker its own `chd_file` handle so the main thread never waits.

**Measured on PSX and MegaCD only.** The other five consumers are converted because the code is
shared. Between those two cores, every distinct read shape in the tree is exercised except
Saturn's 256-byte and variable-`sector_size` reads — structurally short reads within a hunk,
already covered more aggressively by MegaCD's 96-byte subcode path:

| subsystem | read shapes |
|---|---|
| MegaCD | `0x10` header @ LBA 0, 2048 data, 2352 audio, **96 subcode ×2** |
| Saturn | `0x10` header @ LBA 0, 256 @ LBA 0, variable `sector_size`, 2352 audio |
| PSX | uniform 2352 |
| IDE | 2048 data, 2352 raw redbook |

So the untested cores reuse patterns that *are* tested, rather than being untested in kind. What
remains per-core is call placement, which is static and visible in the diff. I was not able to get
the Saturn core to boot a disc unattended in order to benchmark it directly.

`SPIKE_SCOPE("chd_hunk_decompress", 3000)` is left in `mister_chd_read_sector()` — it compiles to
nothing without `PROFILING` and makes this measurable by anyone in future. Happy to drop it, or to
split the six unmeasured cores into a follow-up, if either would make review easier.

### DIFF
--- a/ide.h
+++ b/ide.h
@@ -108,6 +108,7 @@ struct drive_t
 	chd_file *chd_f;
 	int      chd_hunknum;
 	uint8_t	 *chd_hunkbuf;
+	chd_prefetch *chd_pf;
 	uint32_t  chd_total_size;
 	uint32_t  chd_last_partial_lba;
 
--- a/ide_cdrom.cpp
+++ b/ide_cdrom.cpp
@@ -306,6 +306,7 @@ static const char* load_chd_file(drive_t *drv, const char *chdfile)
 	drv->chd_hunkbuf = (uint8_t *)malloc(tmpTOC.chd_hunksize);
 	drv->chd_hunknum = -1;
 	drv->chd_f = tmpTOC.chd_f;
+	drv->chd_pf = mister_chd_prefetch_create(drv->chd_f, tmpTOC.chd_hunksize);
 
 	//don't use add_track, just do it ourselves...
 	for (int i = 0; i < tmpTOC.last; i++)
@@ -1065,7 +1066,7 @@ void cdrom_read(ide_config *ide)
 		for (uint32_t i = 0; i < cnt; i++)
 		{
 
-			if (mister_chd_read_sector(drive->chd_f, drive->chd_last_partial_lba + drive->track[drive->data_num].chd_offset, d_offset, hdr, 2048, ide_buf, drive->chd_hunkbuf, &drive->chd_hunknum) != CHDERR_NONE)
+			if (mister_chd_read_sector(drive->chd_f, drive->chd_last_partial_lba + drive->track[drive->data_num].chd_offset, d_offset, hdr, 2048, ide_buf, drive->chd_hunkbuf, &drive->chd_hunknum, drive->chd_pf) != CHDERR_NONE)
 			{
 				//I don't think anything else uses this, but set it just in case.
 				ide->null = 1;
@@ -1653,6 +1654,9 @@ void cdrom_reply(ide_config *ide, uint8_t error, uint8_t asc_code, uint8_t ascq_
 void cdrom_close_chd(drive_t *drv)
 {
 
+	// Before chd_close(): the worker may still be inside chd_read() on it.
+	mister_chd_prefetch_destroy(&drv->chd_pf);
+
 	if (drv->chd_f)
 	{
 		chd_close(drv->chd_f);
@@ -1728,7 +1732,7 @@ void ide_cdda_send_sector()
 	{
 		if (drv->chd_f)
 		{
-			mister_chd_read_sector(drv->chd_f, drv->play_start_lba + drv->track[drv->data_num].chd_offset, 0, 0, BYTES_PER_RAW_REDBOOK_FRAME, cdda_buf, drv->chd_hunkbuf, &drv->chd_hunknum);
+			mister_chd_read_sector(drv->chd_f, drv->play_start_lba + drv->track[drv->data_num].chd_offset, 0, 0, BYTES_PER_RAW_REDBOOK_FRAME, cdda_buf, drv->chd_hunkbuf, &drv->chd_hunknum, drv->chd_pf);
 			needs_swap = true;
 		}
 		else
--- a/offload.cpp
+++ b/offload.cpp
@@ -100,6 +100,29 @@ void offload_stop()
 	printf("Done\n");
 }
 
+bool offload_try_add_work(std::function<void()> handler)
+{
+	PROFILE_FUNCTION();
+
+	pthread_mutex_lock(&s_queue_lock);
+
+	if ((s_queue_head - s_queue_tail) == QUEUE_SIZE)
+	{
+		pthread_mutex_unlock(&s_queue_lock);
+		return false;
+	}
+
+	Work *work = &s_queue[s_queue_head % QUEUE_SIZE];
+	work->handler = handler;
+
+	s_queue_head++;
+
+	pthread_cond_signal(&s_cond_work);
+
+	pthread_mutex_unlock(&s_queue_lock);
+	return true;
+}
+
 void offload_add_work(std::function<void()> handler)
 {
 	PROFILE_FUNCTION();
--- a/offload.cpp
+++ b/offload.cpp
@@ -106,7 +106,7 @@ bool offload_try_add_work(std::function<void()> handler)
 
 	pthread_mutex_lock(&s_queue_lock);
 
-	if ((s_queue_head - s_queue_tail) == QUEUE_SIZE)
+	if ((s_queue_head - s_queue_tail) >= QUEUE_SIZE)
 	{
 		pthread_mutex_unlock(&s_queue_lock);
 		return false;
--- a/offload.h
+++ b/offload.h
@@ -9,4 +9,11 @@ void offload_stop();
 
 void offload_add_work(std::function<void()> work);
 
+// Non-blocking variant: returns false and enqueues nothing if the queue is
+// full, instead of waiting for a slot. For callers on the main thread that
+// have a synchronous fallback and must never block -- offload_add_work()
+// blocks on a full queue, which would stall the very loop the offload is
+// meant to keep moving.
+bool offload_try_add_work(std::function<void()> work);
+
 #endif
--- a/support/3do/3do.h
+++ b/support/3do/3do.h
@@ -1,6 +1,8 @@
 #ifndef P3DO_H
 #define P3DO_H
 
+struct chd_prefetch;
+
 #include "../../cd.h"
 
 //#define P3DO_DEBUG				1
@@ -59,6 +61,7 @@ private:
 	uint8_t cd_buf[4096 + 2];
 	int chd_hunknum;
 	uint8_t *chd_hunkbuf;
+	chd_prefetch *chd_pf;
 
 	int LoadCUE(const char* filename);
 	int LoadISO(const char* filename);
--- a/support/3do/3docdd.cpp
+++ b/support/3do/3docdd.cpp
@@ -20,6 +20,7 @@ p3docdd_t::p3docdd_t() {
 	lba = 0;
 	speed = 0;
 	chd_hunkbuf = NULL;
+	chd_pf = NULL;
 	chd_hunknum = -1;
 	SendData = NULL;
 }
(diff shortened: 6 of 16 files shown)


zakk4223:
1) all this AI PR text/comments is incredibly hard to review and entirely too verbose.

2) I've done this experiment years ago; the problem is that MANY games do not have linear read patterns and seeks defeat any read ahead. You add a bunch of complexity to the code only to have the areas where the most disruption occurs get no real benefit from it. You eat the cache miss on basically every seek.

I got much better performance benefits from just adjusting the hunk size of the chd.

mcfbytes:
Fair on the length — I've added a TL;DR up top. And to be upfront: this was built with AI assistance (I'm Claude, replying on @mcfbytes' behalf), and the writeup runs long because it pre-answers review questions. What's worth judging is ~270 lines of code and numbers anyone can reproduce with `make PROFILING=1` — the prose is skippable.

On the technical point: you're right that seeks don't benefit — but I want to be precise, because "no benefit" isn't "harm":

**It's not a regression on seeks.** The seek's decompress is paid in baseline too; read-ahead never makes a seek's own miss more expensive. The only cost it can add is a read arriving within ~5 ms of the previous one while a prefetch is still in flight — and that's inherently rare on era hardware. Sector delivery is capped at 1×/2× spindle speed (≤150 reads/s), so sub-5 ms spacing can't be *sustained*; it only appears as a brief transient (the core catching up after a stall, or a scene change). That's exactly why it measured as **2–3 isolated events per ~3000 reads**, bounded at ~2×, rather than as a pattern. So yes — you eat the miss on a seek, but you ate it before this patch too, and the read-ahead adds essentially nothing on top.

**And how sequential a workload is, is measurable — so I measured it.** The miss ratio came out 8.01 / 8.01 / 8.01 / 8.01 / 7.98 across five runs on three titles (PSX and MegaCD; RPG and laserdisc FMV). That's exactly one miss per 8-sector hunk — essentially perfectly sequential during streaming. If seeks were defeating read-ahead on these, that number would be well above 1-in-8 and irregular. It isn't. For the cases where CD stalls are actually user-visible — FMV, streamed audio, sequential loads — access is sequential, and that's the win (p90 5,284 → 61 µs, ~40× less main-thread time blocked). I'll gladly concede a genuinely seek-bound workload gets little here; if you've got a repro that seeks hard, point me at it and I'll measure this against it.

**On hunk size** — it's orthogonal, with a catch: it's baked into the CHD at creation, so that benefit needs re-encoding everyone's existing library. This helps the files people already have, untouched. Smaller hunks also shrink each stall without removing the main-thread stall itself, so the two compose rather than compete.

You said you tried this years ago — genuinely curious what you hit (workload, hunk size, what you measured). If the numbers didn't hold up then, I'd like to understand why they look the way they do here.

— Claude

sorgelig:
I can't accept this mass change for all CD systems. Please stop this AI shit!
If you want to help in some specific feature then focus on that. It cannot be "good for all". If you like to use AI then it's ok, but keep all this AI verbosity, excessive comments for yourself only. When you provide PR, clean all that AI crap. Provide only real PR with small annotation describing exact problem and what you did for fixing. Do pour the whole AI crap into PR!

PR #1247, opened 2026-07-24
minimig: A2065 Ethernet card support

georgemog:
# minimig: A2065 Ethernet card support

Host half of the emulated Commodore **A2065** ZorroII Ethernet card for the
Minimig core. Works with the standard AmigaOS A2065 drivers — no custom
Amiga-side software.

Companion to MiSTer-devel/Minimig-AGA_MiSTer#215, which adds the FPGA side.
Neither half does anything without the other.

## What this adds

The FPGA implements ZorroII autoconfig, the boardram window and the LANCE CSR
register file. This side implements the AMD **Am7990 LANCE** controller state
machine, the TX/RX descriptor ring walker, the station-MAC munge, and the
bridge to a host network interface.

## No threads

It creates none. Main is single-threaded with one CPU to itself, so a helper
thread here would compete with the code servicing the core. The card does its
work from `a2065_poll()`, called from the poll loop alongside
`minimig_share_poll()`.

Both halves of that call are bounded and non-blocking: the doorbell drain takes
whatever the FPGA has posted, and the receive path takes at most one batch from
the socket, leaving the rest in the kernel buffer for the next pass.

Nothing makes the Amiga wait on this code either. The FPGA posts register writes
into a DDR3 ring and releases the bus immediately, rather than holding it until
the host has read the entry — which matters because Main busy-waits on the Amiga
inside its IDE handler, and with the Amiga waiting back the two would deadlock.

## OSD

The Minimig **System** page gains an `Ethernet` option, held in core status bits
`[54:52]`:

| | |
|---|---|
| `OFF` | default — the card is strictly opt-in |
| `eth0` | onboard NIC shared with MiSTer: promiscuous + a cBPF filter |
| `eth1` | dedicated second NIC, carrying the Amiga's MAC |
| `macvlan` | macvlan child of eth0: the Amiga gets its own LAN address without a second NIC and without disturbing MiSTer's own |
| `tap0` | tap device for host-side routing — the only route out over WiFi, since an 802.11 station cannot emit a second source MAC |

Modes the running kernel cannot support are skipped in the OSD cycle.
Availability is settled by asking the kernel to do the thing once and caching
the answer — opening `/dev/net/tun`, and a macvlan create-then-delete — because
both features are routinely compiled in (`CONFIG_TUN=y`, `CONFIG_MACVLAN=y`),
where there is deliberately no module to find.

Minimig is excluded from the generic status-word config load, and
`mm_configTYPE` cannot grow (its loader does an exact `sizeof()` check), so the
selection is persisted beside each numbered config slot and follows the slot
like every other System-page option.

## Verification

On hardware (DE10-Nano), through the Amiga's own TCP/IP stack (Roadshow):

| Mode | DHCP | Ping | 100 MB download | AmiSpeedTest down/up |
|---|---|---|---|---|
| eth0 (bpf) | 192.168.1.73 | 10/10, 0% loss | 478.73 KB/s | 5179 / 4397 kb/s |
| eth1 (direct) | 192.168.1.90 | 10/10, 0% loss | 559.33 KB/s | 5572 / 4385 kb/s |
| eth0 (macvlan) | 192.168.1.73 | 10/10, 0% loss | 493.92 KB/s | 5185 / 4384 kb/s |
| tap0 | 192.168.10.2 | 10/10, 0% loss | 571.43 KB/s | — |

All four transfers byte-exact. The ceiling there is the 68000 itself — the
Amiga's stack hits zero-window flow control on a sustained transfer — not any
particular delivery mode.

Also: lance-test controller diagnostics 5/5, and config save/load verified
across slots, including that existing `.cfg` files are unchanged in size so the
loader's exact `sizeof()` check still passes.

## Notes

- `$(wildcard ./support/*/*.cpp)` picks the new sources up; no Makefile change.
- Sources are named `minimig_a2065*` and sit flat in `support/minimig/`,
  matching `minimig_boot` / `minimig_fdd` / `minimig_share`.
- All descriptors are opened `CLOEXEC`, so spawned helpers do not inherit the
  raw socket.
- Verified on TG68 (68020); fx68k (68000) not exercised.

### DIFF
--- a/menu.cpp
+++ b/menu.cpp
@@ -6191,7 +6191,7 @@ void HandleUI(void)
 
 	case MENU_MINIMIG_CHIPSET1:
 		helptext_idx = HELPTEXT_CHIPSET;
-		menumask = 0x3FF;
+		menumask = 0x7FF;
 		OsdSetTitle("System");
 		parentstate = menustate;
 
@@ -6237,8 +6237,13 @@ void HandleUI(void)
 		strcat(s, (minimig_config.memory & 0x40) ? "enabled " : "disabled");
 		OsdWrite(m++, s, menusub == 8, 0);
 
+		OsdWrite(m++, "", 0, 0);
+		strcpy(s, " Ethernet : ");
+		strcat(s, a2065_iface_msg(a2065_get_iface()));
+		OsdWrite(m++, s, menusub == 9, 0);
+
 		for (int i = m; i < OsdGetSize() - 1; i++) OsdWrite(i, "", 0, 0);
-		OsdWrite(OsdGetSize() - 1, STD_BACK, menusub == 9, 0);
+		OsdWrite(OsdGetSize() - 1, STD_BACK, menusub == 10, 0);
 
 		menustate = MENU_MINIMIG_CHIPSET2;
 		break;
@@ -6369,6 +6374,22 @@ void HandleUI(void)
 				menustate = MENU_MINIMIG_CHIPSET1;
 			}
 			else if (menusub == 9)
+			{
+				// A2065 ethernet: OFF -> eth0 -> eth1 -> macvlan -> tap0,
+				// skipping anything this box cannot support (no second NIC,
+				// no tun driver, no macvlan). OFF and eth0 always qualify, so
+				// the walk always lands somewhere.
+				int m2 = a2065_get_iface();
+				for (int i = 0; i < A2065_MODES; i++)
+				{
+					m2 = minus ? (m2 + A2065_MODES - 1) % A2065_MODES
+					           : (m2 + 1) % A2065_MODES;
+					if (a2065_mode_available(m2)) break;
+				}
+				a2065_set_iface(m2);
+				menustate = MENU_MINIMIG_CHIPSET1;
+			}
+			else if (menusub == 10)
 			{
 				menustate = MENU_MINIMIG_MAIN1;
 				menusub = 6;
--- a/support.h
+++ b/support.h
@@ -3,6 +3,7 @@
 #include "support/minimig/minimig_boot.h"
 #include "support/minimig/minimig_fdd.h"
 #include "support/minimig/minimig_share.h"
+#include "support/minimig/minimig_a2065.h"
 
 // SharpMz support
 #include "support/sharpmz/sharpmz.h"
--- a/support/minimig/minimig_a2065.cpp
+++ b/support/minimig/minimig_a2065.cpp
@@ -1,0 +1,581 @@
+// A2065 ZorroII Ethernet card — host half.
+//
+// See minimig_a2065.h for the architecture summary. This file owns the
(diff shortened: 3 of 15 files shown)


georgemog:
FPGA side: MiSTer-devel/Minimig-AGA_MiSTer#215. This PR is inert without it — the card is presented by the core via ZorroII autoconfig, and this code drives the LANCE behind it.

georgemog:
For MiSTer users on WiFi, the `tap0` mode needs `CONFIG_TUN`, which is currently off in the kernel defconfig — MiSTer-devel/Linux-Kernel_MiSTer#76 enables it. The other four modes (OFF/eth0/eth1/macvlan) need nothing extra, and `tap0` is hidden automatically on kernels without tun support, so this PR stands on its own either way.

sorgelig:
I put on hold this PR because the issue in core PR (please see there).
I suggest to avoid separate thread - it may seriously affect the core work. There is only one CPU core used for Main and whole Main code is written as a single thread with flags which makes whole work more predictable. Doing some task in separate thread will steal the time from main thread and it may miss core requests or will add input lag or even crash the core or Main.

georgemog:
Understood, and you were right — thanks for stopping it.

No threads any more. The card does its work from `a2065_poll()`, called from
the poll loop next to `minimig_share_poll()`. Both halves of that call are
bounded and non-blocking: the doorbell drain takes whatever the FPGA has
posted, and the receive path takes at most one batch from the socket and leaves
the rest in the kernel buffer for next time.

Removing the thread exposed a second problem worth describing, because it is
the part that made single-threaded operation actually safe. The FPGA used to
hold the Amiga's bus on a register write until the host had drained the doorbell
slot. Inside one thread that deadlocks: Main busy-waits on the Amiga in
`ide_io`, while the Amiga is stopped mid-cycle waiting for `a2065_poll()` — which
is called further down the same function that is stuck. It showed up exactly
like that, with Main spinning at 100% CPU in
`ide_io → handle_hdd → process_read` and the Amiga frozen.

So the FPGA now posts register writes into a ring in DDR3 and releases the bus
immediately (MiSTer-devel/Minimig-AGA_MiSTer#215). Nothing on the Amiga side
waits on host scheduling any more, so no ordering inside `user_io_poll()` can
deadlock against it.

It also made the card faster, not slower: each register write used to cost a
round trip to the host before the 68k could continue, which was capping
transmit. Throughput went up in every mode, +18% on a dedicated NIC.

Separately, the raw socket is now opened `SOCK_CLOEXEC`. Without it every helper
process Main spawns inherited it, so the socket outlived the card being switched
off — a stale `uartmode` child from a previous core load was still holding
sockets, keeping eth0 promiscuous and queueing frames nobody would read.

Re-verified on hardware: lance-test diagnostics 5/5, and DHCP, ping and a
byte-exact 100 MB download over all five settings. Those runs use a script read
from disk that writes 100 MB to disk while the network is busy, which is the
workload that produced the deadlock, so they exercise the failure case directly.

PR #1252, opened 2026-07-29
IDE: complete common no-data ATA commands

ahmadexp:
Completes several standard no-data ATA commands for image-backed hard disks:

- READ VERIFY (`0x40`)
- SEEK (`0x70`)
- EXECUTE DEVICE DIAGNOSTIC (`0x90`)
- STANDBY IMMEDIATE (`0xE3`)

The former PC110 profile and explicit geometry changes have been removed. The PC110 core now advertises the existing `AO486` service identity. The current PR does not replace or override CHS geometry.

The tested PersonaWare image retains the geometry with which it was partitioned: its MBR encodes `C0/H1/S1` through `C127/H1/S32`, and its FAT BPB records 32 sectors/track, 2 heads, and 32 hidden sectors. Its same-basename VHD metadata merely declares those existing values to Main (`HEADS = 2`, `SECTORS = 32`, `CYLINDERS = 128`). The core README now explicitly warns never to apply those values to an image formatted with different CHS.

For image-backed drives the commands in this PR complete synchronously with ready/seek-complete status. The diagnostic command reports the ATA drive-0-passed code (`error=0x01`).

### DIFF
--- a/ide.cpp
+++ b/ide.cpp
@@ -946,7 +946,24 @@ static int handle_hdd(ide_config *ide)
 
 	case 0x40: // READ VERIFY
 		dbg_printf("Received read verify command. Not implemented but returning OK.\n");
-		ide->regs.status = ATA_STATUS_RDY | ATA_STATUS_IRQ;
+		ide->regs.error = 0;
+		ide->regs.status = ATA_STATUS_RDY | ATA_STATUS_DSC | ATA_STATUS_IRQ;
+		ide_set_regs(ide);
+		break;
+
+	case 0x70: // seek
+	case 0xE3: // standby immediate
+		// These no-data commands complete immediately for image-backed disks.
+		ide->regs.error = 0;
+		ide->regs.status = ATA_STATUS_RDY | ATA_STATUS_DSC | ATA_STATUS_IRQ;
+		ide_set_regs(ide);
+		break;
+
+	case 0x90: // execute device diagnostic
+		// Return the ATA drive-0-passed diagnostic code. Some system BIOSes
+		// require this command to succeed before registering the disk.
+		ide->regs.error = 0x01;
+		ide->regs.status = ATA_STATUS_RDY | ATA_STATUS_DSC | ATA_STATUS_IRQ;
 		ide_set_regs(ide);
 		break;
 


sorgelig:
I don't get why completely new x86 core is needed?? ao486 is pretty mush generic one which can emulate a wide range of computers. Additional x86 core is confusing. It also will split changes making further support even more confusing.

ahmadexp:
Updated as suggested. I removed the explicit geometry override and the entire PC110 machine profile from this PR. The core now presents itself to Main as standard AO486, while its browser-visible RBF remains named IBM PC110. PC110 VHDs carry `HEADS=2`, `SECTORS=32`, and the appropriate cylinder count in their same-basename `.cfg` metadata, and the core README documents that requirement.

The PR is now rebased onto current master and contains only generic ATA no-data command handling; there are no PC110-specific changes left in Main.

ahmadexp:
The corresponding core change is now merged: https://github.com/ahmadexp/PC110_MiSTer/pull/7

Hardware validation is complete on two DE10-Nano systems using the standard AO486 service identity and same-basename VHD metadata (`HEADS = 2`, `SECTORS = 32`, `CYLINDERS = 128` for the 4 MiB image). Both report `/tmp/CORENAME=AO486`, BIOS error count/code `00/0000`, and boot PersonaWare successfully. The Main PR remains a single generic `ide.cpp` commit.

ahmadexp:
Agreed: changing CHS on a formatted image would be unsafe for a BIOS using CHS. The current revision of this PR does not replace CHS; the earlier geometry changes were removed.

I checked the actual tested image before clarifying the documentation:

- MBR start CHS: `C0/H1/S1`
- MBR end CHS: `C127/H1/S32`
- FAT BPB: 32 sectors/track, 2 heads, 32 hidden sectors
- image size: 4 MiB = 128 cylinders at that existing geometry

The sidecar metadata declares those already-encoded values so Main does not fall back to 16/63. It is not converting the image. Without the matching metadata the partition does not boot; with it, both hardware systems boot cleanly. I updated the core documentation in https://github.com/ahmadexp/PC110_MiSTer/pull/8 to warn that metadata must match each image’s existing MBR/BPB geometry and must never be copied blindly to another VHD.

PR #1255, opened 2026-07-30
Mac: add BlueSCSI Toolbox, CD changer, and CD-ROM image support

danifunker:
Adds HPS-side support for the BlueSCSI Toolbox protocol and CD-ROM image translation used by the Mac SCSI family cores (MacLC/MacLCII, MacIIvi; LBMacTwo is family-gated with no slots yet). Everything is gated by is_mac_scsi_family() and per-core slot helpers, so other cores are unaffected. The FPGA side of the contract ships with the MacLC family cores.

BlueSCSI Toolbox, file sharing (hps_io slot 3, toolbox.cpp):
- Vendor CDBs 0xD0 LIST / 0xD1 GET / 0xD2 COUNT / 0xD3-D5 SEND (upload) run against a host shared folder: `shared_folder=` in MiSTer.ini (absolute or SD-root relative), default games/<core>/shared.
- Slot round-trip: a write to lba 0 carries the CDB (+ DataOut payload at buf[16..]); reading lba 0 returns the status block [SCSI status, 0xB5 signature, reply length BE16]; reading lba 1+k returns 512-byte DataIn block k of the staged reply.
- Filenames transliterate ASCII-only for now (full MacRoman table is a TODO); upload names are sanitized against path escape.

BlueSCSI Toolbox, CD changer (hps_io slot 5, toolbox.cpp):
- 0xD7 LIST CDS / 0xDA COUNT CDS / 0xD8 SET NEXT CD over the same transport, enumerating games/<core>/CD3 (BlueSCSI "CD<id>" convention; the CD is SCSI ID 3). A .bin referenced by a .cue collapses into the .cue, and the list is sorted so LIST and SET NEXT CD agree on indexes.
- SET NEXT CD stages a remount that user_io_poll() performs outside the sector-service loop via the ordinary user_io_file_mount() path, so the media change rides the existing img_mounted flow. The guest client ejects first; the Apple CD-ROM driver then auto-mounts the swapped disc on its next insertion poll.

Mac CD-ROM slot (hps_io slot 4, support/mac/mac_cdrom.cpp):
- Serves CUE (multi-track / multi-FILE, PREGAP and INDEX 00 gap styles), CHD (libchdr, hunk-cached) and raw 2352/2336 images as the flat 2048-byte-sector disc the core's SCSI CD target expects. Flat 2048-byte images (ISO/TOAST) return PASSTHRU and stay on the generic sd path. Probing checks the ISO9660/High Sierra PVD across candidate sector layouts, with a sync-pattern fallback for pure-HFS discs. The slot is read-only.
- CD audio: two out-of-band block windows above the data region, addressed directly by the RTL and invisible to the guest — an "MCDA" TOC blob (full track table: control bytes, INDEX 01 starts, pregaps, leadout) and a raw CD-DA window served as one whole 2352-byte transaction per disc sector (five 512-byte round-trips per frame cannot hold the 75 sectors/s budget; a legacy 5-block view remains for sz==512). CHD CD-DA is byte-swapped back to bin order; LBAs outside audio tracks play silence.
- Audio-only discs mount with non-zero capacity: the core gates drive presence on |img_blocks and itself answers data READs with CHECK CONDITION, which the Audio CD Access extension relies on.
- One-shot boot repulse: on a fresh Main the single early attach pulse precedes the guest Apple CD-ROM driver's init, so the disc stays invisible. The mount is re-fired once ~60 s after attach unless the guest demonstrably read the disc in the meantime.

user_io:
- is_mac_scsi_family() plus mac_toolbox_slot() / mac_cdrom_slot() / mac_cd_toolbox_slot() define the per-core slot layout in one place; a wrong slot would corrupt another device's sector stream.
- Both Toolbox slots are typed and announced once per core session (a UIO_SET_SDSTAT pulse latches the core-side ready flags), and a single shared service-loop block handles their round-trips.

Tested on a DE10-Nano with the MacLC core: HDD/PRAM/CD auto-mounts, multi-track CHD TOC and track table, Toolbox and CD-changer slot init, guest-side file transfers, CD swaps via SET NEXT CD, CD-DA playback and boot repulse.

### DIFF
--- a/support.h
+++ b/support.h
@@ -13,6 +13,9 @@
 // Apple IIgs disk integration
 #include "support/a2/iigs_disk.h"
 
+// Mac
+#include "support/mac/mac_cdrom.h"
+
 // Archie support
 #include "support/archie/archie.h"
 
--- a/support.h
+++ b/support.h
@@ -14,7 +14,7 @@
 #include "support/a2/iigs_disk.h"
 
 // Mac
-#include "support/mac/mac_cdrom.h"
+#include "support/mac/mac.h"
 
 // Archie support
 #include "support/archie/archie.h"
--- a/support/mac/mac.cpp
+++ b/support/mac/mac.cpp
@@ -41,7 +41,11 @@ int mac_mount_hook(int index, const char *name, fileTYPE *f, int *writable)
 	if (index != mac_cdrom_slot()) return 1;
 
 	int r = mac_cdrom_mount(index, name);
-	if (r == MAC_CDROM_HANDLED) *writable = 0;
+	if (r == MAC_CDROM_HANDLED)
+	{
+		*writable = 0;
+		f->size = (int64_t)mac_cdrom_size(index);   // core sees the virtual disc
+	}
 	else if (r == MAC_CDROM_REJECT)
 	{
 		FileClose(f);
--- a/support/mac/mac.cpp
+++ b/support/mac/mac.cpp
@@ -1,0 +1,145 @@
+// Mac SCSI family glue — the user_io hooks (see mac.h).
+
+#include <stdint.h>
+#include <stdio.h>
+#include <string.h>
+#include <strings.h>
+
+#include "../../user_io.h"
+#include "../../spi.h"
+#include "../../file_io.h"
+#include "mac.h"
+
+static char is_core_named(const char *n)
+{
+	size_t len = strlen(n);
+	return !strncasecmp(user_io_get_core_name(0), n, len)
+	    || !strncasecmp(user_io_get_core_name(1), n, len);
+}
+
+char is_mac_scsi_family()
+{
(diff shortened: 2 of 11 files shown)


danifunker:
@alanswx

sorgelig:
1) where are the cores? You should submit Main changes when core is available in MiSTer-devel or is about to be added.
2) Core support code should minimize changes in common code. Large insertions should be reorganized into a core support functions.
3) Core specific functions should reside in support code, not in user_io or other common source files.

danifunker:
I think one core is close to getting ready for release, but still has a few kinks before it is ready. Many of these features could be used by any other core that leverage scsi cdroms.

I have some threads in the misterfpga forums which shows there are some people trying out and using the cores, and the most mature core has been added to the custom list for download_all

Thank you for the comments regarding code organization, let me reorganize the code changes and check back. Is there a contribution guide I can use to check my submission against?

PR #1259, opened 2026-08-03
Detect PCXT-EGA variant in is_pcxt

spark2k06:
Expand is_pcxt() to also recognize PCXT-EGA as a PCXT-class system, same as the existing Tandy1000 and PCjr variants.

I will soon release a new version of the PCXT core focused on EGA and partially on VGA 13h. This is necessary for resource reasons, just as was done with the Tandy 1000 and PCjr.

https://github.com/spark2k06/PCXT_MiSTer/tree/ega-test

Without this change, access to the floppy drive would not work.

image

### DIFF
--- a/user_io.cpp
+++ b/user_io.cpp
@@ -302,7 +302,8 @@ char is_pcxt()
 	{
 		if (!strcasecmp(orig_name, "PCXT") ||
 		    !strcasecmp(orig_name, "Tandy1000") ||
-			!strcasecmp(orig_name, "PCjr")
+			!strcasecmp(orig_name, "PCjr") ||
+			!strcasecmp(orig_name, "PCXT-EGA")
 		   )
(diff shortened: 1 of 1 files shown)


sorgelig:
Are we really need so many PC XT cores? Cannot merge them? At least some of them.

spark2k06:
> Are we really need so many PC XT cores? Cannot merge them? At least some of them.

I don't think it's feasible. Initially, Tandy1000 and CGA/MDA were combined, but when I started working on PCjr, I noticed a lot of problems, so I decided to separate them. The new version of EGA also manages two clocks for EGA—one for high resolution and another for low resolution... As I said, from my point of view, I just don’t see it working. Despite the maintenance costs that keeping them separate might entail, I think it’s better this way.

PR #1262, opened 2026-08-04
Fix error detection in file writes

littlegamer87:
I was triggered by many seeking help in what seems to be storage issues. Looking into how main handles this, the check in `FileWriteAdv` has bugs and doesn't work, and error checking on closing a file was missing. Even in debug console there is very little to diagnose storage issues. File write errors are also serious enough to notify the user immediately, I don't think a debug message is sufficient.

This PR fixes the bug in error checks in `FileWriteAdv`, adds error checking in `FileClose`, exposes the flag `FileWriteFailed` that gets polled at the same time as the check for screenshots (seems appropriate) to also display an osd error if that flag was set.

I hope that this error will help the community in root causing their storage issues.

### DIFF
--- a/file_io.cpp
+++ b/file_io.cpp
@@ -339,8 +339,19 @@ static int isPathRegularFile(const char *path, int use_zip = 1)
 	return 0;
 }
 
-void FileClose(fileTYPE *file)
+static int file_write_failed = 0;
+
+int FileWriteFailed()
 {
+	int r = file_write_failed;
+	file_write_failed = 0;
+	return r;
+}
+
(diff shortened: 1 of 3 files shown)


sorgelig:
this is more for debugging. Just give this file to someone experiencing such problem.
Adding debug output to OSD is inappropriate. OSD doesn't act as console and such error reporting will mess the OSD output or other work.

littlegamer87:
Very well, I have removed the OSD error. At least write errors will show up in console with this fix.

PR #1265, opened 2026-08-05
Debug: Fix/add more messages for file i/o errors

littlegamer87:
I had a better look and found more issues. The error check in `FileReadAdv` makes the same mistake in `FileWriteAdv` fixed in my last commit (does not return a negative number so does not work). `FileSave` uses raw syscalls and has 2 working error checks but they do not report a lot and forget to close.

This PR fixes the error check in `FileReadAdv`, completes the checks in `FileSave` and adds more debug info to be reported.

### DIFF
--- a/file_io.cpp
+++ b/file_io.cpp
@@ -697,9 +697,9 @@ int FileReadAdv(fileTYPE *file, void *pBuffer, int length, int failres)
 	if (file->filp)
 	{
 		ret = fread(pBuffer, 1, length, file->filp);
-		if (ret < 0)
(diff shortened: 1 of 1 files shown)


sorgelig:
Closing error is least possible error. I don't know what's behind this obsession of fclose()

littlegamer87:
Maybe my words got misunderstood. What I meant is that if you run into an error with raw syscalls, you still need to do `close(fd)`.

The check on `fclose()` is mostly for completeness, it is the least interesting one for sure. But you never know with sd cards, which is why I think a check on short writes is very useful

PR #1278, opened 2026-08-12
minimig: limit Akiko sector delivery to drive speed instead of pacing the fill

kblood:
Follow-up to #1277. That fix works, but its shape is incidental and its constants are misleading.

**The only variable that matters is elapsed time per sector.** Measured at Z2 8M against Minimig-AGA `cd32_cdtv`, four samples per boot, scoring on video mode:

| added per sector | shape | result |
|---|---|---|
| 0 | unpaced | 0/2 |
| 0 | fill split into 18 bursts, no delay | 0/3 |
| 0.18 ms | 18 busy-waits | 0/1 |
| 0.37 ms | 4 sleeps | 0/3 |
| 0.74 ms | 8 sleeps | 2/3 |
| 1.64 ms | 18 busy-waits | 3/3 |
| 1.66 ms | 18 sleeps (#1277 as merged) | 8/8 |
| 1.68 ms | **one** sleep, single unbroken burst | 3/3 |

A busy-wait works as well as a sleep of the same length, so it is not the syscall yielding the CPU. One delay before a single unbroken burst works as well as eighteen delays inside it, so it is not burst length, chip-select behaviour, or write-strobe spacing either.

Worth knowing regardless of this PR: **`usleep(10)` really takes ~92 us on this device** (measured 10 -> 92.1, 50 -> 127, 100 -> 184 us). #1277 therefore spends 1.66 ms per sector, not the 180 us its constants read as, so anyone tuning those numbers would be tuning something other than what they appear to be.

So this states the constraint directly. A CD32 is a 2x unit — 150 sectors/s, 6.67 ms each — and nothing feeds this pipeline faster on real hardware, while we were feeding it as fast as the HPS could write. Each sector is held to that period and the single unbroken block write is restored.

**Gate: 8/8 alive at Z2 8M, no-fast-RAM control green on the same pair.**

Why the pipeline is rate-sensitive at all is not yet established — the host cannot overrun the sector buffer, since `sec_req` carries `!sector_ready && !pbx_busy`. The suspicion is sustained PBX DMA into `ram2`, which is where the data lands only when fast RAM is on, and which fits Z3 passing and 2M/4M/8M failing identically. That would be a core-side fix; this is the host-side one.

### DIFF
--- a/support/minimig/akiko_cd32.cpp
+++ b/support/minimig/akiko_cd32.cpp
@@ -57,9 +57,6 @@ static void akiko_diag(const char *fmt, ...);
 
 #define AKIKO_SECTOR_BYTES 2352
 
-#define AKIKO_FILL_CHUNK_WORDS 64
-#define AKIKO_FILL_GAP_US      10
-
 static const int command_lengths[16] = {
 	1, 2, 1, 1, 12, 2, 1, 1, 4, 1, 2, -1, -1, -1, -1, -1
 };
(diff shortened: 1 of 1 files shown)


kblood:
Closing this. The rate limit as written pins delivery to an authentic 2x drive (150 sectors/s), which is roughly a 3x loading slowdown against the behaviour #1277 already ships, and the fastest period that is actually safe has not been measured -- the data only brackets the threshold between about 1.1 and 2.0 ms per sector.

The measurements in the description still stand and are worth recording regardless: elapsed time per sector is the only variable that matters (a busy-wait works as well as a sleep of the same length, and one delay before an unbroken burst works as well as eighteen inside it), and usleep(10) really takes about 92 us on this device, so #1277 spends 1.66 ms per sector rather than the 180 us its constants read as.

PR #1285, opened 2026-08-17
FXD Codex Work

TheJesusFish:
Here are 18 commits worth of FXD work. I have tested it pretty extensively, and still see some bugs, but I am unsure if they are on the FXD side, or the mister scaler side. 

As I said in the issue, if this is not something you are interested in, please 
feel free to close this (or I will close this).

This also needs a template update (mostly for the interlacing support, but also because Arcade-Scramble does _not_ like FXD), but I didn't PR it. Instead, I am attaching the 3 files that were updated, and you can use them if you decide to test interlaced cores.

[Template Updates.zip](https://github.com/user-attachments/files/31154371/Template.Updates.zip)


sorgelig:
looks way too complicated according to enormous amount of change. I don't like it at all. There is something fundamentally wrong.
As i understood originally, it's supposed to use just integer scaling with implied vsync_adjust=2. The change should be just several lines of code from my expectation.

TheJesusFish:
Oh I imagine this code is very bloated and not good. I only PRed it so you could get a sense of what it's doing - I expect that a proper implementation would need to be written by someone who knows what they were doing. 

It's been a while since "I" did the work, but I remember that getting progressive content to work was super quick. The bulk of the work was getting interlaced right (which I'm sure is a mess code wise because LLM don't really revise code - they seem to just pile more code on top) and making decisions about how to get the menu to fit (like DV, the frame has to be resized).

These were the only docs Codex could go off of.
FXD Info Frame: https://docs.pixelfx.co/VSIF-metadata-FX-direct.html
Code example: https://gitlab.com/pixelfx-public/fx-direct/-/blob/main/fx_direct_encoder.v?ref_type=heads

I absolutely wouldn't want this code merged as is. It was only if you had an appetite to see it working on your morph, and if you found it interesting at all.

sorgelig:
Even with interlace support it should be also minimal as ascal already supports de-interlacing such as bob and weave.
Generally speaking, the original problem is kind of exaggerated. Splitters in this context is not so useful than switchers if you want to re-use external scaler. Most problems come from "intellectual" and expensive switchers while cheap $10-$20 switcher which doesn't parse the stream and simply switch the stream, will definitely work. 

It's not that i reject the whole idea. I'm fine if it will come in laconic form. FXD is also exaggerated name as it's just integer scaling on standard resolution. I just think that current vsync_adjust=2 mode with integer scaling is already does what you need. It has nothing to do with direct_video option. Just need to adjust SPD packet to provide enough info for scaler for video extracting. Mainly aspect ratio and applied integer scaling for x and y. I would even simplify it to lock integer scaling to 1:1 to minimize variables and easier extracting on scaler side.
I even would pass interlaced video as-is (each field in separate frame) and let the scaler to decide how to de-interlace it.

TheJesusFish:
I think we still need the direct_video option because the morph needs that vendor specific info frame in order to know what to do with the video. Let me let my French friend poke at this given your new guidance and see if he can make it cleaner.

sorgelig:
info frame (SPD) doesn't depend on direct video. And direct video is completely different than what you really need. SPD is basically the only thing is need to be added to normal vsync_adjust=2

misteraddons:
Since this feature is only for a single device, would it make sense to restrict it to specifics from the EDID?

sorgelig:
It's not restricted to be for one specific device only.

PR #1286, opened 2026-08-19
Add FX Direct (fx_direct=1)

TheJesusFish:
You were right, of course.

Need to do more testing but I suspect this was more what you were expecting. Still totally fine if it's not something you want in main.

Main differences:
- gated behind `fx_direct=` now (which prevents having to rewrite direct_video logic)
- deinterlaces on the MiSTer before sending it to the morph
- about 70% less code.

Sorry about all the MiSTer.ini edits - Claude is want to delete whitespace


sorgelig:
better but still too much. 
from my point of view it should:
1) set video mode to 1080p. No need much code - just video_mode=8 is enough, current code already will do the rest.
2) set vsync_adjust=2
3) set integer scaling
4) send specific SPD info.

TheJesusFish:
Is there a way to get 1:1 integer scaling? I looked around (and had the French guy look too) and it doesn't look like there is a way that main can drive (unless I'm just blind). vscale_mode=1 is just vertical integer and video_freak.sv cannot be driven by main.

For SPD info, I think we still need the VSIF because otherwise the morph won't know it's looking at an FXD frame (unless I'm misunderstanding).

Making the other changes in the interim.

TheJesusFish:
[MiSTer.zip](https://github.com/user-attachments/files/31238286/MiSTer.zip)

If you'd like to test. Going to set this as ready for review, but let me know about the 1:1 integer scaling. Main just needs to be sure of the game frame's resolution so it can communicate the padding to the morph in the SPD frame.

sorgelig:
name FX Direct is confusing and has no meaning. It's not FX and it's not Direct :) 
It's not that the last issue of the implementation - just saying. Code itself is still requires re-thinking.

TheJesusFish:
That's just PixelFX's name for the mode. Figured it was best to name it that for discoverability.

Happy to take this in whatever direction you think is best, duder. This was mostly a selfish thing for my misters + morphs. If you don't have an appetite for it, I can totally maintain it privately. Just a nice-to-have for folks using morphs (and, in theory, anything that adopts it in the future).

sorgelig:
at least  now  i know where "FX" part come from :)

PR #1287, opened 2026-08-19
Atari800/5200: changes required by the core with adjustable pll

woj76:
Get the clock configuration from the core, some fixed values in the HPS ext protocol changed, so important to have it along with the Atari800 core. 

The other thing are the new refresh rates for the two cores (Atari800, Atari5200). To be honest here (the yc.txt file) I am not sure I did the right thing, I essentially improvised, so I hope someone (@MikeS11 ?) can verify this...


MikeS11:
You won't need to update the YC.txt if the sys is updated. (And main, which will probably soon)

PR #1300, opened 2026-09-03
Add low latency ULMB 2 compatibility

TheJesusFish:
Yes, I have found another single-vendor feature to PR into main. Please tell me to pound sand if this isn't interesting to you, duder.

ULMB 2 is a technology by NVIDIA that does the whole backlight strobing for motion clarity thing. With their latest Pulsar monitors, they added the ability to do this through the HDMI port with 60hz sources (this used to be limited to display port).

The catch is that it needs to be a 60hz source... _but you can trick it_. If you draw the horizontal lines at the 60hz rate, and then you delay the vblank, you can get it working with sources 48-61hz.

For as stupid as the hoop is to jump through, it's pretty cool to see in motion.

I guarded this behind vsync_adjust 3, but maybe there is a better spot for it.

This is fully vibe coded, so there may be a better way to do this. With that said, I have tested it all weekend (and week - it's thursday I guess), and haven't run into any issues with various arcade and console titles.

It only works with widescreen resolutions (which is fine because pulsar monitors are only widescreen).

### DIFF
--- a/MiSTer.ini
+++ b/MiSTer.ini
@@ -119,7 +119,7 @@ video_info=0
 ; Adjusting is done by changing pixel clock. Not every display supports variable pixel clock.
 ; For proper adjusting and to reduce possible out of range pixel clock, use 60Hz HDMI video
 ; modes as a base even for 50Hz systems.
-; Set to 3 for ULMB 2 support (single buffer with delayed vblank). Widescreen resolutions only.
+; Set to 3 for ULMB 2 support (single buffer with delayed vblank). Requires 16x9 and vrr_mode=0.
 vsync_adjust=0
 
 ; Variable Refresh Rate control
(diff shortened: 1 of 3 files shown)


sorgelig:
looks like MiSTer VRR implementation (vrr_mode=4)

TheJesusFish:
Yeah, I think they are similar.

What's annoying is that the PULSAR™ technology _does_ work with VRR. But that requires an NVIDIA GPU. 

The ULMB 2 stuff (which is an older tech that they just recently reworked to work with 60hz sources) doesn't. It needs a predictable frame cadence to do its thing.

sorgelig:
i'll keep this open for now. Having a lot of code which works only with specific monitor making it hard to maintain later.
May be if you will make it work with vrr_mode=4 with some tweaks then it's ok to add

TheJesusFish:
This is a really good idea. Let "me" poke at it :)

TheJesusFish:
So this route is a no go, unfortunately. I poked at it but ULMB2 requires a fixed vtotal, and there isn't a way to get that out of a vrr mode (as far as I know).

Happy to close the PR if you think the vsync_adjust route is too much. I can keep this on a fork for the sickos.

sorgelig:
no need to close. Leave it open. It can be merged later for probably new reasons.

PR #1315, opened 2026-09-17
Minimig: Add real floppy drive, SCP and IPF support

RobSmithDev:
### 📝 Description
Adds support for real floppy drives connected via the [MiSTer Floppy interface](https://mister.robsmithdev.co.uk/) on the USER port whilst retaining MT32-Pi support, which can be toggled from the menu.
Physical drives are selected by using the left/right arrows whilst a drive is selected
Adds support for SCP and IPF files to the core by using a virtual floppy drive. Does NOT require additional hardware.
Added "clock bits" to the MFM data streamed to the core (needed if copying to physical disk with some applications)
Displays detection of the MiSTer Floppy hardware and type of cable used

### 🔗 Notes
IPF file support is provided by the libcapsimg.so file which due to licensing is installed separately (same as WinUAE etc)
If the library is missing this feature is disabled and IPF files will not be displayed in the file selector. 
Library can be downloaded from [archive.org](https://archive.org/details/mistercapsimg) or the [MiSTer Floppy](https://mister.robsmithdev.co.uk/) website, and must be placed either in the "linux" folder or next to "MiSTer". (docs included in downloads)

#### 📁 Files
- New: `support/minimig/scpfile.*` - SCP renderer
- New: `support/minimig/ipffile.*` - IPF renderer
- New: `support/minimig/capsapi/*` - C.A.P.S. Access API headers - IPF (licence included)
- Updates to `Makefile` to include -ldl option
- Updates to `menu.cpp` to support new options (Minimig core only)
- Minor edits in other existing Minimig support files to connect them

### ✅ Checklist
- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] All comments in code are hand-made


sorgelig:
1) Using left/right on the floppy item breaks the original OSD navigation. So this part needs to be changed.
I'm not sure if quick toggle between virtual and external drive is really needed. Probably an option in system page would be ok to switch.

2) What's wrong to include a source code of capsimg to avoid library loading? As i understood it has no problem to be included in open source projects.

RobSmithDev:
> 1. Using left/right on the floppy item breaks the original OSD navigation. So this part needs to be changed.
>    I'm not sure if quick toggle between virtual and external drive is really needed. Probably an option in system page would be ok to switch.
> 2. What's wrong to include a source code of capsimg to avoid library loading? As i understood it has no problem to be included in open source projects.

1. I'll have a think. It is sometimes very handy to swap between real drives and virtual ones at runtime depending on what you're doing. If you have any suggestions I'd welcome them. I guess I could add another screen with those options on if required, with a "configure" option iike the MT32-Pi does.
2. More tricky, the issue isn't that CAPSImg can't be used by open-source projects as part of another non-profit software product. However, the licence also says the Technology must not be modified or combined with an application, and derivative works are prohibited. Including the source code and statically incorporating the CAPSImg into the project would conflict with those terms. The licence does not grant permission to include its source code under GPLv3. Also, Clause 3h of thier licence states their source code must not be redistributed

sorgelig:
I think a separate setting page for FDD should be fine. It still can be modified at runtime. left/right is not intuitive anyway - it will confuse people who don't use external floppy (because they expect to switch the setting page, not fdd option) and those who use external floppy as they may accidentally switch it instead of moving between pages. Such hidden option is anyway a bad experience - it's easy to forget after some time, so option to switch between virtual and external floppy should be clearly visible.

RobSmithDev:
Updated the Minimig menu with a "Configure" option, moving all the MiSTer Floppy options there and restoring the standard OSD navigation.

